PR #22, opened 2018-01-30
Minor README fixes

killerducky:
Add -l for model argument for main.py examples.
Also add BOARD_SIZE=19.

Looks like the other main.py commands take the model as a positional argument so maybe that's what the real fix should be.


killerducky:
I signed it!

PR #29, opened 2018-01-31
Autofmt all python

Kashomon:
Autoformat all the python files with autopep8

### DIFF
--- a/cloud_logging.py
+++ b/cloud_logging.py
@@ -25,8 +25,8 @@ LOGGING_PROJECT = os.environ.get('LOGGING_PROJECT', '')
 def configure(project=LOGGING_PROJECT):
     if not project:
         sys.stderr.write('!! Error: The $LOGGING_PROJECT enviroment '
-              'variable is required in order to set up cloud logging. '
-              'Cloud logging is disabled.\n')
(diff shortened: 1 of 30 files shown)


Kashomon:
I basically just ran: autopep8 --in-place *.py

Kashomon:
For: https://github.com/tensorflow/minigo/issues/30

Not fixed yet. Will need to hook it up as a presubmit and add instructions.

killerducky:
I ran autopep8 on dual_net.py, and I get a few changes from what is in master. Did you run with --max-line-length other than default 79? It also added a blank line after each class declarations.

```
-        probabilities, value = outputs['policy_output'], outputs['value_output']
+        probabilities, value = outputs[
+            'policy_output'], outputs['value_output']
```

Kashomon:
I left it at the default.

killerducky:
Ok what else could be the difference? I have autopep8 0.9.1. Anything else?

Kashomon:
Ah. That's a pretty old version

```shell
autopep8 --version
autopep8 1.3.4 (pycodestyle: 2.3.1)
```

killerducky:
That's what ubuntu pulled when I did apt install python-autopep8.  :(

PR #31, opened 2018-01-31
Script for looking at model accuracy on pro positions

jmgilmer:


requirements.txt
```diff
@@ -8,3 +8,5 @@ protobuf
 six
 petname
 google.cloud.logging
+matplotlib
```
amj: these are some heavy dependencies; matplotlib in particular will probably give folks headaches.  Could we maybe make this 'requirements-analysis.txt' so folks don't necessarily need to add them all to run the oneoffs

sgf/fetch_dataset.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
amj: Can you put a comment in here that these games are public domain?

training_curve.py
```diff
@@ -0,0 +1,250 @@
+"""
```
amj: can we move this file to the oneoffs/ directory?

training_curve.py
```diff
@@ -0,0 +1,250 @@
...
+  if props.get('KM') is None:
+    return False
+  return props.get('KM')[0] == komi_str
+
+def find_and_filter_sgf_files(base_dir, min_year = None, komi = None):
+  base_dir = "sgf/baduk_db"
+  sgf_files = []
+  count = 0
+  print("Finding all sgf files in {} with year >= {} and komi = {}".format(base_dir, min_year, komi))
+  for i, (dirpath, dirnames, filenames) in enumerate(os.walk(base_dir)):
```
amj: recommend using tqdm for ez progress bars (here and for all for loops that take longer than 10s :)

sgf/fetch_dataset.sh
jmgilmer: Done

requirements.txt
jmgilmer: Done

training_curve.py
jmgilmer: Done

training_curve.py
jmgilmer: Done

oneoffs/training_curve.py
```diff
@@ -0,0 +1,251 @@
...
+from tqdm import tqdm
+from sgf_wrapper import sgf_prop
+from gtp_wrapper import make_gtp_instance, MCTSPlayer
+from utils import logged_timer as timer
+from tensorflow import gfile
+
+tf.app.flags.DEFINE_string(
+    "sgf_dir", "sgf/baduk_db/", "sgf database")
+
+tf.app.flags.DEFINE_string("model_dir", "gs://mugozero-v1/models",
```
amj: nit: remove bucket name, practice is to make a dir called 'saved_models'

amj: i'll do it, nm

PR #32, opened 2018-02-01
Move Python files into a py subdir

Kashomon:
Moving the python files into a py subdirectory, which cleans up the top level significantly.

Note: I need to rebase the PR. It's currently kinda a mess.


Kashomon:
Sounds like we're going to table this for now.

Issue #28, opened 2018-01-31
Using too much GPU memory


killerducky:
Running `BOARD_SIZE=19 python3 /home/aolsen/projects/minigo/minigo/main.py gtp -l /home/aolsen/minigo-models/000199-lightning -r 100 -v 3`, `nvidia-smi` shows it uses 2558MiB of GPU memory. After running `genmove b` it goes to 2662MiB. This is preventing me from running two minigo instances to test changes since my GPU only has 4GB.

LZ only uses 483MB for a larger 20x256 network (note LZ computes policy and value heads on the CPU).

@amj suggested this link https://stackoverflow.com/questions/34199233/how-to-prevent-tensorflow-from-allocating-the-totality-of-a-gpu-memory

PR #33, opened 2018-02-01
Set gpu_options.allow_growth = True

killerducky:
Reduces memory used for 199-lightning from 2552MiB to 523MiB.

Fixes issue #28.


amj:
Thanks!

Issue #35, opened 2018-02-01
Add some automatic python linting


Kashomon:
Would help catch lots of bugs and help ensure style conformance.

Kashomon:
Plan is to auto-run linting, formatting check, and tests on code submission. for linting, there'll be a blacklist while things are cleaned up.

Kashomon:
I've still got more files to clean up, but this is automatic now.

amj:
Have we got some docs on how to do this/what our project-wide flags are for autopep8?

tommadams:
By default, autopep8 is fairly lax about whitespace and will for example allow function arguments that span multiple lines to be formatted in a bunch of different ways. This is different from something like clang-format, which enforces a one-true-way of formatting.

You can run autopep8 with two -a command line arguments to tell it to be much more aggressive about reformatting whitespace. Maybe we should consider using that?

brilee:
I'm happy with the linter in our automated testing enforcing no style regressions and gradually adding things to be linted.

PR #38, opened 2018-02-01
Several python cleanups

Kashomon:
- Add some `__init__.py` files so we can make importing work.
- Move shipnames into shipname.py. (Not having to drag the data.txt around is nice).
- Add a basic .pylintrc file, generated with `pylint --generate-rcfile > .pylintrc`
- Ignore all files except cloud_logging and coords, which I've fixed up.
  I'll fix up the other files in a later CL.
- Add a test.sh script for running linting + tests.
- Pin tensorflow to 1.4 for now

Partially addresses: https://github.com/tensorflow/minigo/issues/35


Kashomon:
Good idea. Fixed the makefile.

PR #41, opened 2018-02-02
Implement progressive widening.

killerducky:
AKA parentQ - 0.4, AKA in LZ as FPU reduction (First Play Urgency). I still don't know what to think of this, but it shows good results in both LZ and now minigo. I ran a quick test with just `-r 100`. One of them I left default, the other I set `MINIGO_PWIDEN=0.4`

```
minigo-000199-lightning-p-100-pwiden-0.0 v minigo-000199-lightning-p-100-pwiden-0.4 (30/1000 games)
unknown results: 1 3.33%
board size: 19   komi: 7.5
minigo-000199-lightning-p-100-pwiden-0.0      6 20.00%       2  12.50%     4  28.57%     48.57
minigo-000199-lightning-p-100-pwiden-0.4     23 76.67%       9  64.29%     14 87.50%     49.47
                                                             11 36.67%     18 60.00%
```

See also:
https://github.com/gcp/leela-zero/issues/696
https://github.com/gcp/leela-zero/commit/47c06a8d87809fd9b600fac67e979223c0a46221

Edit: The 1 unknown result was a game that ended pass-pass. Seems ringmaster didn't like that, I'll have to look later. I guess my tests with LZ always had someone resign?

### DIFF
--- a/mcts.py
+++ b/mcts.py
@@ -22,6 +22,7 @@ import numpy as np
 import collections
 import random
 import math
+import os
 
 import coords
 import go
@@ -34,6 +35,9 @@ c_PUCT = 1.38
 
 def D_NOISE_ALPHA(): return 0.03 * 19 / go.N
 
+
+# Progressive widening
+PWIDEN = float(os.environ.get('MINIGO_PWIDEN', 0.0))
 
 class DummyNode(object):
     """A fake node of a MCTS search tree.
@@ -84,7 +88,11 @@ class MCTSNode(object):
 
     @property
     def child_action_score(self):
-        return self.child_Q * self.position.to_play + self.child_U - self.illegal_moves
+        return self.child_Q * self.position.to_play + self.child_U - self.illegal_moves - self.child_PWiden
+
+    @property
+    def child_PWiden(self):
+        return PWIDEN * (self.child_N == 0)
 
     @property
     def child_Q(self):
@@ -268,13 +276,13 @@ class MCTSNode(object):
         output = []
         output.append("{q:.4f}\n".format(q=self.Q))
         output.append(self.most_visited_path())
-        output.append(
-            "move:  action      Q      U      P    P-Dir    N  soft-N  p-delta  p-rel\n")
-        output.append("\n".join(["{!s:6}: {: .3f}, {: .3f}, {:.3f}, {:.3f}, {:.3f}, {:4d} {:.4f} {: .5f} {: .2f}".format(
+        output.append("move:  action      Q       U    PWiden   P    P-Dir    N  soft-N  p-delta  p-rel\n")
+        output.append("\n".join(["{!s:6}: {: .3f}, {: .3f}, {: .3f}, {:.3f}, {:.3f}, {:.3f}, {:4d} {:.4f} {: .5f} {: .2f}".format(
             coords.to_human_coord(coords.unflatten_coords(key)),
             self.child_action_score[key],
             self.child_Q[key],
             self.child_U[key],
+            self.child_PWiden[key],
             self.child_prior[key],
             self.original_prior[key],
             int(self.child_N[key]),


mcts.py
```diff
@@ -35,6 +36,9 @@
 def D_NOISE_ALPHA(): return 0.03 * 19 / go.N
 
 
+# Progressive widening
```
Kashomon: Can you document here what progressive widening is and why it's useful? Part of the goal of the project is to be understandable to ml newbies like myself (although obviously, we have a long way to go).

mcts.py
```diff
@@ -35,6 +36,9 @@
...
+PWIDEN = float(os.environ.get('MINIGO_PWIDEN', 0.0))
```
Kashomon: How about a constant instead. I'd like to keep the proliferation of env vars to a minimum. 

I can imagine passing in a config file of ML constants if we want to configure things like this.

mcts.py
killerducky: Progressive widening is just a name I made up. This pull is more of a way to start a conversation about what to do, I'm not very comfortable with this as is but the results are quite good. When things become more clear I'll add some notes or a link to something. For now you can follow the long thread here, or look at my most recent summary post:
https://github.com/gcp/leela-zero/issues/696#issuecomment-362687915

mcts.py
killerducky: @amj what are the plans for these sort of tunable things? Stick with env vars? Command line options? Config file?

I don't want to use a constant because to test it I use the same code and pass different tunings with the env var.

amj:
TL;DR i'm very intrigued at these results, and i think they're definitely worth looking at further.  I think we need to measure what this does better before we adapt it.

These changes do represent a very marked difference from the algorithm as described, so i think the standard for adopting them is somewhat high, in the absence of the ability to test them at scale quickly.
Consequently, i don't think 'stronger play vs itself' is a sufficient argument, yet.  

Here's why:
 - In selfplay, MCTS' purpose is to improve the policy network.  This 'improvement' is both finding good moves and also exploring subtrees unknown to the model.
 -  We can measure how much it's improving policy by looking at how many moves selfplay ends up playing that were *not* its first choice, how many of those moves were unknown prior to the addition of noise, etc.
 - In "init-to-0", we could see that the winning player would almost always play its first-chosen move, due to the asymmetry init-to-0 provided to the two sides.
- Init-to-0 would also greatly dilute the MCTS of the losing player by sending at least one move down nearly every open point on the board
 - As a result, we could reason that init-to-0 was really limiting how much MCTS could improve policy (we probably should've quantified this)

All this is a long way to say that i don't think stronger results necessarily mean better MCTS (for the purposes of selfplay/training).  This might be a key way to get better performance out of CGOS bots, which we should definitely do and measure.  That doesn't mean it's the right thing to do for selfplay/developing a better model (although it could be :) -- just that we need to be clear about what we hope to accomplish with this change and how we expect it will be a better policy improvement operator than the one described.

(also, the pass-pass game was probably because we don't support the GTP command for counting the score.  That should be trivial to add, i'll open an issue.)

mcts.py
amj: usually command line options have worked well although they can be obnoxious to plumb through everywhere.  They are easy to provide tests for, though ;)

killerducky:
I'm also a little worried about the impact on the feedback cycle probability_p -> modified PUCT search -> probability_pi -> training, new probability_p, repeat.

Also I was just rereading the paper AGZ refers to regarding PUCT, "Multi-armed Bandits with Episode Context". They have a very different formula from what AGZ used. I wonder what the reason is? I'm going to try the formula from the paper and see what happens.

killerducky:
Ok I have a starting point for implementing the PUCT version from the Rosin paper I mentioned above:
https://github.com/killerducky/minigo/blob/puct/mcts.py#L85

Their formula uses [0,1] for values, not [-1,1] so it needs to be adjusted. It also doesn't incorporate any prior knowledge of parent_Q (x[i] in the paper) because the paper was written before Value nets. 

It looks like these property functions are called a lot, and it recalculates 362 element numpy arrays every time. Is there some way to change this code? Also see the loop I added to handle the exceptions, there must be a numpy way to do this.

About the formula itself, first a cheatsheet:
```
M = probability_p, sums to 1
Mi = M for move i
Z = all possible board states
z = a board state
D = payoff distibution [0,1]
Di has mean ui
K = number legal moves
t = trial number, start from 1 not 0
xit = payoff for trial t, move i, drawn from Di

Figure 1: Multi-armed bandit policy PUCB
it = argmaxi (xi + c(t,si) - m(t,i))
si = number pulls on i 
xi = average payoff so far if si>0, otherwise 1
c(t,s) = sqrt(3*log(t)/(2s)) if s>0, otherwise 0
m(t,i) = (2/Mi) * sqrt(log(t)/t) if t>1, otherwise 2/Mi
```

The way I understand it compared to AGZ, there are two major differences. First, U is split into separate c and m terms. c handles exploration. m handles favoring high M[i] (probability_p) moves, which diminishes as more playouts are done. Second, they scale according to sqrt(ln(t)) instead of just sqrt(t).

brilee:
@killerducky I haven't delved into the paper you mentioned so I won't comment on that. I'll just note that the repeated, seemingly wasted recomputed child_action_score is because action score changes basically every time a node is touched (N is incremented, sum(child_N) is incremented, virtual losses get applied, etc.)

amj:
hey, i'm happy to close this as obsolete as this PR is about a year old here (maybe a few merge conflicts ;) and we have a cpuct formula from the published 3rd paper to use for progressive widening.  @killerducky any objections?

PR #45, opened 2018-02-02
faster opening statistics in C

marcocalignano:
@amj I tried a lot of way to make the opening stats faster but the only one that is really fast is C. 
It is at least 5 time faster than grep, and easier to add feature to. I would like to know if you can use it in your systems, if I need to write a CMakeList.txt to compile it. After I know you can use it I will put all the feature you requested in.

### DIFF
--- a/sgf/openstat.c
+++ b/sgf/openstat.c
@@ -1,0 +1,47 @@
+#include <stdio.h>
+#include <dirent.h>
+#include <string.h>
+
+#define BUFFER_SIZE 400
+int main(int argc, char *argv[]) {
+    DIR *dir;
+    char buffer[BUFFER_SIZE+1] = {0};
+    struct dirent *ent;
+    long count = 0;
+    long count55 = 0;
+    long count44 = 0;
+    long count54 = 0;
+    FILE *fp;   
+    int bufSize;
+    char filename[300] = {0};
+    dir = opendir(argv[1]);
+    while((ent = readdir(dir))) {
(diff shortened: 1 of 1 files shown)


marcocalignano:
I signed it

amj:
This is some good code -- thanks for taking the initiative and getting this out here.  That said, i'm not sure if the self-play analysis is going to need to be written in C for speed.  It seems more likely that the analysis process is less performance oriented and will probably need to be flexible.  Why don't we revisit this after we've released the self-play data and we have some better ideas around how folks will try to use it?

marcocalignano:
@amj I had an Idea this morning. Every position in Og in a SGF file is just a string or a pattern. 
For example the "orthodox" opening is something like that ``` B[qp];W[**];B[pd];W[**];B[oq] ```. 
Of course you have to keep track of the symmetry and maybe the move order. But if my C program would learn to parse at command line patterns like this it would really helpful even to search joseki in the game or just opening patterns instead of only the first move. We could also create a dictionary with the name and the string of the pattern like 
```
Orthodox     B[qp];W[**];B[pd];W[**];B[oq]
Sanrensei   B[pd];W[**];B[pp];W[**];B[pj]
...
```
It is also easy to create this pattern you can just play them on your favorite program, just save the sgf and then extract them with a text editor. I bet you can go online and find all the joseki and save them into sgf too. 
Let me know if you like this Idea.

amj:
@marcocalignano we've had a bunch of opening statistics available on Seth's site cloudygo -- mind if i close this as obsolete?

PR #47, opened 2018-02-03
Terminate at depth

### LINKED ISSUE #21: Redo selfplay termination after 450 moves
with vlosses the 'terminate after too many moves' check got taken out.  Needs to go back in before our v4 run.

amj:
Fixes #21 .

Nice side effect: improves the `show_path_to_root` method.


go.py
```diff
@@ -299,6 +299,7 @@ def __init__(self, board=None, n=0, komi=7.5, caps=(0, 0),
             Should satisfy next_pos.board - next_pos.board_deltas[0] == pos.board
         to_play: BLACK or WHITE
         '''
+        assert(type(recent)) is tuple
```
brilee: can drop the outer ()

strategies.py
```diff
@@ -26,10 +26,10 @@
 
 import go
 
-MAX_GAME_DEPTH = int(go.N * go.N * 1.25)
 # When to do deterministic move selection.  ~30 moves on a 19x19, ~8 on 9x9
 TEMPERATURE_CUTOFF = int((go.N * go.N) / 12)
 
+MAX_DEPTH = (go.N ** 2) * 1.4  # ~500 moves for 19x19, ~110 for 9x9
```
brilee: would be nice to cast this to int and give exact number in the comment, just in case somebody goes grepping for "why did my game end at move 505"

strategies.py
```diff
@@ -165,9 +165,10 @@ def tree_search(self, num_parallel=None):
             failsafe += 1
             leaf = self.root.select_leaf()
             if self.verbosity >= 4:
-                self.show_path_to_root(leaf)
+                path = self.show_path_to_root(leaf)
+                print(path)
             # if game is over, override the value estimate with the true score
-            if leaf.position.is_game_over():
+            if leaf.position.is_game_over() or leaf.position.n >= MAX_DEPTH:
```
brilee: Oh man, this is a weird hidden dependency. So, this prevents the MCTS tree search from expanding beyond the max depth, not because it's explicitly forbidden in select_leaf, but because this if clause prevents is_expanded from ever getting set to true. This is some nonintuitive behavior that should probably get called out in a comment somewhere.

tests/test_mcts.py
```diff
@@ -124,8 +124,8 @@ def test_backup_incorporate_results(self):
         self.assertEqual(leaf.N, 2)
         self.assertEqual(leaf2.N, 1)
         # average of 0, -1, -0.2
+        self.assertAlmostEqual(leaf.Q, root.child_Q[leaf.fmove])
```
brilee: I had left all these tests in when doing my child_X refactor just to double check I didn't break anything, but we could probably nuke all these redundant tests now.

tests/test_strategies.py
```diff
@@ -198,6 +197,17 @@ def test_ridiculously_parallel_tree_search(self):
             player.tree_search(num_parallel=50)
         self.assertNoPendingVirtualLosses(player.root)
 
+    def test_long_game_tree_search(self):
+        player = MCTSPlayerMixin(DummyNet())
+        player.initialize_game(SEND_TWO_RETURN_ONE)
+        player.root.position.n=108
```
brilee: can we make this "= MAX_DEPTH - 1" or whatever the relevant relationship is?

tests/test_strategies.py
```diff
@@ -198,6 +197,17 @@ def test_ridiculously_parallel_tree_search(self):
             player.tree_search(num_parallel=50)
         self.assertNoPendingVirtualLosses(player.root)
 
+    def test_long_game_tree_search(self):
```
brilee: I think an ideal test would be one that's up against the move limit, where placing a stone in enemy territory is the "correct" move because of the end-of-game limit.

strategies.py
```diff
@@ -165,9 +165,10 @@ def tree_search(self, num_parallel=None):
             failsafe += 1
             leaf = self.root.select_leaf()
             if self.verbosity >= 4:
-                self.show_path_to_root(leaf)
+                path = self.show_path_to_root(leaf)
```
Kashomon: nit: just print(self.show_...)?

strategies.py
Kashomon: 500 Seems kinda small. I've definitely seen early-phase minigo games last longer than 500 moves. Maybe it should be a bit larger, like 1000?

brilee: AGZ used 2 * N^2. The point isn't to let all games gracefully finish, though - it's explicitly cutting games off to save computation, cull the weird tail-end of such games from the training data, and as a failsafe for weird infinite loop situations. (I wouldn't be surprised if a longer game end led to worse training because one player accidentally suicides his whole-board group.)

Kashomon: Why don't we just use 2*N^2 then, if AGZ used that ?

amj: Because it really affects how slow things are at the beginning.  A lot  I'm happy to go ahead and say there aren't really meaningful go games longer than 400 moves, and 500 moves vs 720 moves is basically like a whole game's worth of wasted work.

strategies.py
amj: Done

tests/test_strategies.py
amj: good idea.  Done.

tests/test_strategies.py
amj: Done!

tests/test_mcts.py
amj: i don't think there's anything hurt by having them, and actually they were pretty helpful for me to look at when reviewing the vlosses code.  I'm happy with them.

strategies.py
amj: Done

strategies.py
amj: I don't know if it's that unintuitive -- go.py has the rules of stones on boards.  strategies.py (which should probably be renamed to players.py or something) has all the idiosyncracies about how humans and computers use boards and stones.

go.py
amj: done.

amj:
Ok thanks, PTAL!

strategies.py
```diff
@@ -190,16 +190,23 @@ def is_done(self):
...
     def show_path_to_root(self, node):
         pos = node.position
+        diff = node.position.n - self.root.position.n
         if len(pos.recent) == 0:
             return
-        moves = list(map(coords.to_human_coord,
-                         [move.move for move in pos.recent[self.root.position.n:]]))
+        fmt = lambda move: "{}-{}".format('b' if move.color == 1 else 'w',
+                                          coords.to_human_coord(move.move))
+        path = " ".join(list(map(fmt,
```
brilee: this is actually more verbose than the comprehension:
`path = " ".join(fmt(move) for move in pos.recent[-diff:])`

tests/test_strategies.py
```diff
@@ -198,6 +210,25 @@ def test_ridiculously_parallel_tree_search(self):
...
+            komi=2.5,
+            ko=None,
+            recent=(go.PlayerMove(go.BLACK, (0, 1)),
+                    go.PlayerMove(go.WHITE, (0, 8))),
+            to_play=go.BLACK
+        )
+        player.initialize_game(endgame)
+
+        # Test that an almost complete game
+        for i in range(100):
```
brilee: bump this down? I noticed that lots of tree searching in tests can take like a second to run (a whole second!). It should only take like 10*8 searches to figure this test out.

amj: That won't actually work to test the behavior, unfortunately.  With all the empty spots in the new test, and an equal policy for all points, it literally tries every permutation, so even at only N-2 moves out, we won't actually ground out.  I will rework the board position though standby

strategies.py
amj: quite a bit more verbose actually.  Done!

tests/test_strategies.py
amj: Done

Issue #1, opened 2018-01-29
cos-nvidia-driver-installer needs updating to k8s 1.9, TF 1.5


amj:
doesn't work for TF 1.5
might not be correct to begin with :)
So GPUs on kubernetes are changing how they are used and consumed in 1.8 (https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/)

Our current daemonset job also installs the drivers AND exposes them for scheduling -- this seems to conflict or otherwise trample on the drivers in the base NVIDIA image, and means that the cuda 9 images don't actually work.

So the work here is to:

figure out what is needed to get a working device-plugin for k8s 1.8 that exposes our GPU
get that device-plugin sorted with its nvidia drivers
get TF1.5 working
update the docs to reflect this final solution.

amj:
So full list of upgrades to make
 - py3.6
 - TF 1.5
 - k8s 1.9
 - Cuda 9

PR #48, opened 2018-02-03
Clean up the cluster scripts

Kashomon:
- Add a K8S Version option (and use 1.9.2)
- Switch to Beta track with 1.9.2 EAP
- Add a Zone option
- Explicitly set PROJECT/ZONE
- Make the destroy script less scary
- Make a cluster-up-gpu-small for test-size clusters

### DIFF
--- a/cluster/cluster-up
+++ b/cluster/cluster-up
@@ -13,6 +13,9 @@ echo "Using K8S Version:  ${K8S_VERSION}"
 gcloud beta container clusters create \
   --num-nodes 4 \
   --machine-type n1-standard-8 \
+  # Intel Broadwells are a bit newer, and can provide up to a 30% speedup,
+  # since we're so CPU bound.
+  --min-cpu-platform "Intel Broadwell" \
   --disk-size 30 \
   --zone $ZONE \
   --project $PROJECT \
--- a/cluster/cluster-up-gpu
+++ b/cluster/cluster-up-gpu
@@ -11,7 +11,7 @@ echo "Using K8S Version:  ${K8S_VERSION}"
 
 # Create a Kubernetes cluster. This setup is designed for creating large clusters.
 gcloud beta container clusters create \
-  --num-nodes 5 \
+  --num-nodes 600 \
   --accelerator type=nvidia-tesla-k80,count=1 \
   --machine-type n1-standard-2 \
   --disk-size 20 \
--- a/cluster/cluster-up-gpu
+++ b/cluster/cluster-up-gpu
@@ -4,20 +4,26 @@ source ./common.sh
 
 set -e
 
+echo "Using Project:      ${PROJECT}"
+echo "Using Zone:         ${ZONE}"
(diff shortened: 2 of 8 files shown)


cluster/cluster-up-gpu
```diff
@@ -4,20 +4,26 @@ source ./common.sh
...
-  --enable-kubernetes-alpha \
+gcloud beta container clusters create \
   --num-nodes 5 \
   --accelerator type=nvidia-tesla-k80,count=1 \
   --machine-type n1-standard-2 \
   --disk-size 20 \
   --preemptible \
-  --cluster-ipv4-cidr=10.0.0.0/10 \
-  --zone=asia-east1-a \
+  --cluster-ipv4-cidr=10.0.0.0/10 \ # Allocate a big CIDR block for a big cluster
```
amj: this is only needed for clusters with > 1008 nodes; maybe make the comment specific to that?

cluster/cluster-up-gpu-small
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
amj: what's the logic on small/large?  seems like we should clarify just how we expect folks to use the repo.  Is this an example code?  is this code that all the people who run clusters check in?

it smacks of the eternal configuration vs code in VCS debates, i guess.  

What do we get from having a second cluster-up script?

cluster/deploy-gpu-player.sh
```diff
@@ -16,4 +16,6 @@
 
 source ./common.sh
 
+# TODO(kashomon): envsubst doesn't exist for OSX. needs to be brew-installed
+# via gettext. Should probably warn the user about that.
```
amj: in the README?  Otherwise this comment is pretty much all that you need :)

Kashomon: Added it directly in the command

cluster/cluster-up-gpu-small
Kashomon: I've simplified this from large cluster. In particular, the GPUS are non-preemptible.

> it smacks of the eternal configuration vs code in VCS debates, i guess.

I don't have the context here.

cluster/cluster-up
```diff
@@ -4,18 +4,22 @@ source ./common.sh
...
+echo "Using Project:      ${PROJECT}"
+echo "Using Zone:         ${ZONE}"
+echo "Using Cluster Name: ${CLUSTER_NAME}"
+echo "Using K8S Version:  ${K8S_VERSION}"
+
 # Create a Kubernetes cluster
 gcloud beta container clusters create \
   --num-nodes 4 \
   --machine-type n1-standard-8 \
-  --min-cpu-platform "Intel Broadwell" \
```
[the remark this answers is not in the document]
Kashomon: Ack. I'll add it back in

Kashomon:
Addresses https://github.com/tensorflow/minigo/issues/1
- Updates cluster to 1.9.2

PR #49, opened 2018-02-03
Clean up evaluation.py and add a bunch of comments.

Kashomon:
Changes should be safe, but a bit hard to tell since I don't think there are any tests for evaluation.py.

- Remove unused imports
- Clean up some line-length problems
- Add comments
- Simplify one unnecessary list comprehension
- Rename some variables

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -12,47 +12,66 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import go
-import time
-import numpy as np
+"""Evulation plays games between two opponents."""
 
-from dual_net import DualNetwork
+import time
+
 from gtp_wrapper import MCTSPlayer
 
 
 def play_match(black_net, white_net, games, readouts, verbosity):
+    """Plays matches between two players.
+
+    black_net: Instance of minigo.DualNetwork, a wrapper around a tensorflow
+        convolutional network.
+    white_net: Instance of the minigo.DualNetwork.
+    games: number of games to play. We play all the games at the same time.
+    readouts: number of readouts to perform for each step in each game.
+    """
+
+    # For n games, we create lists of n black and n white playerss
     black_players = [MCTSPlayer(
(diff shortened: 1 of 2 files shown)


evaluation.py
```diff
@@ -12,47 +12,66 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import go
+"""Evulation plays games between two opponents."""
```
amj: typo

amj: also that's not quite what it does -- plays games between two separate nets, reloading the weights between iterations

evaluation.py
```diff
@@ -12,47 +12,66 @@
...
-        p2.initialize_game()
+    # The number of moves that have been played
+    num_moves = 0
+
+    for black, white in player_pairs:
+        black.initialize_game()
+        white.initialize_game()
 
+    # The heart of the game-playing loop. Each iteration through the while loop
+    # is plays one move for each player. That means we:
```
amj: moar typo

evaluation.py
```diff
@@ -12,47 +12,66 @@
...
 from gtp_wrapper import MCTSPlayer
 
 
 def play_match(black_net, white_net, games, readouts, verbosity):
+    """Plays matches between two players.
+
+    black_net: Instance of minigo.DualNetwork, a wrapper around a tensorflow
+        convolutional network.
+    white_net: Instance of the minigo.DualNetwork.
+    games: number of games to play. We play all the games at the same time.
```
amj: Now that we have vlosses, it makes sense to refactor this whole thing to just play a single game instead of the root-level parallelism.  it'll probably simplify the code a lot.

evaluation.py
Kashomon: Done

evaluation.py
Kashomon: Added a TODO

evaluation.py
Kashomon: Fixed

PR #52, opened 2018-02-04
Fix bad result not respecting resign

brilee:
Plz look over very carefully and let me know ifyou have ideas on how to test this better.


selfplay_mcts.py
```diff
@@ -60,14 +59,13 @@ def play(network, readouts, resign_threshold, verbosity=0):
             print(player.root.position)
             print(player.root.describe())
 
-        # Sets is_done to be True if player.should resign.
-        if player.should_resign():  # TODO: make this less side-effecty.
+        if player.should_resign():
+            player.set_result(-1 * player.root.position.to_play, was_resign=True)
```
amj: nice, this is much cleaner.  Thank you.

strategies.py
```diff
@@ -208,37 +196,32 @@ def show_path_to_root(self, node):
...
-        return False
+        return self.root.Q_perspective < self.resign_threshold
 
-    def make_result_string(self, pos):
-        if abs(self.result) == 2:
-            res = "B+Resign" if self.result == 2 else "W+Resign"
+    def set_result(self, winner, was_resign):
+        self.result = winner
+        if was_resign:
+            string = "B+R" if winner == go.BLACK else "W+R"
```
amj: sort of raises a larger question that we don't have to resolve on this PR, but: will we ever run with integer komi?

brilee: I think the big bridge to cross there is not even the question of integer komi, but how to make a neural net handle multiple komi types. Agreed that it's sort of weird that ties aren't possible here.

amj: did you see  https://arxiv.org/abs/1705.10701 ?

tests/test_strategies.py
```diff
@@ -277,3 +277,38 @@ def test_only_check_game_end_once(self):
         player.tree_search()
         # check that we didn't visit the pass node any more times.
         self.assertEqual(player.root.child_N[pass_move], 1)
+
```
amj: these tests look great and i'm happy to LGTM right now.  I think some tests for selfplay_mcts.py would also be useful to make sure the should_resign and set_result calls are happening correctly.

strategies.py
brilee: Yep, i've read that paper.

amj: what did you think of it?

PR #54, opened 2018-02-04
Upgrade a bunch of Deps

Kashomon:
- Upgrade Tensorflow to 1.5
- Upgrade CUDA to 9.1

Not upgrading python because we're on Ubuntu 16.04 and python3.6 isn't part of the default apt packages for this version of ubuntu.


amj:
The python version is only explicitly specified in the dockerfiles, either by the base image or installed on top.  Our VMs etc would need to be updated separately.  (assuming we start requiring 3.6 for type hints)

Kashomon:
Going to punt on python 3.6 because our Cuda version is tied to Ubuntu 16.04. Python 3.6 isn't included in the standard apt packages in 16.04 (have to wait until 16.10), so it's annoying to install. See cuda packages @ https://hub.docker.com/r/nvidia/cuda/

Kashomon:
We could use Ubunto 17.04 but that looks experimental?

Kashomon:
Also, going to use Cuda 9.1 since that's available -- might as well. Uses the same base Ubuntu image (16.04)

Kashomon:
I'm

Kashomon:
oops. Misclick

cluster/Dockerfile
```diff
@@ -2,7 +2,36 @@ FROM ubuntu:16.04
...
+        python3-numpy \
+        python3-dev \
+        python3-pip \
+        python3-wheel \
+        python3-setuptools \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN curl -v -L -O https://github.com/GoogleCloudPlatform/gcsfuse/releases/download/v0.23.0/gcsfuse_0.23.0_amd64.deb && \
```
amj: do we still use GCSfuse?  I think this can go away, as can the uuid-runtime

cluster/gpu-provision-daemonset.yaml
```diff
@@ -1,48 +1,82 @@
+# Copyright 2017 Google Inc. All rights reserved.
```
amj: Do we need these for YAML?

Kashomon: Dunno. I just copied this whole hog via 

`kubectl create -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/k8s-1.9/nvidia-driver-installer/cos/daemonset-preloaded.yaml`

cluster/Dockerfile
Kashomon: Ah, cool. I was making the CPU dockerfile consistent with the GPU settings. I'll remove the fuse stuff from both. And the uuid-runtime

amj: Oh!  does the CPU one still build?  The CPU one is a bit more complicated because it has to build tensorflow from source in the container to have the CPU optimizations, whereas the GPU one gets all the magic from `pip install tensorflow-gpu`

Kashomon:
Note: https://github.com/tensorflow/minigo/issues/58 -- we're not currently compiling TF-CPU from scratch

Kashomon:
Aw, this is currently failing: 

standard_init_linux.go:178: exec user process caused "exec format error"

More work to do.

Kashomon:
Maybe it's https://github.com/moby/moby/issues/23865?

cluster/Makefile
```diff
@@ -1,8 +1,10 @@
 # Version and project can be overridden with the following:
 #  make target VERSION=0.XY PROJECT=foo-project
 
-VERSION?=0.13
+VERSION_TAG?=0.13
```
amj: Is this double defined with the common.sh definition below?

cluster/gpu-player.yaml
```diff
@@ -21,7 +21,7 @@ spec:
             topologyKey: kubernetes.io/hostname
       containers:
       - name: player
-        image: gcr.io/$PROJECT/minigo-gpu-player:0.10
+        image: gcr.io/$PROJECT/$GPU_PLAYER_CONTAINER:$VERSION_TAG
```
amj: This seems like a good way to inadvertently push a new version when you only want to change e.g. the parallelism.

cluster/staging/requirements.txt
```diff
@@ -0,0 +1,12 @@
+argh
```
amj: is this supposed to be here?

cluster/gpu-player.yaml
Kashomon: I think this is MUCH better and safer than the original approach. Now, all you need to do is say `export VERSION_TAG=0.10` and then that tag is used everywhere consistently -- whether via `make` or deploying a cluster.

More likely, I can see people forgetting to edit the YAML and then pushing a downgrade for their cluster.

cluster/Makefile
Kashomon: Sort of? This sets an override-able default. I should probably just not set it here though.

cluster/staging/requirements.txt
amj: (i mean, this whole file)

cluster/gpu-player.yaml
amj: Ok, i definitely agree this is better :)
But there's still the problem that it's defined with a default in the Makefile *and* in common.sh.  If i don't run common.sh to export things, then i'm gonna end up with the wrong default.

amj:
LGTM! :)

Kashomon:
Fixes: https://github.com/tensorflow/minigo/issues/1

cluster/Makefile
Kashomon: Removed

PR #57, opened 2018-02-05
update selfplay to write 5% of games to a holdout directory

amj:
- Changes the signature on preprocessing.make_dataset_from_selfplay to return the tf_examples instead of write a file
- sets a default holdout directory and makes the appropriate changes in rl_loop.py
- currently not an argument, nor is the amount of holdout a parameter.  Should it be?


preprocessing.py
```diff
@@ -161,15 +161,15 @@ def get_input_tensors(batch_size, tf_records, num_repeats=None,
 # End-to-end utility functions
 
 
-def make_dataset_from_selfplay(data_extracts, tf_record):
+def make_dataset_from_selfplay(data_extracts):
     '''
+    Returns a set of tf.Examples.
```
brilee: return an iterable of tf.Example

rl_loop.py
```diff
@@ -28,6 +28,7 @@
 BASE_DIR = "gs://{}".format(BUCKET_NAME)
 MODELS_DIR = os.path.join(BASE_DIR, 'models')
 SELFPLAY_DIR = os.path.join(BASE_DIR, 'games')
```
brilee: While we're rejiggering our folder naming scheme, can we rename the selfplay dir? "games" is pretty nondescriptive.

main.py
```diff
@@ -138,8 +139,15 @@ def selfplay(
     with gfile.GFile(os.path.join(output_sgf, '{}.sgf'.format(output_name)), 'w') as f:
         f.write(player.to_sgf())
 
-    fname = os.path.join(output_dir, "{}.tfrecord.zz".format(output_name))
-    preprocessing.make_dataset_from_selfplay(game_data, fname)
+    tf_examples = preprocessing.make_dataset_from_selfplay(game_data)
+
+    # Hold out 5% of games for evaluation.
```
brilee: make 0.05 a flag?

rl_loop.py
amj: Done.

preprocessing.py
amj: Done.

main.py
amj: gilding the lily a bit but done :)

main.py
```diff
@@ -118,10 +118,12 @@ def evaluate(
 def selfplay(
         load_file: "The path to the network model files",
         output_dir: "Where to write the games"="data/selfplay",
+        holdout_dir: "Where to write the games"="data/holdout",
```
Kashomon: is this written to local disk?

amj: It's written to wherever you pass it.  This can be a gs:// path

PR #66, opened 2018-02-06
Lots more GPU / CPU K8S cluster updates

Kashomon:
- Make parallelism a var since we can calculate it.
- Make the default GPU script the small one to match the CPU one
- Move the docs to cluster/README.md
- Use beta GPUs instead of alpha GPUs since that's what API we're using.


cluster/README.md
```diff
@@ -0,0 +1,258 @@
+# Kubernetes Cluster and Image Management
+
+As you might notice, playing games is fairly slow.  One way to speed up playing
```
amj: Now without a segue from the previous paragraph ;)  Maybe rephrase

cluster/cluster-up-cpu
```diff
@@ -9,12 +9,17 @@ echo "Using Project:      ${PROJECT}"
 echo "Using Zone:         ${ZONE}"
 echo "Using Cluster Name: ${CLUSTER_NAME}"
 echo "Using K8S Version:  ${K8S_VERSION}"
+echo "Number of Nodes:    ${NUM_K8S_NODES}"
+
+export PARALLELISM="$((3 * ${NUM_K8S_NODES}))"
```
amj: Eep.  Having parallelism defined as an env var is great when N is integers greater than one -- how's your bash math with floating points?  (For GPU, parallelism is usually 98% or 95% of number of nodes)

cluster/cluster-up-cpu
```diff
@@ -9,12 +9,17 @@ echo "Using Project:      ${PROJECT}"
...
+
+check_gcloud_exists
 
 # Create a Kubernetes cluster
 # Note, we require Intel Broadwells since they are a bit newer, and can provide
 # up to a 30% speedup, since we're so CPU bound.
 gcloud beta container clusters create \
-  --num-nodes 3 \
+  --num-nodes $NUM_K8S_NODES \
   --machine-type n1-standard-8 \
```
amj: if these are standard-8's, parallelism should probably be higher, in 6-8

README.md
```diff
@@ -319,260 +319,7 @@ Additionally, you can follow along with the training progress with TensorBoard -
...
-  ```
-
-* Monitoring how long it's taking the daemonset to install the nvidia driver on
-  your nodes
-  ```
-  kubectl get no -w -o yaml | grep -E 'hostname:|nvidia-gpu'
-  ```
-
-
-If you've run rsync to collect a set of SGF files (cheatsheet: `gsutil -m cp -r
```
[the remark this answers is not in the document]
Kashomon: Sounds totally reasonable. This doc move was pretty coarse.

cluster/README.md
Kashomon: Removed segue

cluster/cluster-up-cpu
Kashomon: I think Parallelism is an int in the Kubernetes job api

cluster/cluster-up-cpu
Kashomon: Changed to 8

Kashomon: (I'll do some test also)

PR #68, opened 2018-02-07
update Dockerfile to build tensorflow from source, optimized for extended CPU instructions

amj:


cluster/Dockerfile
```diff
@@ -16,17 +16,57 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
         zlib1g-dev \
-        python3 \
-        python3-numpy \
-        python3-dev \
-        python3-pip \
-        python3-wheel \
-        python3-setuptools \
+        openjdk-8-jdk \
+        openjdk-8-jre-headless \
+				python3 \
```
Kashomon: funny indentation here

cluster/Dockerfile
```diff
@@ -16,17 +16,57 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+# Running bazel inside a `docker build` command causes trouble, cf:
+#   https://github.com/bazelbuild/bazel/issues/134
+# The easiest solution is to set up a bazelrc file forcing --batch.
+RUN echo "startup --batch" >>/etc/bazel.bazelrc
+# Similarly, we need to workaround sandboxing issues:
+#   https://github.com/bazelbuild/bazel/issues/418
+RUN echo "build --spawn_strategy=standalone --genrule_strategy=standalone" \
+    >>/etc/bazel.bazelrc
+# Install the most recent bazel release.
+ENV BAZEL_VERSION 0.5.4
```
Kashomon: this is an ancient version of bazel. Let's use a more recent version. Like 0.10.0
https://github.com/bazelbuild/bazel/releases

requirements-cpu.txt
```diff
@@ -3,7 +3,6 @@ numpy>=1.14.0
 pygtp>=0.4
 sgf==0.5
 tqdm>=4.17
-tensorflow>=1.5,<1.6
```
Kashomon: This is a bit awkward for local work. If you pip install first, shouldn't the bazel install override the tensorflow settings?

Here's another alternative: We make a requirements-base.txt. This has the shared deps from requirements-cpu and requirements-gpu.  Then we just install the base in the dockerfile and our instructions for local development are:

- install requirements-base.txt
- install either requirements-cpu.txt or requirements-gpu.txt

cluster/common.sh
```diff
@@ -24,7 +24,7 @@ export CPU_PLAYER_CONTAINER=${CPU_PLAYER_CONTAINER:-"minigo-player"}
 # So, we prefix the project name to avoid collisions
 #
 # For more details, see https://cloud.google.com/storage/docs/best-practices
-export BUCKET_NAME=${BUCKET_NAME:-"${PROJECT}-minigo-v2-${BOARD_SIZE}"}
+export BUCKET_NAME=${BUCKET_NAME:-"${PROJECT}-minigo-v3-${BOARD_SIZE}"}
```
Kashomon: maybe [v3] should also be an env variable?

Seems to me we should use more robust versions for our minigo versions. at least vMajor-Minor

cluster/Dockerfile
```diff
@@ -16,17 +16,57 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+# --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=0" \
+RUN ln -s /usr/bin/python3 /usr/bin/python  && tensorflow/tools/ci_build/builds/configured CPU \
+    bazel build -c opt --copt=-march="haswell" \
+        tensorflow/tools/pip_package:build_pip_package --verbose_failures && \
+    bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/pip && \
+    pip3 --no-cache-dir install --upgrade /tmp/pip/tensorflow-*.whl && \
+    rm -rf /tmp/pip && \
+    rm -rf /root/.cache
+
+WORKDIR /
```
Kashomon: is this necessary?

cluster/Dockerfile
Kashomon: Ah, this should be Bazel 0.8.0 -- https://www.tensorflow.org/install/install_sources#common_installation_problems

cluster/Dockerfile
amj: the workdir thing?  yes - we had previously set a diff workdir to build tf in.

requirements-cpu.txt
amj: negative -- if you pip install first, the final line of the install is `pip3 --no-cache-dir install --upgrade /tmp/pip/tensorflow-*.whl`, which will die as there is already a version of tensorflow installed.

Kashomon: Did you see the second part of my message?

Kashomon: Also, you could do pip3 uninstall I think

cluster/Dockerfile
amj: fixed.

cluster/Dockerfile
amj: Done.

cluster/common.sh
amj: there's probably a bit of cleanup here.  I don't think everything in common.sh needs to have a default, and sometimes the 'see if it's set and don't change it if it is' isn't quite the right behavior when you want to source common.sh and have everything get updated.  Let's chat about it sometime.

requirements-cpu.txt
amj: I'm not sure why this is awkward for local work.  There's kind of an assumption that after `pip install -r requirements.txt` everything is ready, but that's not really necessarily true.

Anyway, if the two -cpu.txt and -gpu.txt files only contain the line 'tensorflow' vs 'tensorflow-gpu', i'd probably lean away from it in favor of having the dockerfile do 'pip uninstall'

amj: Oh, and actually the problem is the reverse -- with the wheel installed, `pip install -r requirements-cpu.txt` fails.

i'd rather have requirements-cpu.txt track the modules needed for the docker image, and for local  development -- everyone's kind of on their own w.r.t. TF/nvidia/openCL/anaconda/whatever

Kashomon: > i'd rather have requirements-cpu.txt track the modules needed for the docker image, and for local development

If that's the case, I'd rather we have a requirements-local-cpu.txt (I can add that later). It creates a lot of friction for new users to say: Hey, here's the start of something and then debug until you figure it out. I'd rather we have a story we think works rather than one we're sure needs extra fidling

cluster/Dockerfile
Kashomon: not seeing changes? Push new version?

requirements-cpu.txt
Kashomon: Or, we could have a requirements-cpu-docker.txt

PR #69, opened 2018-02-07
Small changes to dockerfiles

Kashomon:
- Use the CUDA runtime rather than devel (I don't think the devel deps are needed and are much heavier).
- Use CMD instead of ENTRYPOINT. They're mostly equivalent except that for the former I can run 

```
docker run -ti gcr.io/jhoak-mini/minigo-gpu-player:0.14 /bin/sh
```

for testing.

### DIFF
--- a/cluster/Dockerfile.gpu
+++ b/cluster/Dockerfile.gpu
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04
+FROM nvidia/cuda:9.0-cudnn7-runtime-ubuntu16.04
 
 RUN apt update && apt-get install -y --no-install-recommends \
         build-essential \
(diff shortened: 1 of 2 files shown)


cluster/Dockerfile.gpu
```diff
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04
+FROM nvidia/cuda:9.0-cudnn7-runtime-ubuntu16.04
```
amj: so does this GPU file actually work atm?

Kashomon: I think it does, oddly enough. I built this from my desktop, and it appears to work.

I think it might *not* work if I build it from my macbook. More studies need to be done.

PR #70, opened 2018-02-07
Update the Bucket ACLs so that the service account can read the models

Kashomon:
This is really only a problem for bootstrap, but it is annoying.

### DIFF
--- a/cluster/utils.sh
+++ b/cluster/utils.sh
@@ -74,8 +74,13 @@ function create_service_account_key() {
     gcloud iam service-accounts create $SERVICE_ACCOUNT --project=$PROJECT
   fi
 
-  # Grant it read/write permissions on our bucket. Should be safe to do multiple times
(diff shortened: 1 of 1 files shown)


cluster/utils.sh
```diff
@@ -74,8 +74,13 @@ function create_service_account_key() {
     gcloud iam service-accounts create $SERVICE_ACCOUNT --project=$PROJECT
   fi
 
-  # Grant it read/write permissions on our bucket. Should be safe to do multiple times
-  gsutil acl ch -u "${SERVICE_ACCOUNT_EMAIL}":W gs://${BUCKET_NAME}
+  # Make sure the service account can actually read the existing model entries
+  gsutil ls gs://$BUCKET_NAME/models >/dev/null 2>&1 && {{
```
amj: isn't the acl call idempotent?

PR #74, opened 2018-02-08
Make training job 'pause' if there aren't enough new games.

amj:
Adds function to rl_loop.py to check number of games finished since last training was complete.  Check that number before beginning the next training job.  It should ideally be ~100% the size of the generation, but if not, sleep a bit before training again.

### DIFF
--- a/rl_loop.py
+++ b/rl_loop.py
@@ -20,6 +20,7 @@ import logging
 import os
 import main
 import shipname
+import time
 from utils import timer
 from tensorflow import gfile
 
--- a/rl_loop.py
+++ b/rl_loop.py
@@ -20,6 +20,7 @@ import logging
 import os
 import main
 import shipname
+import sys
 import time
 from utils import timer
 from tensorflow import gfile
@@ -34,6 +35,13 @@ HOLDOUT_DIR = os.path.join(BASE_DIR, 'data/holdout')
 SGF_DIR = os.path.join(BASE_DIR, 'sgf')
 TRAINING_CHUNK_DIR = os.path.join(BASE_DIR, 'data', 'training_chunks')
 
+"""
+This is dependent on about how long it takes a new model to train and how long
+the cluster can play a new generation's worth of games.  if N is the desired
+number of games per generation, play_time is the time required for the cluster
+to play N games, and train_time is the time taken to train a new model, then this number should be 
+N * max(1, (train_time / play_time))
+"""
 GAMES_BEFORE_TRAINING = 4000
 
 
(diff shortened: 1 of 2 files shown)


rl_loop.py
```diff
@@ -33,6 +34,8 @@
 SGF_DIR = os.path.join(BASE_DIR, 'sgf')
 TRAINING_CHUNK_DIR = os.path.join(BASE_DIR, 'data', 'training_chunks')
 
+GAMES_BEFORE_TRAINING = 4000
```
Kashomon: I'm big on comments, of course. Can we add one here?

rl_loop.py
```diff
@@ -49,20 +52,36 @@ def print_flags():
                     for flag, value in flags.items()))
 
 
+def get_models():
```
Kashomon: `+` method doc

rl_loop.py
```diff
@@ -49,20 +52,36 @@ def print_flags():
...
-    model_filenames = [os.path.basename(m) for m in all_models]
-    model_numbers_names = [
-        (shipname.detect_model_num(m), shipname.detect_model_name(m))
-        for m in model_filenames]
+    model_numbers_names = get_models()
     latest_model = sorted(model_numbers_names, reverse=True)[0]
     return latest_model
 
 
+def enough_games_to_train():
```
Kashomon: `+` method doc

rl_loop.py
```diff
@@ -49,20 +52,36 @@ def print_flags():
...
+    model_numbers_names = get_models()
+    if len(model_numbers_names) <= 2:
```
Kashomon: if the number of models is less then 2, then we always have enough models to train? This is a little unintuitive, so a comment would be helpful

rl_loop.py
```diff
@@ -104,6 +123,8 @@ def gather():
 
 
 def train(logdir=None):
+    while not enough_games_to_train():
+        time.sleep(60 * 5)
```
Kashomon: How about we make the sleep-time configurable or a constant

rl_loop.py
amj: there's an edge case if you only have the first model.  So either you bootstrap, and let your cluster spin for a little bit, or you just shrug and go ahead and train the first model with too little data and let it catch up for all models > 2

amj: (comment added)

rl_loop.py
amj: actually, the right thing to do is exit -- otherwise the gather step won't actually be repeated.

rl_loop.py
amj: Done!

rl_loop.py
amj: Done!

rl_loop.py
amj: Done!

PR #75, opened 2018-02-08
Pausable worker

amj:
Depends on #74.

Checks the number of games played for the latest model at the beginning of selfplay.  Should it do exponential backoff?


rl_loop.py
```diff
@@ -82,6 +103,10 @@ def bootstrap():
 
 def selfplay(readouts=1600, verbose=2, resign_threshold=0.99):
     _, model_name = get_latest_model()
+    games = gfile.Glob(os.path.join(SELFPLAY_DIR, m, '*.zz'))
```
brilee: m?

rl_loop.py
```diff
@@ -49,20 +54,36 @@ def print_flags():
...
-        (shipname.detect_model_num(m), shipname.detect_model_name(m))
-        for m in model_filenames]
+    model_numbers_names = get_models()
     latest_model = sorted(model_numbers_names, reverse=True)[0]
     return latest_model
 
 
+def enough_games_to_train():
+    model_numbers_names = get_models()
+    if len(model_numbers_names) <= 2:
```
brilee: Not sure I understand this condition

rl_loop.py
```diff
@@ -49,20 +54,36 @@ def print_flags():
...
+        return True
+
+    second_latest_model = sorted(model_numbers_names, reverse=True)[1]
```
brilee: what does second latest model have to do with whether you should train?

rl_loop.py
amj: model_name, whoops.

rl_loop.py
amj: with just the bootstrap model, you don't have a 2nd-most-recent... so just go ahead and train one on whatever you've got and then relax.

rl_loop.py
amj: fixed

rl_loop.py
```diff
@@ -33,6 +34,16 @@
 SGF_DIR = os.path.join(BASE_DIR, 'sgf')
 TRAINING_CHUNK_DIR = os.path.join(BASE_DIR, 'data', 'training_chunks')
 
+"""
+This is dependent on about how long it takes a new model to train and how long
```
Kashomon: s/This/Games before training 
to make antecedents clear.

rl_loop.py
```diff
@@ -49,20 +60,48 @@ def print_flags():
                     for flag, value in flags.items()))
 
 
+def get_models():
```
Kashomon: merge? I thought this was already added?

amj: This branch is based off of #74, but the diffbase is master, so it's not actually added.  Sorry, i know it makes the diffs hard.

brilee: Can you rebase this commit onto a separate branch off master? This has changes for selfplay workers which I think are useful on their own and should be merged in, regardless of what we do with #74.

amj: Done, thanks, PTAL :)

rl_loop.py
amj: from the POV of the trainer, the most recent model is brand new and won't have games for potentially many minutes.  If we wait for those many minutes, we will be that many minutes "behind" how fast we could possibly train.  So the 'second most recent model' at the point the trainer is looking is actually the most recent model we've got finished games from.

PR #80, opened 2018-02-10
Add `validate` to DualNetworkTrainer.  Add command to main.py

amj:
@brilee  I'm uncomfortable with the amount of copypasta shared between validate() and train(), would love it if you could let me know if that's not actually a code smell here.

### DIFF
--- a/dual_net.py
+++ b/dual_net.py
@@ -41,14 +41,17 @@ EXAMPLES_PER_GENERATION = 2000000
 # How many positions can fit on a graphics card. 256 for 9s, 16 or 32 for 19s.
 TRAIN_BATCH_SIZE = 256
 
+# The shuffle buffer size determines the max 'distance' an example ends up from
+# where it started.
+SHUFFLE_BUFFER_SIZE = int(4*1e6)
+
 
 class DualNetworkTrainer():
-    def __init__(self, save_file=None, **hparams):
+    def __init__(self, save_file=None, logdir=None, **hparams):
         self.hparams = get_default_hyperparams(**hparams)
         self.save_file = save_file
+        self.logdir = logdir
         self.sess = tf.Session(graph=tf.Graph())
-        # TODO: eval stuff
-        # self.eval_logdir = os.path.join(model_dir, 'logs', 'eval')
 
     def initialize_weights(self, init_from=None):
         """Initialize weights from model checkpoint.
@@ -88,15 +91,17 @@ class DualNetworkTrainer():
             sess.run(tf.global_variables_initializer())
             tf.train.Saver().save(sess, self.save_file)
 
-    def train(self, tf_records, init_from=None, logdir=None, num_steps=None,
+    def train(self, tf_records, init_from=None, num_steps=None,
               logging_freq=100, verbosity=1):
+        logdir = self.logdir + '-train' if self.logdir is not None else None
+
         def should_log(i):
             return logdir is not None and i % logging_freq == 0
         if num_steps is None:
             num_steps = EXAMPLES_PER_GENERATION // TRAIN_BATCH_SIZE
         with self.sess.graph.as_default():
             input_tensors = preprocessing.get_input_tensors(
-                TRAIN_BATCH_SIZE, tf_records)
+                TRAIN_BATCH_SIZE, tf_records, shuffle_buffer_size=SHUFFLE_BUFFER_SIZE)
             output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
                                       train_mode=True, **self.hparams)
             train_tensors = train_ops(
@@ -134,25 +139,29 @@ class DualNetworkTrainer():
                     logger.add_summary(weight_summaries, global_step)
             self.save_weights()
 
-    def validate(self, tf_records, batch_size=128, logdir=None, init_from=None, num_steps=1000):
-        if logdir is None:
-            print("Need to run eval with a logdir")
+    def validate(self, tf_records, batch_size=128, init_from=None, num_steps=1000):
+        """Compute only the cost components for a set of tf_records, ideally a
+        holdout set, and report them to an 'eval' subdirectory of the logs.
+        """
+        cost_tensor_names = ['policy_cost', 'value_cost', 'l2_cost',
+                             'combined_cost']
+        if self.logdir is None:
+            print("Error, trainer not initialized with a logdir.")
             return
-        if not logdir.endswith('_eval'):
-            logdir += "_eval"
+
+        logdir = self.logdir + '-eval'
 
         with self.sess.graph.as_default():
             input_tensors = preprocessing.get_input_tensors(
-                TRAIN_BATCH_SIZE, tf_records)
+                batch_size, tf_records, shuffle_buffer_size=10000, filter_amount=0.5)
             output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
                                       train_mode=False, **self.hparams)
             train_tensors = train_ops(
                 input_tensors, output_tensors, **self.hparams)
 
             # just run our cost tensors
-            validate_tensors = {k: train_tensors[k] for k in (
-                'policy_cost', 'value_cost', 'l2_cost',
-                'combined_cost')}
+            validate_tensors = {k: train_tensors[k]
+                                for k in cost_tensor_names + ['global_step']}
             self.initialize_weights(init_from)
             training_stats = StatisticsCollector()
             logger = tf.summary.FileWriter(logdir, self.sess.graph)
@@ -162,11 +171,12 @@ class DualNetworkTrainer():
                     tensor_values = self.sess.run(validate_tensors)
                 except tf.errors.OutOfRangeError:
                     break
-                training_stats.report(tensor_values)
+                training_stats.report(
+                    {k: tensor_values[k] for k in cost_tensor_names})
 
-                if i % 10 == 9:
-                    accuracy_summaries = training_stats.collect()
-                    logger.add_summary(accuracy_summaries, i)
+            accuracy_summaries = training_stats.collect()
+            logger.add_summary(accuracy_summaries,
+                               tensor_values['global_step'])
 
 
 class DualNetwork():
--- a/dual_net.py
+++ b/dual_net.py
@@ -41,12 +41,6 @@ EXAMPLES_PER_GENERATION = 2000000
 # How many positions can fit on a graphics card. 256 for 9s, 16 or 32 for 19s.
 TRAIN_BATCH_SIZE = 256
 
-# The shuffle buffer size determines how far an example could end up from
-# where it started; this and the interleave parameters in preprocessing can give
-# us an approximation of a uniform sampling.  The default of 4M is used in
-# training, but smaller numbers can be used for aggregation or validation.
-SHUFFLE_BUFFER_SIZE = int(4*1e6)
-
 
 class DualNetworkTrainer():
     def __init__(self, save_file=None, logdir=None, **hparams):
@@ -104,7 +98,7 @@ class DualNetworkTrainer():
             num_steps = EXAMPLES_PER_GENERATION // TRAIN_BATCH_SIZE
         with self.sess.graph.as_default():
             input_tensors = preprocessing.get_input_tensors(
-                TRAIN_BATCH_SIZE, tf_records, shuffle_buffer_size=SHUFFLE_BUFFER_SIZE)
+                TRAIN_BATCH_SIZE, tf_records)
             output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
                                       train_mode=True, **self.hparams)
             train_tensors = train_ops(
--- a/dual_net.py
+++ b/dual_net.py
@@ -134,6 +134,40 @@ class DualNetworkTrainer():
                     logger.add_summary(weight_summaries, global_step)
             self.save_weights()
 
+    def validate(self, tf_records, batch_size=128, logdir=None, init_from=None, num_steps=1000):
+        if logdir is None:
+            print("Need to run eval with a logdir")
+            return
+        if not logdir.endswith('_eval'):
+            logdir += "_eval"
+
+        with self.sess.graph.as_default():
+            input_tensors = preprocessing.get_input_tensors(
+                TRAIN_BATCH_SIZE, tf_records)
+            output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
+                                      train_mode=False, **self.hparams)
+            train_tensors = train_ops(
+                input_tensors, output_tensors, **self.hparams)
+
+            # just run our cost tensors
+            validate_tensors = {k: train_tensors[k] for k in (
+                'policy_cost', 'value_cost', 'l2_cost',
+                'combined_cost')}
+            self.initialize_weights(init_from)
+            training_stats = StatisticsCollector()
+            logger = tf.summary.FileWriter(logdir, self.sess.graph)
+
+            for i in tqdm(range(num_steps)):
+                try:
+                    tensor_values = self.sess.run(validate_tensors)
+                except tf.errors.OutOfRangeError:
+                    break
+                training_stats.report(tensor_values)
+
+                if i % 10 == 9:
+                    accuracy_summaries = training_stats.collect()
+                    logger.add_summary(accuracy_summaries, i)
+
 
 class DualNetwork():
     def __init__(self, save_file, **hparams):
@@ -329,12 +363,14 @@ def logging_ops():
 
 def compute_update_ratio(weight_tensors, before_weights, after_weights):
     """Compute the ratio of gradient norm to weight norm."""
-    deltas = [after - before for after, before in zip(after_weights, before_weights)]
+    deltas = [after - before for after,
+              before in zip(after_weights, before_weights)]
     delta_norms = [np.linalg.norm(d.ravel()) for d in deltas]
     weight_norms = [np.linalg.norm(w.ravel()) for w in before_weights]
     ratios = [d / w for d, w in zip(delta_norms, weight_norms)]
     all_summaries = [
-        tf.Summary.Value(tag='update_ratios/' + tensor.name, simple_value=ratio)
+        tf.Summary.Value(tag='update_ratios/' +
+                         tensor.name, simple_value=ratio)
         for tensor, ratio in zip(weight_tensors, ratios)]
     return tf.Summary(value=all_summaries)
 
@@ -348,6 +384,7 @@ class StatisticsCollector(object):
     executions. Therefore, we aggregate the accuracy/cost ourselves at
     the python level, and then generate the summary protobufs for writing.
     """
+
     def __init__(self):
         self.accums = collections.defaultdict(list)
 
--- a/dual_net.py
+++ b/dual_net.py
@@ -143,7 +143,7 @@ class DualNetworkTrainer():
             self.save_weights()
 
     def validate(self, tf_records, batch_size=128, init_from=None, num_steps=1000):
-        """Compute only the cost components for a set of tf_records, ideally a
+        """Compute only the error terms for a set of tf_records, ideally a
         holdout set, and report them to an 'test' subdirectory of the logs.
         """
         cost_tensor_names = ['policy_cost', 'value_cost', 'l2_cost',
@@ -156,7 +156,7 @@ class DualNetworkTrainer():
 
         with self.sess.graph.as_default():
             input_tensors = preprocessing.get_input_tensors(
-                batch_size, tf_records, shuffle_buffer_size=10000, filter_amount=0.5)
+                batch_size, tf_records, shuffle_buffer_size=10000, filter_amount=0.05)
             output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
                                       train_mode=False, **self.hparams)
             train_tensors = train_ops(
--- a/dual_net.py
+++ b/dual_net.py
@@ -156,7 +156,7 @@ class DualNetworkTrainer():
 
         with self.sess.graph.as_default():
             input_tensors = preprocessing.get_input_tensors(
-                batch_size, tf_records, shuffle_buffer_size=10000, filter_amount=0.05)
+                batch_size, tf_records, shuffle_buffer_size=1000, filter_amount=0.05)
             output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
                                       train_mode=False, **self.hparams)
             train_tensors = train_ops(
--- a/dual_net.py
+++ b/dual_net.py
@@ -41,8 +41,10 @@ EXAMPLES_PER_GENERATION = 2000000
 # How many positions can fit on a graphics card. 256 for 9s, 16 or 32 for 19s.
 TRAIN_BATCH_SIZE = 256
 
-# The shuffle buffer size determines the max 'distance' an example ends up from
-# where it started.
+# The shuffle buffer size determines how far an example could end up from
+# where it started; this and the interleave parameters in preprocessing can give
+# us an approximation of a uniform sampling.  The default of 4M is used in
+# training, but smaller numbers can be used for aggregation or validation.
 SHUFFLE_BUFFER_SIZE = int(4*1e6)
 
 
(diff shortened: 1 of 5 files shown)


dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
                     logger.add_summary(weight_summaries, global_step)
             self.save_weights()
 
+    def validate(self, tf_records, batch_size=128, logdir=None, init_from=None, num_steps=1000):
+        if logdir is None:
+            print("Need to run eval with a logdir")
```
brilee: I think logdir should maybe just be passed into the DualNetTrainer constructor, and then you can have train/eval subdirectories of the logdir.

dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
...
+            return
+        if not logdir.endswith('_eval'):
```
brilee: tensorboard's logdir can have subdirectories like train and eval, in which case they'll appear on tensorboard as separate curves named train and eval.

dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
...
+            output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
+                                      train_mode=False, **self.hparams)
+            train_tensors = train_ops(
+                input_tensors, output_tensors, **self.hparams)
+
+            # just run our cost tensors
+            validate_tensors = {k: train_tensors[k] for k in (
+                'policy_cost', 'value_cost', 'l2_cost',
+                'combined_cost')}
+            self.initialize_weights(init_from)
```
brilee: there might be some weirdness here; if you redefine the same network (albeit with different inputs), then because we didn't put explicit names on the graph nodes, it'll continue autoincrementing the weight names.

So if you call train() and then eval(), then initialize_weights will reinitialize the weights in the train()-defined network, as opposed to the eval()-defined network. It's some pretty unfortunate global state in TF that you have to solve with different sessions.

dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
...
+
+            for i in tqdm(range(num_steps)):
+                try:
+                    tensor_values = self.sess.run(validate_tensors)
+                except tf.errors.OutOfRangeError:
+                    break
+                training_stats.report(tensor_values)
+
+                if i % 10 == 9:
+                    accuracy_summaries = training_stats.collect()
```
brilee: collect() should happen once at the very end, since you basically want to average all the eval() things into a single datapoint.

dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
...
+                    logger.add_summary(accuracy_summaries, i)
```
brilee: the second arg of add_summary is global_step; you want to fetch the global step by doing sess.run(train_tensors['global_step']). That way, the eval metrics show up correctly aligned to the train metrics.

dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
...
+            train_tensors = train_ops(
+                input_tensors, output_tensors, **self.hparams)
+
+            # just run our cost tensors
+            validate_tensors = {k: train_tensors[k] for k in (
+                'policy_cost', 'value_cost', 'l2_cost',
+                'combined_cost')}
+            self.initialize_weights(init_from)
+            training_stats = StatisticsCollector()
+            logger = tf.summary.FileWriter(logdir, self.sess.graph)
```
brilee: tbh I don't even know why a logger needs to know about the sess.graph. I suspect this is to generate an explorable graph widget in tensorboard.

This you might be able to put into the constructor, as well.

main.py
```diff
@@ -93,6 +93,15 @@ def train(chunk_dir, save_file, load_file=None, generation_num=0,
                 logdir=logdir, num_steps=num_steps, verbosity=verbosity)
 
 
+def validate(holdout_dir, load_file=None, logdir=None, num_steps=1000):
+    tf_records = sorted(gfile.Glob(os.path.join(holdout_dir, '*.tfrecord.zz')))
+    preprocessing.SHUFFLE_BUFFER_SIZE = int(1e5)
```
brilee: uh let's not do this :P

dual_net.py
```diff
@@ -134,6 +134,40 @@ def should_log(i):
                     logger.add_summary(weight_summaries, global_step)
             self.save_weights()
 
+    def validate(self, tf_records, batch_size=128, logdir=None, init_from=None, num_steps=1000):
```
brilee: unused argument batch_size

dual_net.py
amj: yeah; honestly i thought about this and figured that punting by having it be a separate subcommand rather than try to bolt it onto `main.py train` was simplest.  But it'd be nice to have this in our end-to-end local loop test.

dual_net.py
amj: yeah so my thought on this:  a run with ~hundreds of checkpoints can produce so much TB data that it takes ~30-45 minutes for tensorboard to chew through it all.  My thought was that by publishing validation stats for each new model with the same x-domain, they would overlay pretty easily and it would be easier to see overfitting.

I think having everything in an -eval directory and only having one point per model will work just as well, though, plus being able to align it with the training metrics is a big win.  Will do.

main.py
amj: i said this code wasn't actually ready for review, didn't i? :) :P

The other thing we can do, btw, is we can change the filter amount from 5% to ...something else. 50%?

dual_net.py
brilee: Yeah, having it as a separate command should avoid the above issues.

dual_net.py
brilee: interesting - my previous long runs have taken 1-2 minutes to load onto TB, but I guess these are even longer runs.

I would reduce the collect() frequency so that instead of having a gazillion noisy summary entries, we just have fewer summary entries averaged over longer time periods.

I don't think the overlaying is going to work the way you described - I think as is, everything will be overlaid where the X axis is 1-1000, and the 1000 y-values will literally be random noise centered around some value. Should collect() once at end and grab the global step and add_summary() once. That way the validation will plot nicely over time / over generations.

main.py
brilee: Do you suspect that I/O is now the bottleneck for training?

dual_net.py
amj: So are you thinking we pass the root logdir to the constructor and let train/eval append the train/eval string?

dual_net.py
amj: yep yep, i had been shooting for something else but it was a bad idea.

dual_net.py

dual_net.py
amj: agreed, bad idea, done.

dual_net.py
amj: If i put it into the constructor, i can't change it under train() and validate(); no problem though.

main.py
amj: On 9x9, probably.  I changed it because, when running validate standalone on past models, it'd take ~5 minutes to fill the shuffle buffer before doing anything.  And on validate, there's no need to shuffle at all.  I think it suggests that really the shuffle buffer size should be an argument on preprocessing.get_input_tensors

dual_net.py

Kashomon: comment? =)

main.py
```diff
@@ -93,6 +93,15 @@ def train(chunk_dir, save_file, load_file=None, generation_num=0,
                 logdir=logdir, num_steps=num_steps, verbosity=verbosity)
 
 
+def validate(holdout_dir, load_file=None, logdir=None, num_steps=1000):
```
Kashomon: also +comment =)

amj:
Ok this is ready for review, no matter what k8s-ci-robot says ;)  PTAL?

dual_net.py
```diff
@@ -41,14 +41,17 @@
 # How many positions can fit on a graphics card. 256 for 9s, 16 or 32 for 19s.
 TRAIN_BATCH_SIZE = 256
 
+# The shuffle buffer size determines the max 'distance' an example ends up from
```
Kashomon: I found this comment a bit obtuse. What are we shuffling? What do you mean by distance?

dual_net.py
```diff
@@ -134,6 +139,45 @@ def should_log(i):
                     logger.add_summary(weight_summaries, global_step)
             self.save_weights()
 
+    def validate(self, tf_records, batch_size=128, init_from=None, num_steps=1000):
+        """Compute only the cost components for a set of tf_records, ideally a
```
Kashomon: cost components? tf_records? 

Trying to nail down jargon since our goal is to be a readable implementation.

dual_net.py
```diff
@@ -134,6 +139,45 @@ def should_log(i):
...
+        holdout set, and report them to an 'eval' subdirectory of the logs.
+        """
+        cost_tensor_names = ['policy_cost', 'value_cost', 'l2_cost',
+                             'combined_cost']
+        if self.logdir is None:
+            print("Error, trainer not initialized with a logdir.")
```
Kashomon: Seems like this should be on STDERR

dual_net.py
```diff
@@ -134,6 +139,45 @@ def should_log(i):
...
+            return
+
+        logdir = self.logdir + '-eval'
+
+        with self.sess.graph.as_default():
+            input_tensors = preprocessing.get_input_tensors(
+                batch_size, tf_records, shuffle_buffer_size=10000, filter_amount=0.5)
```
Kashomon: should this be using SHUFFLE_BUFFER_SIZE?

Kashomon: What is this filter amount constant?

dual_net.py
```diff
@@ -134,6 +139,45 @@ def should_log(i):
...
+            output_tensors = dual_net(input_tensors, TRAIN_BATCH_SIZE,
+                                      train_mode=False, **self.hparams)
+            train_tensors = train_ops(
+                input_tensors, output_tensors, **self.hparams)
+
+            # just run our cost tensors
+            validate_tensors = {k: train_tensors[k]
+                                for k in cost_tensor_names + ['global_step']}
```
Kashomon: what is the global_step thing?

main.py
```diff
@@ -87,10 +87,19 @@ def train(chunk_dir, save_file, load_file=None, generation_num=0,
...
 
-    n = dual_net.DualNetworkTrainer(save_file)
+    n = dual_net.DualNetworkTrainer(save_file, logdir=logdir)
     with timer("Training"):
         n.train(tf_records, init_from=load_file,
-                logdir=logdir, num_steps=num_steps, verbosity=verbosity)
+                num_steps=num_steps, verbosity=verbosity)
+
+
+def validate(holdout_dir, load_file=None, logdir=None, num_steps=1000):
```
Kashomon: still missing a comment here

main.py
```diff
@@ -87,10 +87,19 @@ def train(chunk_dir, save_file, load_file=None, generation_num=0,
...
+    tf_records = sorted(gfile.Glob(os.path.join(holdout_dir, '*.tfrecord.zz')))
+    preprocessing.SHUFFLE_BUFFER_SIZE = int(1e5)
```
Kashomon: this is strange. Why are you overriding a CONSTANT? Should be passed in if it's gonna be overridden.

preprocessing.py
```diff
@@ -106,6 +104,7 @@ def batch_parse_tf_example(batch_size, example_batch):
 
 def read_tf_records(batch_size, tf_records, num_repeats=None,
                     shuffle_records=True, shuffle_examples=True,
+                    shuffle_buffer_size=10000,
```
Kashomon: another instance of that constant

preprocessing.py
```diff
@@ -139,13 +139,14 @@ def read_tf_records(batch_size, tf_records, num_repeats=None,
...
     if shuffle_examples:
-        dataset = dataset.shuffle(buffer_size=SHUFFLE_BUFFER_SIZE)
+        dataset = dataset.shuffle(buffer_size=shuffle_buffer_size)
     dataset = dataset.batch(batch_size)
     return dataset
 
 
 def get_input_tensors(batch_size, tf_records, num_repeats=None,
                       shuffle_records=True, shuffle_examples=True,
+                      shuffle_buffer_size=10000,
```
Kashomon: and here again. I feel like I'm missing something.

rl_loop.py
```diff
@@ -128,10 +128,23 @@ def train(logdir=None):
         logging.exception("Train error")
 
 
+def validate(logdir=None, model_num=None):
```
Kashomon: +comment

rl_loop.py
```diff
@@ -128,10 +128,23 @@ def train(logdir=None):
...
+    if model_num is None:
+        model_num, model_name = get_latest_model()
+    else:
+        model_name = get_model(model_num)
+
+    holdout_set = get_model(model_num - 1)
```
Kashomon: model_num - 1? That's a little counter intuitive. Add a comment?

dual_net.py
brilee: It's also definitely not max; just some vague notion of gaussian distribution with stdev

dual_net.py
```diff
@@ -134,6 +139,45 @@ def should_log(i):
...
+        """Compute only the cost components for a set of tf_records, ideally a
+        holdout set, and report them to an 'eval' subdirectory of the logs.
+        """
+        cost_tensor_names = ['policy_cost', 'value_cost', 'l2_cost',
+                             'combined_cost']
+        if self.logdir is None:
+            print("Error, trainer not initialized with a logdir.")
+            return
+
+        logdir = self.logdir + '-eval'
```
brilee: why not a subdirectory? logs/train and logs/eval sound reasonable to me.

Also, I don't feel like I have a good sense of what the difference between "validation" and "evaluation" means. Let's figure out which term is the right one to use, and stick with it.

dual_net.py
brilee: I'm also afraid of this filter constant. What if our eval results are different for differing values of this constant?

dual_net.py
brilee: remove global_step here and just call `sess.run(train_tensors['global_step'])` once at the end?

dual_net.py
```diff
@@ -134,6 +139,45 @@ def should_log(i):
...
+            self.initialize_weights(init_from)
+            training_stats = StatisticsCollector()
+            logger = tf.summary.FileWriter(logdir, self.sess.graph)
+
+            for i in tqdm(range(num_steps)):
+                try:
+                    tensor_values = self.sess.run(validate_tensors)
+                except tf.errors.OutOfRangeError:
+                    break
+                training_stats.report(
```
brilee: if you don't have global_step in here, you can just say `training_stats.report(tensor_values)`

local_rl_loop.py
```diff
@@ -34,15 +34,17 @@ def rl_loop():
         'k': 8, 'fc_width': 16, 'num_shared_layers': 1, 'l2_strength': 1e-4, 'momentum': 0.9}
 
     dual_net.TRAIN_BATCH_SIZE = 16
+    dual_net.SHUFFLE_BUFFER_SIZE = 10000
 
     #monkeypatch the shuffle buffer size so we don't spin forever shuffling up positions.
```
brilee: comment appears out of order now?

rl_loop.py
brilee: I think validation should probably run on last 50 generations (or maybe even ALL the generations) because we want to verify that our ability to predict game outcomes isn't degrading. So we'd want to go back and look at really old games too, not just the most recent one.

dual_net.py
amj: the max it *could* end up, then ;P  But yes, i'm either a) explaining the entire semantics of tf.Dataset.shuffle here, or b) leaving a comment like "shuffle buffer size: the size of the shuffle buffer".

dual_net.py
amj: The cost components of our graph; tf_records are a set of tf_examples.  these are kind of standard for tensorflow, IIUC, so i'm not sure how much we need to explicitly describe them.  

Happy to elaborate a bit more here but this brings up a kind of a question about DRY-ness of the comments:  we have dual_net.DualNetworkTrainer.train, rl_loop.train, and main.train.  They're all gonna be real similar, and duplicating the docstrings seems pointless, esp. when one is just a shell around the others.

dual_net.py
amj: i'm not sure why they would be different: we aren't going to overfit to them ;P  Our holdout data is already small -- 5% of all games.  Taking 5% of 5% might be very few positions, and running for 1k steps will iterate repeatedly if they are exhausted, [as we call read_tf_records with num_repeats=None everywhere](https://github.com/tensorflow/minigo/blob/master/preprocessing.py#L137), which would lead to *noisier* eval results if we *don't* step up this filter amount.

Josh: i'm not sure if you want a comment on it here -- it's commented in preprocessing.py.

main.py
amj: obsolete, removed. Thanks!

rl_loop.py
amj: So we would be checking a new model against the holdout data of the previous N models?  that sounds fine.  That makes my comments about filter_amount above irrelevant, and i'll revert it to 5%

dual_net.py
amj: Validate = compute error on holdout set
Evaluation = model vs model playoff match used by AGZ.

dual_net.py
amj: this makes more sense, will do.

local_rl_loop.py
amj: Done.

jmgilmer:
Here are some plots I could imagine generating:

1. Train error by minibatch idx - already done, and shows weird behavior in value net due to overfitting early.
2. Current self play validation error by minibatch idx - same as 1. but evaluated on held out data for the current generation, this is what is currently being implemented?
3. Pro validation error by minibatch idx - same as 2. but evaluated on pro data instead. 
4. Current generation self play validation error by model id - Note the x-axis is different now. This would measure how good all previous models are at predicting the current selfplay data, as @brilee ilee mentioned this should ideally be a monotone curve. 
5. Pro self play validation error by model id - Same as 4. but on pro data. 

It seems like currently 2. is being implemented. Not sure the best way to handle all 5 plots (I think they are all valuable). Perhaps 4 and 5 could be a one-off script we call every once in a while and 1-3 is done by trainer? Question is where would the holdout pro data live?

local_rl_loop.py
jmgilmer: Are you thinking 1 plot per generation? Or 1 plot that measures all prior generations in predicting this generations self play data?

rl_loop.py
jmgilmer: Are you thinking 1 plot per generation (that's a lot of plots)? Or 1 plot that has generation id on the x-axis and measures the error on this generation's self play data/pro data?

dual_net.py
amj: comment updated.

dual_net.py

dual_net.py
amj: no; i can make the constant TRAIN_SHUFFLE_BUFFER_SIZE but that seems a little verbose.

amj: There's no consequences to not shuffling the validation data at all, because we're only computing the overall average error.

preprocessing.py
amj: I don't want this to depend on dual_net, but i don't want to not have a default value.

rl_loop.py
amj: @brilee this would make it share a lot of the functionality i was noodling with on my [train from n branch](https://github.com/amj/minigo/blob/train-from-n/main.py#L89), as far as filtering the training chunks by their model name.  I could do something like that here as well.

amj:
@jmgilmer 
1. Right
2. Correct, this is what we're currently implementing.  @brilee suggests making this on the holdout sets of the most recent 50 generations instead of on the holdout set from the current generation.  Thoughts?
3. This would be great to have, and could/should probably share most the same code; would just need to be pointed at a directory of preprocessed pro games.
4 & 5. This seems neat -- how expensive would it be to compute it?  It seems like it would get progressively more expensive to compute.

jmgilmer:
5. has been implemented in oneoffs/training_curve.py. It could be easily modified to run on selfplay data as well to do 4. It takes about an hour to generate the curve on 150 19x19 models. One could also just evaluate every k'th model if speed is an issue.

dual_net.py
Kashomon: Dual net should probably be the SOT for comments then.

amj:
Ok @brilee @Kashomon i think this is good to go as far as the functionality in dual_net and main.py; we don't yet have a good story for how this fits into the full-auto solution, here's why:

 - If we want to validate on the most-recent-20/50/N models, this would be similar to our train() semantics
 - But our training data is preaggregated and lives all in one flat directory
 - This means it's easy to only look at e.g. the most recent 400 files for 4M positions.
 - Our holdout data is all per-generation:  data/holdout/<model name>/foo.tfrecord.zz ...
 - this means we either have to 
   - extract the model name from the load_file parameter (not great)
   - specify the model num as a param (eh)
   - aggregate the holdout data (seems yuck)

What do you think?

amj:
I went with "specify the model num as a param, default to most recent" and used the most recent 20ish generations of holdout data.  This pushed around a couple calls.

PTAL, thanks :)

dual_net.py
```diff
@@ -329,12 +376,14 @@ def logging_ops():
...
     """Compute the ratio of gradient norm to weight norm."""
-    deltas = [after - before for after, before in zip(after_weights, before_weights)]
+    deltas = [after - before for after,
+              before in zip(after_weights, before_weights)]
     delta_norms = [np.linalg.norm(d.ravel()) for d in deltas]
     weight_norms = [np.linalg.norm(w.ravel()) for w in before_weights]
     ratios = [d / w for d, w in zip(delta_norms, weight_norms)]
     all_summaries = [
-        tf.Summary.Value(tag='update_ratios/' + tensor.name, simple_value=ratio)
+        tf.Summary.Value(tag='update_ratios/' +
```
brilee: splitting this line across the + is weird - which code autoformat are you using?

main.py
```diff
@@ -87,10 +87,23 @@ def train(chunk_dir, save_file, load_file=None, generation_num=0,
...
 
-    n = dual_net.DualNetworkTrainer(save_file)
+    n = dual_net.DualNetworkTrainer(save_file, logdir=logdir)
     with timer("Training"):
         n.train(tf_records, init_from=load_file,
-                logdir=logdir, num_steps=num_steps, verbosity=verbosity)
+                num_steps=num_steps, verbosity=verbosity)
+
+
+def validate(tf_records, load_file=None, logdir=None, num_steps=1000):
```
brilee: I'm not sure I agree with passing in tf_records via command line args. It will almost certainly run into restrictions for bash wildcard expansion or process limits on sys.argv or something weird like that.

dual_net.py
```diff
@@ -41,14 +41,19 @@
 # How many positions can fit on a graphics card. 256 for 9s, 16 or 32 for 19s.
 TRAIN_BATCH_SIZE = 256
 
+# The shuffle buffer size determines how far an example could end up from
+# where it started; this and the interleave parameters in preprocessing can give
+# us an approximation of a uniform sampling.  The default of 4M is used in
+# training, but smaller numbers can be used for aggregation or validation.
+SHUFFLE_BUFFER_SIZE = int(4*1e6)
```
brilee: definitely not a fan of this, but am okay with it pending a better solution for hyperparams.

main.py
brilee: Also I think for this to work properly via command line, arghparse expects the signature to be `*tf_records`

Anyway, what about passing in *tf_record_dirs, and then globbing inside main.validate() to recursively find all contained tf_record files?

amj: oh good point.  What would you say to making validate *not* a CLI command, only an rl_loop command?

dual_net.py
amj: Ok, i put it back in preprocessing and had the func bodies check for =None, which lets us monkeypatch in a new default (damn python's sticky defaults)

dual_net.py
amj: autopep8

main.py
amj: main.validate now takes a list of directories ;)  Done & merging!

Issue #7, opened 2018-01-29
Clean up naming in coords.py


brilee:
Our naming scheme for coord type to coord type is a mess. Some homogenization would be nice.

PR #81, opened 2018-02-10
Improve coords.py

nealwu:
Fixes https://github.com/tensorflow/minigo/issues/7

Also fixed a bucket name discrepancy in README.md.

Passes unit tests, and local_rl_loop.py seems to train normally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ pip-selfcheck.json
 sgf
 pyvenv.cfg
 .DS_store
+minigo/
 # Vim temp files
 *.swp
 *.swo
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ pip-selfcheck.json
 sgf
 pyvenv.cfg
 .DS_store
+logs/
 
 # virtualenv folder
 minigo/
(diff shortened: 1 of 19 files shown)


.gitignore
```diff
@@ -8,6 +8,11 @@ pip-selfcheck.json
 sgf
 pyvenv.cfg
 .DS_store
+logs/
+
+# virtualenv folder
+minigo/
```
brilee: Interesting, my virtualenv got installed in ./bin/ and other misc. files. And I think Andrew's gets installed in ~/.virtualenvs/minigo. I'm starting to wonder how many virtualenv/venv/pipenv variants there are out there...

nealwu: Haha I see. I followed [this link](http://docs.python-guide.org/en/latest/dev/virtualenvs/#lower-level-virtualenv) from the README which just tells you to run `virtualenv my_project`, so naturally I named it `minigo`.

brilee:
Ah, I see. In this case, I think it's appropriate to put it in a personal
.gitignore rather than the repo-specific gitignore? Since other people may
not have implemented their virtualenvs the same way.

nealwu:
Sure that's fine; took it out.

PR #82, opened 2018-02-12
Fix some lint errors.

Kashomon:


local_rl_loop.py
```diff
@@ -29,6 +29,7 @@
 
 
 def rl_loop():
+    """Run the reinforcement learning loop"""
```
amj: Maybe leave a comment that this is more of an e2e integration test than a serious way to  run training?

Kashomon: done

PR #86, opened 2018-02-13
Re-add Set gpu_options.allow_growth = True

killerducky:
Reduces memory used for 199-lightning from 2552MiB to 523MiB.

This was added in Pull #33, but got lost.

There are a couple other tf.Session objects creates for other use cases, maybe this should be added to those also?

### DIFF
--- a/dual_net.py
+++ b/dual_net.py
@@ -141,7 +141,9 @@ class DualNetwork():
         self.hparams = get_default_hyperparams(**hparams)
         self.inference_input = None
         self.inference_output = None
-        self.sess = tf.Session(graph=tf.Graph())
+        config = tf.ConfigProto()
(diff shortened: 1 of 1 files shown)


Kashomon:
Refresh my memory: Why do we want to reduce GPU usage? I was reading the docs on this: https://www.tensorflow.org/programmers_guide/using_gpu. It seems like it's more efficient to just allocate all the memory. For self-play/training we're running essentially 1 pod/gpu.  However, for users this might be useful. 

So, perhaps this should be configurable rather than hardcoded?

killerducky:
I'm doing test matches of base minigo vs modified minigo own my home machine, so I need two instances to fit in my one GPU. Sure it could be configurable.

PR #87, opened 2018-02-13
Use dynamic parent Q for children's prior

killerducky:
As the search collects more data, update unvisited children with the new more accurate prior.

This is an alternative to #41. LZ is testing this now https://github.com/gcp/leela-zero/pull/866, so far it is slightly worse than the current FPU reduction code in LZ (140-162 = 46%). But IMO it's still better since it seems more logical.

Here are my test results for minigo.

```
mgbase-000199-lightning-r-800       5  7.69%       3   9.09%     2   6.25%    239.29
mgdynpq-000199-lightning-r-800     60 92.31%       30 93.75%     30 90.91%    250.58
                                                   33 50.77%     32 49.23%
```

### DIFF
--- a/mcts.py
+++ b/mcts.py
@@ -122,6 +122,16 @@ class MCTSNode(object):
         "Return value of position, from perspective of player to play."
         return self.Q * self.position.to_play
 
+    # As the search collects more data, update unvisited children with the
+    # new more accurate prior.
+    # TODO: Maybe we could update even the nodes we have already visited?
+    # We could remove the prior from child_W and always add it in one of
+    # the getter functions
+    def update_child_W_prior(self):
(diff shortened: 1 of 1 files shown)


amj:
This looks awesome, thanks for doing it!  Do you have an idea if this change affects playout speed?  I don't think we've got any instructions for how to profile it, but is that something you'd be able to do?  MCTS is the most executed code in any given game, by far, and it can be pretty easy to fall off the numpy-optimized paths.

killerducky:
This code needs to account for virtual losses. I'll work on it later.

amj:
We've learned some interesting things here ;)  Closing as obsolete.

PR #92, opened 2018-02-15
Adds validate call to rl_runner. Fixes #43

### LINKED ISSUE #43: Create validation set (self-play and professional)
For some small percent of self-play games, we should set them aside, never train on them, and check the value net loss on these games. This should help us figure out if we're overfitting.

A variant on this is to create a validation set of pro games. Self-play validation sets are nice because there's not a very large library of pro 9x9 games. But for 19x19, a pro validation set is doable.

amj:
Also:
- Changes default # of validation steps 
- Ensures the summary log is written afterwards

Fixes #43

### DIFF
--- a/rl_runner.py
+++ b/rl_runner.py
@@ -44,6 +44,10 @@ def loop(logdir=None):
             subprocess.call(
                 ("python rl_loop.py train --logdir=%s" % logdir).split())
 
+        with timer("validate"):
+            subprocess.call(
+                ("python rl_loop.py validate --logdir=%s" % logdir).split())
+
 
 if __name__ == '__main__':
     argh.dispatch_command(loop)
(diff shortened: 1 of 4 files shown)


rl_runner.py
```diff
@@ -44,6 +44,10 @@ def loop(logdir=None):
             subprocess.call(
                 ("python rl_loop.py train --logdir=%s" % logdir).split())
 
+        with timer("validate"):
```
Kashomon: Hm. Do we want the trainer to do both validation and training?

amj: yeah, why not?  validate is ~a minute, or about ~1% of overhead

Kashomon: SGTM.

amj: We might re-visit this when we containerize the trainer; Brian mentioned that he'd like more validation steps, which would move us from 1% overhead up to 2% or 3% which seems like enough to break validation off of the main trainer process.

PR #109, opened 2018-02-22
Remove anti-affinity constraint for scheduling gpu-player

amj:
not needed and seems to greatly slow down scheduling.

### DIFF
--- a/cluster/gpu-player.yaml
+++ b/cluster/gpu-player.yaml
@@ -3,28 +3,20 @@ kind: Job
 metadata:
   name: minigo-gpu-player
 spec:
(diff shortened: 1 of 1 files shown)


cluster/gpu-player.yaml
```diff
@@ -3,29 +3,21 @@ kind: Job
 metadata:
   name: minigo-gpu-player
 spec:
-  parallelism: $NUM_K8S_NODES
+  parallelism: 10
```
Kashomon: Should this be configurable?

amj: nah.  it's not worth the math hassle, plus it's easier to fly it manually.

PR #111, opened 2018-02-24
Fix evaluate

amj:
Rewrites the evaluate command to actually work w/ the new trainer rewrite and the new vlosses rewrite.

At some point, the 'name' field was removed from the DualNetwork object, although not actually removed from places that either set or referenced it.  Happily, those codepaths did not diverge...
Perhaps the player object should keep track of its semantic 'name', distinct from the filename/path used to reference its network files, and also distinct from the identifier it uses for GTP.

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -20,6 +20,8 @@ import sgf_wrapper
 
 from gtp_wrapper import MCTSPlayer
 
+SIMULTANEOUS_LEAVES = 8
+
 
 def play_match(black_net, white_net, games, readouts, sgf_dir, verbosity):
     """Plays matches between two neural nets.
@@ -34,9 +36,9 @@ def play_match(black_net, white_net, games, readouts, sgf_dir, verbosity):
 
     # For n games, we create lists of n black and n white players
     black = MCTSPlayer(
-        black_net, verbosity=verbosity, two_player_mode=True)
+        black_net, verbosity=verbosity, two_player_mode=True, num_parallel=SIMULTANEOUS_LEAVES)
     white = MCTSPlayer(
-        white_net, verbosity=verbosity, two_player_mode=True)
+        white_net, verbosity=verbosity, two_player_mode=True, num_parallel=SIMULTANEOUS_LEAVES)
 
     black_name = os.path.basename(black_net.save_file)
     white_name = os.path.basename(white_net.save_file)
@@ -78,13 +80,13 @@ def play_match(black_net, white_net, games, readouts, sgf_dir, verbosity):
             move = active.pick_move()
             active.play_move(move)
             inactive.play_move(move)
-            print(active.root.position)
 
             dur = time.time() - start
             num_move += 1
 
             if (verbosity > 1) or (verbosity == 1 and num_move % 10 == 9):
                 timeper = (dur / readouts) * 100.0
+                print(active.root.position)
                 print("%d: %d readouts, %.3f s/100. (%.2f sec)" % (num_move,
                                                                    readouts,
                                                                    timeper,
--- a/evaluation.py
+++ b/evaluation.py
@@ -14,84 +14,78 @@
 
 """Evalation plays games between two neural nets."""
 
+import os
 import time
+import sgf_wrapper
 
 from gtp_wrapper import MCTSPlayer
 
 
-def play_match(black_net, white_net, games, readouts, verbosity):
+def play_match(black_net, white_net, games, readouts, sgf_dir, verbosity):
     """Plays matches between two neural nets.
 
     black_net: Instance of minigo.DualNetwork, a wrapper around a tensorflow
         convolutional network.
     white_net: Instance of the minigo.DualNetwork.
     games: number of games to play. We play all the games at the same time.
+    sgf_dir: directory to write the sgf results.
     readouts: number of readouts to perform for each step in each game.
     """
 
     # For n games, we create lists of n black and n white players
-    black_players = [MCTSPlayer(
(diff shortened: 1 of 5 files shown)


local_rl_loop.py
```diff
@@ -79,6 +80,10 @@ def rl_loop():
             holdout_pct=100,
             readouts=10)
 
+        print("See sgf files here?")
+        sgf_listing = subprocess.check_output("ls -l {}/full".format(sgf_dir).split())
```
Kashomon: Just use https://docs.python.org/3/library/os.html#os.listdir and filter?

local_rl_loop.py
```diff
@@ -79,6 +80,10 @@ def rl_loop():
             holdout_pct=100,
             readouts=10)
 
+        print("See sgf files here?")
```
Kashomon: Not sure I understand this comment.

amj: these are 'integration tests' with big scare quotes.  If you're running this and you want to see that everything still works, you should see some files with nonzero size.

local_rl_loop.py
amj: see below; i'd like to see these are nonzero size files.

evaluation.py
```diff
@@ -14,84 +14,80 @@
...
+                                                  active.result_string, black_name=black_name,
+                                                  white_name=white_name)
+                    _file.write(sgfstr)
+                print("Finished game", i, active.result_string)
+                break
+
             move = active.pick_move()
             active.play_move(move)
             inactive.play_move(move)
+            print(active.root.position)
```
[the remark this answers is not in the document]
amj: Good point, i should check the verbosity level.

evaluation.py
```diff
@@ -14,84 +14,82 @@
...
-        # print some stats on the search
-        if verbosity >= 3:
-            print(player_pairs[0][0].root.position)
-
-        for black, white in player_pairs:
-            active = white if num_moves % 2 else black
-            inactive = black if num_moves % 2 else white
+        while True:
+            start = time.time()
+            for _ in range(readouts):
```
brilee: this should read `while player.num_readouts < readouts`, because parallel vloss search will add up to 8 readouts per tree_search call.

evaluation.py
```diff
@@ -14,84 +14,82 @@
...
+                player = white if num_move % 2 else black
```
brilee: this should happen outside of for loop,

evaluation.py
```diff
@@ -14,84 +14,82 @@
...
+                player.tree_search()
+
+            # print some stats on the search
+            if verbosity >= 3:
+                print(player.root.position)
+
+            active = white if num_move % 2 else black
```
brilee: this computation should happen once at top

evaluation.py
```diff
@@ -14,84 +14,82 @@
...
+                active.set_result(-1 *
+                                  active.root.position.to_play, was_resign=True)
+                inactive.set_result(
+                    active.root.position.to_play, was_resign=True)
+
+            if active.is_done():
+                fname = "{:d}-{:s}-vs-{:s}-{:d}.sgf".format(int(time.time()),
+                                                            white_name, black_name, i)
+                with open(os.path.join(sgf_dir, fname), 'w') as _file:
+                    sgfstr = sgf_wrapper.make_sgf(active.position.recent,
```
brilee: I could have sworn I had some cute code snippet that would splice in alternate comments from both players so you'd get to see both players' debug perspectives. Let me figure out where I put it...

evaluation.py
```diff
@@ -14,84 +14,82 @@
...
+                print(player.root.position)
+
+            active = white if num_move % 2 else black
+            inactive = black if num_move % 2 else white
             # First, check the roots for hopeless games.
             if active.should_resign():  # Force resign
-                continue
+                active.set_result(-1 *
+                                  active.root.position.to_play, was_resign=True)
+                inactive.set_result(
```
brilee: do you need to set this result if you don't use inactive's result string?

evaluation.py
```diff
@@ -14,84 +14,82 @@
...
+                    active.root.position.to_play, was_resign=True)
+
+            if active.is_done():
```
brilee: seems like you could just remove this check and fuse the contents of this IF with the contents of the IF above. Then you wouldn't need is_done() either.

local_rl_loop.py
brilee: https://docs.python.org/3/library/os.html#os.scandir  is the equivalent of `ls -l` in pure python

brilee: also, `subprocess.X(some_string.split())` is an antipattern. If `some_string` contains an unintended space, it would be the equivalent of passing in a filename to a bash command where the filename had a space in it. In bash, you would quote the filename; the way python would know to quote the filename is by passing in a list as you intended it to be split: `subprocess.check_output(['ls', '-l', filename])`

evaluation.py
amj: Oops, right.

evaluation.py
amj: Added.

evaluation.py
amj: Done

amj:
'scandir' is not quite what i want.  i just want it to spit out 'ls -l' so i can verify there's some kind of nonzero-sized files.  I don't want to check what they are, or read them, or make any other assertions other than 'looks like small SGFs were written'.  

Re: subprocess.x(foo.split()) right right right, fixed.

PR #112, opened 2018-02-24
Refactor cluster directory.

Kashomon:
Refactor the cluster directory somewhat to set-up for a more complex
topology (evaluation server, etc.). In particular, this commit:

- Moves self-play specific logic to a selfplay directory.
- Adds an unset_common.sh helper
- Fixes some bucket-permissions logic.
- Add a bootstrap-model helper
- Make the scripts not depend on the CWD of the caller.


cluster/selfplay/deploy-cpu-player.sh
```diff
@@ -12,9 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#!/bin/bash
-
-source ./common.sh
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source ${SCRIPT_DIR}/../common.sh
```
amj: i'm a little concerned by the "SCRIPT_DIR" shenanigans when they're immediately followed by referencing a parent directory.  Seems like having a pwd-independent script that immediately asserts a dependency on its directory hierarchy is working at cross purposes.

PR #113, opened 2018-02-25
Estimator rewrite

brilee:
Estimator rewrite!

Rough outline: training occurs locally, with checkpointing happening automatically. After a fixed number of steps, as calculated by generation number, we take the latest exported checkpoint on disk and copy it to GCS, where the selfplay workers pick it up.

All logging is confirmed working, including a sessionrunhook to calculate update ratios. Apparently Estimator throws in a "steps per second" logger for free, too. Validation now has a nice framework under which to classify selfplay vs. pro dataset validation.

There's some continued hyperparameter ugliness; my next task will be to clean all this up.

### DIFF
--- a/dual_net.py
+++ b/dual_net.py
@@ -358,6 +358,9 @@ def model_fn(features, labels, mode, params, config):
 
 
     metric_ops = {
+        'accuracy': tf.metrics.accuracy(labels=labels['pi_tensor'],
+                                        predictions=policy_output,
+                                        name='accuracy_op'),
         'policy_cost': tf.metrics.mean(policy_cost),
         'value_cost': tf.metrics.mean(value_cost),
         'l2_cost': tf.metrics.mean(l2_cost),
--- a/dual_net.py
+++ b/dual_net.py
@@ -26,6 +26,7 @@ import os.path
 import itertools
 import sys
 import tensorflow as tf
+from tensorflow.python.training.summary_io import SummaryWriterCache
 from tqdm import tqdm
 from typing import Dict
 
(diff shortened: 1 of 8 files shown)


brilee:
@nkashy1 @jmgilmer

dual_net.py
```diff
@@ -271,47 +133,57 @@ def get_default_hyperparams(**overrides):
...
-        pos_tensor: [BATCH_SIZE, go.N, go.N, features.NEW_FEATURES_PLANES]
-        pi_tensor: [BATCH_SIZE, go.N * go.N + 1]
-        value_tensor: [BATCH_SIZE]
+    Args:
+        features: tensor with shape
+            [BATCH_SIZE, go.N, go.N, features.NEW_FEATURES_PLANES]
+        labels: dict from string to tensor with shape
+            'pi_tensor': [BATCH_SIZE, go.N * go.N + 1]
+            'value_tensor': [BATCH_SIZE]
+        mode: a tf.estimator.ModeKeys
```
amj: which modes are supported?

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+        'accuracy': tf.metrics.accuracy(labels=labels['pi_tensor'],
+                                        predictions=policy_output,
+                                        name='accuracy_op'),
+        'policy_cost': tf.metrics.mean(policy_cost),
+        'value_cost': tf.metrics.mean(value_cost),
+        'l2_cost': tf.metrics.mean(l2_cost),
+        'policy_entropy': tf.metrics.mean(policy_entropy),
+        'combined_cost': tf.metrics.mean(combined_cost),
+    }
+    # Register metrics as tf.summary to make them show up during training
```
amj: "show up" ?

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+    for metric_name, metric_op in metric_ops.items():
+        tf.summary.scalar(metric_name, metric_op[1])
```
amj: [1] is the op, i guess?  Seems like tf should use a namedtuple or something.

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+            'policy_output': policy_output,
+            'value_output': value_output,
+        },
+        loss=combined_cost,
+        train_op=train_op,
+        eval_metric_ops=metric_ops,
+        )
+
+
+def get_estimator(working_dir, **hparams):
```
amj: ah, the glorious payoff :)

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+        hparams: hyperparams of the model.
+    """
+    hparams = get_default_hyperparams(**hparams)
+    # a bit hacky - forge an initial checkpoint with the name that subsequent
+    # Estimator runs will expect to find.
+    #
+    # Estimator will do this automatically when you call train(), but calling
+    # train() requires data, and I didn't feel like creating training data in
+    # order to run the full train pipeline for 1 step.
+    estimator_initial_checkpoint_name = 'model.ckpt-1'
```
amj: what happens if this file exists?

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+        model_fn(features, labels, tf.estimator.ModeKeys.PREDICT, hparams, None)
+        sess.run(tf.global_variables_initializer())
+        tf.train.Saver().save(sess, save_file)
+
+
+def export_model(working_dir, model_path):
+    """Take the latest checkpoint and export it to model_path for selfplay.
+
+    Assumes that all relevant model files are prefixed by the same name.
+    (For example, foo.index, foo.meta and foo.data-00000-of-00001).
```
amj: maybe later we can export & optimize these graphs for inference as a single file.

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+        suffix = filename.partition(latest_checkpoint)[2]
+        destination_path = model_path + suffix
+        print("Copying {} to {}".format(filename, destination_path))
+        tf.gfile.Copy(filename, destination_path)
+
+
+def train(working_dir, tf_records, generation_num, **hparams):
+    assert generation_num > 0, "Model 0 is random weights"
+    estimator = get_estimator(working_dir, **hparams)
+    max_steps = generation_num * EXAMPLES_PER_GENERATION // TRAIN_BATCH_SIZE
```
amj: this makes TRAIN_BATCH_SIZE immutable for the length of a run, correct?  I.e., if i lowered it on the fly, the next training run would be WAY more steps, correct?  Can this be based off the last checkpoint global_step instead?

dual_net.py
```diff
@@ -321,58 +193,133 @@ def my_res_layer(inputs, train_mode):
...
+
+
+def validate(working_dir, tf_records, checkpoint_name=None, **hparams):
+    estimator = get_estimator(working_dir, **hparams)
+    if checkpoint_name is None:
+        checkpoint_name = estimator.latest_checkpoint()
+    input_fn = lambda: preprocessing.get_input_tensors(
+        TRAIN_BATCH_SIZE, tf_records, shuffle_buffer_size=1000,
+        filter_amount=0.05)
+    estimator.evaluate(input_fn, steps=1000)
```
amj: so nice!

dual_net.py
```diff
@@ -190,11 +55,11 @@ def __init__(self, save_file, **hparams):
...
     def initialize_graph(self):
         with self.sess.graph.as_default():
-            input_tensors = get_inference_input()
-            output_tensors = dual_net(input_tensors, batch_size=-1,
-                                      train_mode=False, **self.hparams)
-            self.inference_input = input_tensors
-            self.inference_output = output_tensors
+            features, labels = get_inference_input()
+            estimator_spec = model_fn(features, labels,
+                tf.estimator.ModeKeys.PREDICT, self.hparams, None)
```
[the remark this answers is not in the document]
brilee: done

dual_net.py
```diff
@@ -190,11 +55,11 @@ def __init__(self, save_file, **hparams):
 
     def initialize_graph(self):
         with self.sess.graph.as_default():
-            input_tensors = get_inference_input()
-            output_tensors = dual_net(input_tensors, batch_size=-1,
-                                      train_mode=False, **self.hparams)
-            self.inference_input = input_tensors
-            self.inference_output = output_tensors
+            features, labels = get_inference_input()
```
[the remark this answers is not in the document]
brilee: docstring added

dual_net.py
brilee: comment added

dual_net.py
brilee: Digging into it, `metric_op` is a tuple of
mean: A Tensor representing the current mean, the value of total divided by count.
update_op: An operation that increments the total and count variables appropriately and whose value matches mean_value.

It seems like the only difference is whether all N total/count are used, or N-1 total/count are used. Agree it's still a bit confusing.

dual_net.py
brilee: comment added

dual_net.py
brilee: Hm. Would probably get overwritten. Probably not a big deal, because old checkpoints are not really used other than for backup purposes.

dual_net.py
brilee: Hm. Eventually, yes. The default estimator exporter creates a function signature that isn't based on feed_dict, but rather, hardwired to accept serialized tf.Example objects. Which is to say, I think the Estimator API expects deserialization to happen in the model_fn rather than the input_fn. But I put the deserialization code in the input_fn so that we could use feed_dict as an alternate input pathway into model_fn. Some wrangling will be necessary later on, I think, if we want to make C++ inference work.

dual_net.py
brilee: Yes, this would make the batch size and examples_per_generation immutable per run.

It'd make the code a bit uglier to pull the global_step in from the previous checkpoint, but I agree it'd make for a more robust calculation.

main.py
```diff
@@ -76,38 +76,41 @@ def gtp(load_file: "The path to the network model files"=None,
...
-def validate(*tf_record_dirs, load_file=None, logdir=None, num_steps=100):
-    """Computes the error terms for a set of holdout data specified by
-    `holdout_dir`, using the model specified at `load_file` and logging TB
-    metrics to the dir in `logdir`, using `num_steps` batches of examples
-    """
-    n = dual_net.DualNetworkTrainer(logdir=logdir)
-
+def validate(
+    working_dir: 'tf.estimator working directory',
+    *tf_record_dirs: 'Directory where holdout data are',
```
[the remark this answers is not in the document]
brilee: done

main.py
```diff
@@ -76,38 +76,41 @@ def gtp(load_file: "The path to the network model files"=None,
...
+    checkpoint_name: 'Which checkpoint to evaluate (None=latest)'=None,
+    validate_name: 'Name for validation (i.e. selfplay or professional)'=None):
```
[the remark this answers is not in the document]
brilee: changed to 'human' :P

preprocessing.py
```diff
@@ -85,7 +85,7 @@ def batch_parse_tf_example(batch_size, example_batch):
     Args:
         example_batch: a batch of tf.Example
     Returns:
-        A dict of batched tensors
+        A tuple (feature_tensor, dict of label_tensors)
```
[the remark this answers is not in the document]
brilee: done

rl_loop.py
```diff
@@ -100,7 +103,7 @@ def bootstrap():
     bootstrap_name = shipname.generate(0)
     bootstrap_model_path = os.path.join(MODELS_DIR, bootstrap_name)
     print("Bootstrapping model at {}".format(bootstrap_model_path))
-    main.bootstrap(bootstrap_model_path)
+    main.bootstrap(ESTIMATOR_WORKING_DIR, bootstrap_model_path)
```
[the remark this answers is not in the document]
brilee: done

main.py
```diff
@@ -76,38 +76,41 @@ def gtp(load_file: "The path to the network model files"=None,
             sys.stdout.flush()
 
 
-def bootstrap(save_file):
-    dual_net.DualNetworkTrainer(save_file).bootstrap()
+def bootstrap(working_dir, model_save_file):
+    _ensure_dir_exists(os.path.dirname(model_save_file))
```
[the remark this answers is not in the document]
brilee: comments added; model_save_file renamed to model_save_path

main.py
```diff
@@ -76,38 +76,41 @@ def gtp(load_file: "The path to the network model files"=None,
...
+    working_dir: 'tf.estimator working directory',
+    *tf_record_dirs: 'Directory where holdout data are',
+    checkpoint_name: 'Which checkpoint to evaluate (None=latest)'=None,
+    validate_name: 'Name for validation (i.e. selfplay or professional)'=None):
+    tf_records = []
     with timer("Building lists of holdout files"):
-        tf_records = [item for sublist in map(lambda path: gfile.Glob(
-            os.path.join(path, '*.zz')), tf_record_dirs) for item in sublist]
+        for record_dir in tf_record_dirs:
+            tf_records.extend(gfile.Glob(os.path.join(record_dir, '*.zz')))
```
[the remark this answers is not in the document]
amj: Acknowledged.  Original is a derivative of [this](https://stackoverflow.com/a/952952), although i admit it's a little obfuscated :(.

main.py
```diff
@@ -76,38 +76,41 @@ def gtp(load_file: "The path to the network model files"=None,
...
 
     with timer("Validating from {} to {}".format(os.path.basename(tf_records[0]),
                                                  os.path.basename(tf_records[-1]))):
-        n.validate(tf_records, batch_size=dual_net.TRAIN_BATCH_SIZE,
-                   init_from=load_file, num_steps=num_steps)
+        dual_net.validate(working_dir, tf_records, checkpoint_name=checkpoint_name,
```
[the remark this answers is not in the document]
amj: ack.

dual_net.py
amj: yeah, sorry :(  In theory, they shouldn't change for a run but in practice they might;

amj:
Ok, the code is all LGTM, how should we proceed?  Do we want to do some kind of integration test before merging it in?  Run it on 9x9?

amj:
ok2merge

PR #118, opened 2018-02-26
Initial commit of C++ port.

tommadams:


tommadams:
This is just a port of the Position code; MCTS is still to come.

I've tried to follow the structure of the Python code where it made sense but the actual implementation details of liberty & group tracking is quite different for performance reasons.

Design goals of this port were to avoid heap allocations where possible and to use a compact representation for the board & groups.

A crude benchmark that plays back the same 19x19 game using both the Python and C++ implementations shows that the C++ version is more than 450x faster. The only difference in functionality is that the C++ code does not yet calculate the features tensor.

cc/tf_test.cc
```diff
@@ -0,0 +1,98 @@
...
+  TF_RETURN_IF_ERROR(session->Create(graph_def));
+
+  Tensor input(DT_FLOAT, TensorShape({1, minigo::kN, minigo::kN, 17}));
+  auto x = input.tensor<float, 4>();
+  for (int j = 0; j < minigo::kN; ++j) {
+    for (int i = 0; i < minigo::kN; ++i) {
+      x(0, j, i, 16) = 1;
+    }
+  }
+  for (int j = 0; j < minigo::kN; ++j) {
```
amj: if test code is supposed to serve as a usage example, then some variable names/comments here would be very needed.

tommadams: This is just my first attempt at getting a model to work with the tensorflow C++ API and is not really meant for public consumption. I'll likely delete this whole file once the MCTS player is working.

cc/play_game.cc
```diff
@@ -0,0 +1,72 @@
...
+#include "cc/coord.h"
+#include "cc/position.h"
+
+using minigo::BoardVisitor;
+using minigo::Coord;
+using minigo::GroupVisitor;
+using minigo::Position;
+
+int main() {
+  std::vector<std::string> str_moves = {
```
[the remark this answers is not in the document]
tommadams: This one: http://eidogo.com/#titles/3qcsf2
I've added a link to the source, is there some form of official attribution I should use?

PR #128, opened 2018-03-08
Adds C++ MctsPlayer

tommadams:


cc/mcts_node.cc
```diff
@@ -0,0 +1,254 @@
...
+    if (node->state == State::kCollapsed) {
+      node->state = State::kSelected;
+      return node;
+    }
+    // If the node has already been selected for the next inference batch, we
+    // shouldn't select it again.
+    if (node->state == State::kSelected) {
+      // Revert the visits we just made.
+      --node->stats->N;
+      while (node != this) {
```
brilee: is a do-while loop not kosher?

cc/mcts_node.cc
```diff
@@ -0,0 +1,254 @@
...
+    edges[i].W = value;
+  }
+  BackupValue(value, up_to);
+}
+
+void MctsNode::IncorporateEndGameResult(float value, MctsNode* up_to) {
+  assert(position.is_game_over() || position.n() == kMaxSearchDepth);
+  assert(state == State::kSelected);
+  // We can't expand an end game result, so collapse the node again.
+  state = State::kCollapsed;
```
brilee: oh, this is a clever solution to allow terminal nodes to be revisited repeatedly if necessary.

cc/mcts_player.cc
```diff
@@ -0,0 +1,249 @@
...
+      continue;
+    }
+    if (leaf->position.is_game_over() ||
+        leaf->position.n() >= kMaxSearchDepth) {
+      float value = leaf->position.CalculateScore() > 0 ? 1 : -1;
+      leaf->IncorporateEndGameResult(value, root_);
+    } else {
+      leaf->AddVirtualLoss(root_);
+      leaves.push_back(leaf);
+      if (static_cast<int>(leaves.size()) == batch_size) {
```
brilee: should this happen outside the if-else block?

brilee: I guess this is the only place you append to leaves so it may as well be in here.

cc/mcts_player.cc
```diff
@@ -0,0 +1,249 @@
...
+  // a temperature slightly larger than unity to encourage diversity in early
+  // play and hopefully to move away from 3-3s.
+  if (root_->position.n() < temperature_cutoff_) {
+    // Squash probabilities before normalizing.
+    float sum = 0;
+    for (int i = 0; i < kNumMoves; ++i) {
+      history.search_pi[i] = std::pow(root_->child_N(i), 0.95);
+      sum += history.search_pi[i];
+    }
+    for (int i = 0; i < kNumMoves; ++i) {
```
brilee: move this block to the end to emphasize shared code?

cc/mcts_player.h
```diff
@@ -0,0 +1,145 @@
...
+    int batch_size = 8;
+
+    float komi = kDefaultKomi;
+
+    // Seed used from random permutations.
+    // If the default value of 0 is used, a time-based seed is used.
+    uint64_t random_seed = 0;
+  };
+
+  struct History {
```
brilee: thoughts on whether you want to directly output tfrecords or whether you want to output the search_pi array and postprocess them into tf.records via existing python code? IMO, latter sounds a bit less efficient but also less error prone. It's difficult to detect errors in the data transcription process; we've previously only noticed errors after our training started going wonky after a few days.

cc/mcts_player_test.cc
```diff
@@ -0,0 +1,384 @@
...
+  auto* first_node = player->root()->SelectLeaf();
+  auto output = player->Run(&player->root()->features);
+  first_node->IncorporateResults(output.policy, output.value, player->root());
+  return player;
+}
+
+std::unique_ptr<TestablePlayer> CreateAlmostDonePlayer(
+    MctsPlayer::Options options, int n) {
+  // Always use a deterministic random seed.
+  options.random_seed = 17;
```
brilee: this is the right seed

cc/position.h
```diff
@@ -128,28 +128,44 @@ class GroupVisitor {
...
+           int n = 0);
+
+  // Copies the position's state from another instance, while preserving the
+  // BoardVisitor and GroupVisitor it was constructed with.
+  Position(BoardVisitor* bv, GroupVisitor* gv, const Position& other);
+
   Position(const Position&) = default;
   Position& operator=(const Position&) = default;
 
   void PlayMove(Coord c, Color color = Color::kEmpty);
```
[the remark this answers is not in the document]
amj: c++ has had default args since the 90s i'm pretty sure.  The default move/copy operators i think are as of c++11

cc/position.cc
```diff
@@ -59,23 +54,42 @@ std::array<inline_vector<Coord, 4>, kN* kN> kNeighborCoords = []() {
 inline const inline_vector<Coord, 4>& GetNeighborCoords(Coord c) {
   return kNeighborCoords[c];
 }
+
 }  // namespace
 
+Position::Position(BoardVisitor* bv, GroupVisitor* gv, float komi,
+                   Color to_play, int n)
+    : board_visitor_(bv),
```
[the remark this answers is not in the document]
brilee: (nm, I figured this out - member initialization and such)

cc/color.h
```diff
@@ -28,6 +29,11 @@ enum class Color {
   kWhite,
 };
 
+inline Color OtherColor(Color color) {
+  assert(color != Color::kEmpty);
```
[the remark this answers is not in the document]
tommadams: done

cc/random_test.cc
```diff
@@ -0,0 +1,94 @@
...
+      sum[i] += samples[i];
+    }
+  }
+
+  std::array<float, 40> avg;
+  for (int i = 0; i < 40; ++i) {
+    avg[i] = sum[i] / 10000;
+  }
+
+  EXPECT_NEAR(0.60, avg[0], 0.01);
```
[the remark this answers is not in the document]
tommadams: Thanks :) I was struggled for a little while trying to figure out how best to test that my Dirichlet code was generating reasonable numbers.

cc/mcts_node.h
```diff
@@ -0,0 +1,146 @@
...
+  float original_P() const { return stats->original_P; }
+  float Q() const { return W() / (1 + N()); }
+  float Q_perspective() const {
+    return position.to_play() == Color::kBlack ? Q() : -Q();
+  }
+
+  float child_N(int i) const { return edges[i].N; }
+  float child_W(int i) const { return edges[i].W; }
+  float child_P(int i) const { return edges[i].P; }
+  float original_child_P(int i) const { return edges[i].original_P; }
```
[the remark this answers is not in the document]
tommadams: done

cc/mcts_node.h
```diff
@@ -0,0 +1,146 @@
...
+  float child_Q(int i) const { return child_W(i) / (1 + child_N(i)); }
+  float child_U(int i) const {
+    return kPuct * std::sqrt(1.0f + child_N(i)) * child_P(i) / (1 + child_N(i));
```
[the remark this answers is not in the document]
tommadams: Ouch! Thanks for catching this.

cc/mcts_node.h
```diff
@@ -0,0 +1,146 @@
...
+  // TODO(tommadams): use a better containiner.
+  std::unordered_map<int, std::unique_ptr<MctsNode>> children;
+
+  State state = State::kCollapsed;
+
+  // Current board position.
+  Position position;
+
+  // Input features for this position.
+  DualNet::BoardFeatures features;
```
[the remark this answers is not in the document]
tommadams: The C++ handles the features slightly differently to the Python code. The C++ Position class doesn't keep a history of the board state and so is not able to calculate the input features in isolation. I did it that way because I tried to keep all the MCTS & ML code out of the Position class.

Instead, the MCTS node holds the features and DualNet::UpdateFeatures calculates features for a child node using the parent's features and the child's board state.

When the MctsPlayer plays a move, it does prune all the root's unplayed children, which frees up most of the features arrays. We do keep around the features for the full history of moves played, but this is only around 12MB (19*19*17*500*4) so I don't think we need to worry about it too much.

cc/mcts_node.cc
```diff
@@ -0,0 +1,254 @@
...
+namespace minigo {
+
+constexpr float MctsNode::kPuct;
+
+MctsNode::MctsNode(EdgeStats* stats, const Position& position)
+    : parent(nullptr), stats(stats), move(Coord::kInvalid), position(position) {
+  for (int i = 0; i < kNumMoves; ++i) {
+    illegal_moves[i] = position.IsMoveLegal(i) ? 0 : 1000;
+  }
+  DualNet::InitializeFeatures(position, &features);
```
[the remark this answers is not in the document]
tommadams: This constructor is only called for the root node. The other constructor is used to construct leaf nodes, so we should only call InitializeFeatures once per game.

cc/mcts_node.cc
```diff
@@ -0,0 +1,254 @@
...
+}
+
+MctsNode::MctsNode(MctsNode* parent, Coord move)
+    : parent(parent),
+      stats(&parent->edges[move]),
```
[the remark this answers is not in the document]
tommadams: Yep, that's correct.

I ended up having to write it this way to handle the root node case, because there's no easy way to replicate DummyNode from the Python implementation in C++.

cc/mcts_node.cc
tommadams: I originally tried a do-while loop, but that doesn't handle the case where node == this. You need to break from inside the middle of the loop, which is awkward:
```C++
  while (true) {
    --node->stats->N;
    if (node == this) {
      break;
    }
    node = node->parent;
  }
```

cc/mcts_node.cc
tommadams: A little clever, but I had two bugs in the first version of the code. Fortunately, the awesome Python unit tests caught them when I ported them to C++ :)

cc/mcts_player.cc
tommadams: The two blocks aren't exactly the same: the first copies the squashed values into search_pi, the second doesn't. It's a bit messy, so I was thinking about something like this:
```C++
  if (root_->position.n() < temperature_cutoff_) {
    // Squash probabilities.
    for (int i = 0; i < kNumMoves; ++i) {
      history.search_pi[i] = std::pow(root_->child_N(i), 0.95);
    }
  } else {
    for (int i = 0; i < kNumMoves; ++i) {
      history.search_pi[i] = root_->child_N(i);
    }
  }
  // Normalize.
  float sum = 0;
  for (int i = 0; i < kNumMoves; ++i) {
    sum += history.search_pi[i];
  }
  for (int i = 0; i < kNumMoves; ++i) {
    history.search_pi[i] = root_->child_N(i) / sum;
  }
```

It makes an extra pass over the data but is a little clearer.Given that the C++ code execution time is a tiny fraction of the overall, I'm not too concerned about the extra overhead. What do you think?

cc/mcts_player.h
tommadams: I hadn't given it much thought yet to be honest. I share your concern about potential bugs; one option would be to use the Python code to generate a set of golden outputs to compare the C++ code to.

cc/mcts_player_test.cc
tommadams: :)

cc/mcts_player.cc
brilee: The new version is definitely clearer, LGTM.

cc/mcts_node.h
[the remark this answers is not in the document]
brilee: Ah, I see now - you push the history back 2 slots and then prepend the new state, to incrementally compute the feature planes from the previous position.

cc/mcts_player.cc
tommadams: done

cc/dual_net.cc
```diff
@@ -33,6 +34,50 @@ using tensorflow::error::CANCELLED;
...
+    auto their_stone = stone_color == their_color ? 1 : 0;
+    for (int plane = 0; plane < kPlayerFeature; plane += 2) {
+      (*features)[j++] = my_stone;
+      (*features)[j++] = their_stone;
+    }
+    (*features)[j++] = to_play;
+  }
+}
+
+void DualNet::UpdateFeatures(const BoardFeatures& old_features,
```
amj: Someone later will be happier if there's a comment on this.  It's quite Clever(tm)

cc/mcts_node.cc
```diff
@@ -0,0 +1,254 @@
...
+    if (child_N(i) == 0) {
+      break;
+    }
+    float soft_N = child_N(i) / child_N_sum;
+    float p_delta = soft_N - child_P(i);
+    float p_rel = p_delta / child_P(i);
+    // clang-format off
+    oss << std::left << setw(5) << i.ToKgs() << std::right
+        << ": " << setw(6) << setprecision(3) << child_action_score[i]
+        << ", " << setw(6) << child_Q(i)
```
amj: you don't actually need to propagate my inconsistent string formatting into this version.  sorry :(

cc/mcts_node.cc
```diff
@@ -0,0 +1,254 @@
...
+void MctsNode::PruneChildren(Coord c) {
+  // TODO(tommadams): Allocate children out of a pool and return them here.
+  auto child = std::move(children[c]);
+  children.clear();
+  children[c] = std::move(child);
+}
+
+std::array<float, kNumMoves> MctsNode::CalculateChildActionScore() const {
+  float to_play = position.to_play() == Color::kBlack ? 1 : -1;
+  float U_scale = kPuct * std::sqrt(1.0f + N());
```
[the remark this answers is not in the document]
amj: I'm assuming the relevant bit on the other part got fixed?

cc/mcts_player.cc
```diff
@@ -0,0 +1,247 @@
...
+    std::vector<const DualNet::BoardFeatures*> features;
+    features.reserve(leaves.size());
+    for (auto* leaf : leaves) {
+      features.push_back(&leaf->features);
+    }
+
+    std::vector<DualNet::Output> outputs;
+    outputs.resize(leaves.size());
+
+    RunMany(features, {outputs.data(), outputs.size()});
```
amj: Incidentally, this may need some re-work for TPUs, which may require a constant batch-size for speed.

cc/mcts_player.cc
```diff
@@ -0,0 +1,247 @@
...
+  history.Q = root_->Q();
+
+  // Convert child visit counts to a probability distribution, pi.
+  // For moves before the temperature cutoff, exponentiate the probabilities by
+  // a temperature slightly larger than unity to encourage diversity in early
+  // play and hopefully to move away from 3-3s.
+  if (root_->position.n() < temperature_cutoff_) {
+    // Squash counts before normalizing.
+    for (int i = 0; i < kNumMoves; ++i) {
+      history.search_pi[i] = std::pow(root_->child_N(i), 0.95);
```
amj: Squash is now "0.98," which we arrived at by a combination of superstition and arcane ritualistic incantations

amj:
Is this ready to be merged?

cc/dual_net.cc
tommadams: Done.

cc/mcts_node.cc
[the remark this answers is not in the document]
tommadams: yep

tommadams:
Sorry for the delay, I was focusing on the GTC demo. Merging in a few minutes.

PR #132, opened 2018-03-17
Better TQDM in training_curve

sethtroisi:
TQDM(enumerate(...)) doesn't work because enumerate doesn't have a length
[Per TQDM](https://github.com/tqdm/tqdm/issues/157), enumerate(TQDM(...)) is the magic we want.

Also improved status messages slightly

### DIFF
--- a/oneoffs/training_curve.py
+++ b/oneoffs/training_curve.py
@@ -152,10 +152,10 @@ def find_and_filter_sgf_files(base_dir, min_year = None, komi = None):
   sgf_files = []
   count = 0
   print("Finding all sgf files in {} with year >= {} and komi = {}".format(base_dir, min_year, komi))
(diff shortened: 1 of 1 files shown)


oneoffs/training_curve.py
```diff
@@ -172,15 +172,13 @@ def sample_positions_from_games(sgf_files, num_positions=1):
   move_idxs = []
 
   fail_count = 0
-  for i, path in tqdm(enumerate(sgf_files)):
-    if i % 1000 == 0:
-      print(i)
+  for i, path in enumerate(tqdm(sgf_files, desc="loading sgfs", unit="games")):
     try:
       positions, moves, result, props = parse_sgf(path)
+    except KeyboardInterrupt:
```
sethtroisi: So you can ctrl+c to quit during this loop

amj:
LGTM, merged, thanks!

PR #141, opened 2018-03-29
Add separate configs for building C++ minigo for 9x9 and 19x19 boards.

tommadams:
Split the monolithic //cc:minigo build target into smaller components.

Now instead of having separate minigo_9 and minigo_19 build targets that control the board size, we just have one set of build targets and the board size is controlled by passing either --config=minigo9 or --config=minigo19 on the command line when invoking Bazel.

I've also added a couple of custom minigo_{9,19}_test test rules for tests that depend on a certain build size. When the command line board size config matches the test's required size, the rules generate the test as normal, otherwise they generate an empty test stub.


cc/config/BUILD
```diff
@@ -0,0 +1,19 @@
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+# Build condition label that matches when C++ Minigo is being build for a 9x9
```
amj: nit here and below, build=>built

PR #144, opened 2018-04-01
Add oneoff/heatmap.py and add to oneoff_utils

sethtroisi:
(Do all my commit messages get squashed or should I rebase them? is this an option when accepting a pull request?)

Add the ability to generate policy heatmap data for a set of positions.
This is very useful for a project I'm working on and may be of general interest to the team.

I'll add some notes in a couple of places about why I did things.


oneoffs/heatmap.py
```diff
@@ -0,0 +1,86 @@
...
+import numpy as np
+import tensorflow as tf
+from tqdm import tqdm
+
+import go
+import oneoff_utils
+
+
+tf.app.flags.DEFINE_string(
+    "sgf_dir", "sgf/baduk_db/", "sgf database")
```
sethtroisi: This is 95% not where you want to point but I don't have a better path to default to

sgf_wrapper.py
```diff
@@ -160,7 +160,7 @@ def replay_sgf(sgf_contents):
     komi = 0
     if props.get('KM') != None:
         komi = float(sgf_prop(props.get('KM')))
-    result = utils.parse_game_result(sgf_prop(props.get('RE')))
+    result = utils.parse_game_result(sgf_prop(props.get('RE', '')))
```
sethtroisi: This is needed to replay SGFs without results, it does set the result in the PositionWithContext to '' but I assume that's okay.

amj:
We can either squash and merge, rebase & merge, or create a merge commit.  If there's a lot of little changes i'm likely to prefer to squash.  If you have a preference we can certainly do that.

sethtroisi:
I always prefer squash and merge :)

oneoffs/heatmap.py
amj: this is fine.  The docstring should be better though; "path to directories containing SGFs" or something.

oneoffs/heatmap.py
```diff
@@ -0,0 +1,86 @@
...
+    for pos_name, probs, value in zip(pos_names, eval_probs, eval_values):
+      save_file = os.path.join(data_dir, "eval-{}-{}".format(pos_name, idx))
+
+      with open(save_file, "w") as data:
+        data.write("{},  {},  {}\n".format(
+            idx, value, ",".join(map(str, probs))))
+
+
+def positions_from_sgfs(sgf_files, include_empty=True):
+  # sgf_replay doesn't like empty sgf file
```
amj: i'm not super clear what these lines are doing; why would you want to 'include_empty' ever?  what's that for?

oneoffs/heatmap.py
```diff
@@ -0,0 +1,86 @@
...
+  print("Evaluating models {}-{}, eval_every={}".format(
+        idx_start, len(model_paths), eval_every))
+  for idx in tqdm(range(idx_start, len(model_paths), eval_every)):
+    if idx == idx_start:
+      player = oneoff_utils.load_player(model_paths[idx])
+    else:
+      oneoff_utils.restore_params(model_paths[idx], player)
+
+    pos_names, positions = zip(*eval_positions)
+    eval_probs, eval_values = player.network.run_many(positions)
```
amj: if the positions becomes too large, this call will probably run out of memory.

Also, this will apply a symmetry/rotation transformation at random -- not sure if that's intended or not, but it will make the numbers you get non-repeatable.

oneoffs/heatmap.py
```diff
@@ -0,0 +1,86 @@
...
+tf.app.flags.DEFINE_string("data_dir", "data/eval", "Where to save data")
+tf.app.flags.DEFINE_integer("idx_start", 150,
+                            "Only take models after given idx")
+tf.app.flags.DEFINE_integer("eval_every", 5,
+                            "Eval every k models to generate the curve")
+
+FLAGS = tf.app.flags.FLAGS
+
+
+def eval_for_policy(eval_positions, model_dir, data_dir, idx_start, eval_every):
```
amj: needs a docstring, esp around what gets saved.

oneoffs/oneoff_utils.py
```diff
@@ -1,75 +1,105 @@
...
+    # TODO(sethtroisi): move this to sgf_wrapper.
+    with open(sgf_path) as f:
+        sgf_contents = f.read()
+    collection = sgf.parse(sgf_contents)
+    game = collection.children[0]
+    props = game.root.properties
+    return props
 
 
 def parse_sgf(sgf_path):
```
amj: needs a docstring to distinguish why you'd use this instead of the stuff in sgf_wrapper (should this go in sgf_wrapper?)

oneoffs/oneoff_utils.py
```diff
@@ -1,75 +1,105 @@
...
+            if filename.endswith('.sgf'):
+                path = os.path.join(dirpath, filename)
+                props = get_sgf_props(path)
+                if check_year(props, min_year) and check_komi(props, komi):
+                    sgf_files.append(path)
+    print("Found {} sgf files matching filters".format(len(sgf_files)))
+    return sgf_files
+
+
+def get_model_paths(model_dir):
```
amj: should this and the rl_loop utils around model wrangling move to a common file?

oneoffs/oneoff_utils.py
```diff
@@ -1,75 +1,105 @@
...
+def load_player(model_path):
+    print("Loading weights from %s ... " % model_path)
+    with logged_timer("Loading weights from %s ... " % model_path):
+        network = dual_net.DualNetwork(model_path)
+        network.name = os.path.basename(model_path)
+    player = MCTSPlayer(network, verbosity=2)
+    return player
+
 
+def restore_params(model_path, player):
```
amj: should this be a method on the player?

sgf_wrapper.py
amj: I don't know what this sentence means.  What does it mean to replay an SGF without a result?  what problem is this working around?

oneoffs/heatmap.py
sethtroisi: Yes I'm aware of the rotation problem. Maybe in a later version I'll come up with something better.
I haven't written the batching code because it's not broken for me yet. I can fix it if you deem it important but I'd prefer to punt till later.

oneoffs/oneoff_utils.py
sethtroisi: This can be completely replaced by sgf_wrapper. It was used by training_curve and blindly copied by me. I'll fix this after I submit this PR (which only changed the indention of this block)

oneoffs/heatmap.py
sethtroisi: Updated comment, I want to be able to analyse the empty board position (what does the policy heatmap look like when no stones are on the board) but sgf_replay considers this an error.

oneoffs/oneoff_utils.py
sethtroisi: Probably but that's for a follow up. These were already present in training_curve and I'd like to submit this PR with as few modifications to the base directory as possible.

oneoffs/heatmap.py
amj: You can pass in 'false' and it won't apply the transforms.

You don't have to fix it but you should make a note of it in the docstring.

oneoffs/oneoff_utils.py
sethtroisi: Probably but I think this is more appropriate for a follow up

sgf_wrapper.py
sethtroisi: any SGF file without a RE tag will throw an error here.
For example a study of opening variations or an ongoing game. I can workaround this by add RE[B+R] to all my files (even though they are only three moves in) but I thought this was nicer. I can back out this change if you don't like the implications.

oneoffs/heatmap.py
sethtroisi: Done

oneoffs/heatmap.py
```diff
@@ -34,53 +35,65 @@
...
+                data_dir, "heatmap-{}-{}.csv".format(pos_name, idx))
+
+            with open(save_file, "w") as data:
+                data.write("{},  {},  {}\n".format(
+                    idx, value, ",".join(map(str, probs))))
+
+
+def positions_from_sgfs(sgf_files, include_empty_position=True):
+    data = []
+    if include_empty:
```
amj: Nit, no change needed -- but it seems weird that you're putting this here, FWIW.  Seems like a "give me some Position objects from SGFs" is a very odd place to put "oh yeah and throw in a blank board position too" -- that might be better off where you use it rather than adding it as a special case here.  YMMV.

amj:
ok LGTM, LMK if you want to make any further changes or if this is ok2merge

PR #147, opened 2018-04-02
Add a 'model zoo' evaluation watcher

amj:
Currently run after running `kubectl proxy`

As the commit says, this is probably the worst way to schedule jobs.  Although, the code to watch the API output directly seems cheap and effective, and also provides a basic way to clean up after ourselves.  This seems way simpler than a Redis/Celery solution -- what do you think?


cluster/evaluator/gpu-evaluator.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: evaluator
+        image: gcr.io/$PROJECT/$GPU_EVALUATOR_CONTAINER:$VERSION_TAG
+        imagePullPolicy: Always
+        resources:
+          limits:
+            nvidia.com/gpu: 1
+          requests:
+            nvidia.com/gpu: 1
+        volumeMounts:
+        - name: service-credentials
```
Kashomon: Seems like these should be passed in as Secrets. https://kubernetes.io/docs/concepts/configuration/secret/

amj: does this change all the gpu-cluster-up scripts, then?

cluster/evaluator/launch_eval.py
```diff
@@ -0,0 +1,119 @@
...
+    b = rl_loop.get_model(black_num)
+    w = rl_loop.get_model(white_num)
+
+    b_model_path = os.path.join(rl_loop.MODELS_DIR, b)
+    w_model_path = os.path.join(rl_loop.MODELS_DIR, w)
+
+    kubernetes.config.load_kube_config()
+    configuration = kubernetes.client.Configuration()
+    api_instance = kubernetes.client.BatchV1Api(
+        kubernetes.client.ApiClient(configuration))
```
Kashomon: Since this is the default client, you don't need this and you don't need to create the configuration object. Just kubernetes.client.BatchV1Api, and it'll use the default configuration object.

cluster/evaluator/launch_eval.py
```diff
@@ -0,0 +1,119 @@
...
+        print("Need real model numbers")
+        return
+
+    b = rl_loop.get_model(black_num)
+    w = rl_loop.get_model(white_num)
+
+    b_model_path = os.path.join(rl_loop.MODELS_DIR, b)
+    w_model_path = os.path.join(rl_loop.MODELS_DIR, w)
+
+    kubernetes.config.load_kube_config()
```
Kashomon: As mentioned in chat, this + the api instance should be called more regularly. load_kube_config refreshes GCP tokens; then, these get cached inside the configuration object.

cluster/evaluator/launch_eval.py
```diff
@@ -0,0 +1,119 @@
...
+                                   'bnum': white_num}))
+
+    resp = api_instance.create_namespaced_job('default', body=job_conf)
+
+
+def zoo_loop():
+    desired_pairs = restore_pairs()
+    random.shuffle(desired_pairs)
+
+    kubernetes.config.load_kube_config(persist_config=True)
```
Kashomon: rm the persist_config. Seems not to help anyway and adds noise.

cluster/evaluator/launch_eval.py
```diff
@@ -0,0 +1,119 @@
...
+        launch_eval(model_num, model_num - i)
+    for i in range(10, 20, 2):
+        launch_eval(model_num, model_num - i)
+    for i in range(20, 51, 5):
+        launch_eval(model_num, model_num - i)
+
+
+if __name__ == '__main__':
+    argh.dispatch_command(zoo_loop)
+    #for i in range(270, 280, 1):
```
Kashomon: Remove comments

cluster/evaluator/gpu-evaluator.yaml
amj: These are now passed in as secrets.

cluster/evaluator/launch_eval.py
amj: Doesn't, actually...

amj:
Ok this has been updated to use the new code and is running on v7.  PTAL/ok2merge?

cluster/evaluator/Dockerfile.gpu
```diff
@@ -0,0 +1,46 @@
+FROM nvidia/cuda:9.0-base-ubuntu16.04
+#FROM tensorflow/tensorflow:1.7.1-gpu-py3
+
+RUN apt update && apt-get install -y --no-install-recommends \
+        build-essential \
+        cuda-command-line-tools-9-0 \
```
Kashomon: Do you still need all these cuda libraries?

cluster/evaluator/launch_eval.py
```diff
@@ -0,0 +1,149 @@
+import sys
```
Kashomon: TODO: Put in a pod?

Kashomon: +license

PR #151, opened 2018-04-04
Backfill global step

amj:
Provides a nice wrapper around the snippet to convert old-style models to new-style, and an rl_loop command to automatically backfill all of them.


sethtroisi:
:+1:  :) :100:

amj:
@brilee That kinda makes sense; i'd like `convert` to still be in main for new folks who are just getting started with it.  Once all the models are converted over we can nuke the code entirely, so i'm not really concerned; this code has a hard expiration date.

main.py
```diff
@@ -230,9 +230,40 @@ def gather(
...
+    tf.train.Saver(var_list=var_list).restore(sess, load_file)
+
+    # manually set the global step
+    global_step_tensor = tf.train.get_or_create_global_step()
+    assign_op = tf.assign(global_step_tensor, global_step_value)
+    sess.run(assign_op)
+
+    # export a new savedmodel that has the right global step type
+    tf.train.Saver().save(sess, dest_file)
+
```
sethtroisi: `sess.close()`
`tf.reset_default_graph()`

rl_loop.py
```diff
@@ -177,14 +179,36 @@ def validate(model_num=None, validate_name=None):
                   validate_name=validate_name)
 
 
+def backfill():
+    models = [m[1] for m in get_models()]
+
+    import dual_net
```
sethtroisi: from tqdm import tqdm

rl_loop.py
```diff
@@ -177,14 +179,36 @@ def validate(model_num=None, validate_name=None):
...
+    import tensorflow as tf
+    from tensorflow.python.framework import meta_graph
+    features, labels = dual_net.get_inference_input()
+    dual_net.model_fn(features, labels, tf.estimator.ModeKeys.PREDICT,
+                      dual_net.get_default_hyperparams())
+
+    for model_name in models:
```
sethtroisi: tqdm(models):

rl_loop.py
amj: Done

rl_loop.py
amj: Done

PR #155, opened 2018-04-05
Adds basic C++ SGF parsing and generation.

tommadams:


cc/coord.cc
```diff
@@ -102,6 +102,20 @@ std::string Coord::ToKgs() const {
...
+    return "";
+  } else if (*this == kInvalid) {
+    return "invalid";
+  }
+  int row = value_ / kN;
+  int col = value_ % kN;
+  char buffer[2];
+  buffer[0] = 'a' + col;
+  buffer[1] = 'a' + row;
+  return {buffer, 2};
```
amj: how am i still learning syntax tricks.  Thank you.

cc/move.cc
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cc/move.h"
+
+namespace minigo {
+
+std::ostream& operator<<(std::ostream& os, const Move& move) {
```
amj: I would probably expect `cout << move` to print a KGS/human-style coord rather than an SGF-style one, e.g. `W P16` vs `W[qq]`.  but i don't have any good solution for what to do with the << operator for making sgfs if we do that.  (no action needed, but made me go hmm)

cc/move.h
```diff
@@ -0,0 +1,41 @@
...
+#define CC_MOVE_H_
+
+#include <iostream>
+
+#include "cc/color.h"
+#include "cc/coord.h"
+
+namespace minigo {
+
+struct Move {
```
amj: oh, if these Move objects are only used for SGFs (and not e.g. in tree search), maybe call this out in a comment (or move it under a cc/sgf/ dir?)

cc/sgf_test.cc
```diff
@@ -0,0 +1,185 @@
...
+TEST(SgfTest, CreateSgfStringMoves) {
+  CreateSgfOptions options;
+  options.result = "B+R";
+
+  std::vector<MoveWithComment> moves = {
+      {Color::kBlack, Coord::FromSgf("be"), ""},
+      {Color::kWhite, Coord::FromSgf("aa"), "Hello there"},
+      {Color::kBlack, Coord::FromSgf("hb"), ""},
+      {Color::kWhite, Coord::FromSgf("ge"), "General Kenobi"},
+      {Color::kBlack, Coord::kPass, "You are a bold one"},
```
amj: :thinking:

PR #157, opened 2018-04-12
new oneoff script to constructing PVs of a set of positions

sethtroisi:
I'm using FLAGS as a global instead of passing it around which is just easier. Given these scripts are <100 lines I think it's nice to avoid LONG parameter lists.


amj:
LGTM, thanks :)

PR #161, opened 2018-04-17
"Freeze" graph for inference after training, bootstrap

amj:
also adds command to main.py

### DIFF
--- a/main.py
+++ b/main.py
@@ -263,8 +263,18 @@ def convert(load_file, dest_file):
     tf.reset_default_graph()
 
 
+def export_graph(load_file, dest_dir):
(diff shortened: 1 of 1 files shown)


brilee:
This is a pretty nice snippet of code; I'd be happy with putting it as part of dual_net.export_model so that it's all together.

main.py
```diff
@@ -263,8 +265,17 @@ def convert(load_file, dest_file):
     tf.reset_default_graph()
 
 
+def freeze_graph(load_file):
```
sethtroisi: would this be better named "load_model" or "in_model"?

PR #163, opened 2018-04-17
Rotate every training batch

sethtroisi:
This is the simple version that performs a random per-batch-rotation (same rotation for every item in that batch)

Code is ready for initial review, I wanted to rotate every position but had a harder time doing this (see [my attempt](https://github.com/tensorflow/minigo/compare/master...sethtroisi:training_rotate_every#diff-4f2aa60d95ea26583846318941e5424cR123)

For per position rotation 
This passes the symmetry as a tf.string which can't be used to lookup TF_IMPL. In short I'm not sure how to select a different op passed on a tensor. but maybe this making can be moved before the tf batch read in main or something


symmetries.py
```diff
@@ -80,3 +96,28 @@ def randomize_symmetries_feat(features):
...
+    pi_n_by_n = tf.reshape(pi_move, [batch_size, 1, go.N, go.N])
+    pi_rot = tf.concat(
+        [tf.reshape(impl(pi_n_by_n), [batch_size, squares]), pi_pass],
+        1)
+
+    return impl(x), pi_rot
+
+
+def randomize_symmetries_train(batch_size, x, pi):
+    symmetry = 'identity' # 'fliprot270' # rot180a#random.choice(SYMMETRIES)
```
sethtroisi: this needs to be updated

sethtroisi:
This was superseded by https://github.com/tensorflow/minigo/pull/166

brilee:
Reviving this because it seems that TF impls of rotations are necessary for cloud TPU.

https://github.com/tensorflow/tensorflow/blob/r1.8/tensorflow/python/ops/image_ops_impl.py#L440 for some inspiration on how to efficiently do this

brilee:
On the topic of doing rotations per image or per batch - I'm thinking there are two things we want to accomplish with random rotations
- In a given generation, we want the handful of positions from the same game to be randomized from each other, so that the net doesn't realize that it's seeing the same game multiple times.
- For batchnorm statistics, we'd like a batch to consist of random rotations, in order to avoid potential weirdness with one batch being entirely one rotation, etc.

brilee:
To expand more on how the cloud TPU resnet example does things:
https://github.com/tensorflow/tpu/blob/5f44b6141cea377a46c912c6b557047922ecdce9/models/official/resnet/imagenet_input.py#L175
```
    # Parse, preprocess, and batch the data in parallel
    dataset = dataset.apply(
        tf.contrib.data.map_and_batch(
            self.dataset_parser, batch_size=batch_size,
            num_parallel_batches=8,    # 8 == num_cores per host
            drop_remainder=True))
```
the function `self.dataset_parser` operates on one image and randomly flips it/crops it; it then seems that `tf.contrib.data.map_and_batch` is some special sauce that distributes the random flipping/cropping operations across the 8 cpus of the host. The returned dataset represents batches where each position is independently and randomly transformed.

PR #164, opened 2018-04-18
Add oneoff script to process debug SGFs

brilee:
@sethtroisi @amj take a look?

### DIFF
--- a/oneoffs/prepare_bigquery.py
+++ b/oneoffs/prepare_bigquery.py
@@ -1,0 +1,178 @@
+'''
+Script to process debug SGFs for upload to BigQuery.
+The load commands look like:
+bq load --project_id=$PROJECT_ID \
+    --source_format=NEWLINE_DELIMITED_JSON \
+    $PROJECT_ID:minigo_v5_19.games \
+    gs://$BUCKET_NAME/bigquery/holdout/games/* \
+    oneoffs/bigquery_games_schema.json
+
+bq load --project_id=$PROJECT_ID \
+    --source_format=NEWLINE_DELIMITED_JSON \
+    $PROJECT_ID:minigo_v5_19.moves \
+    gs://$BUCKET_NAME/bigquery/holdout/moves/* \
+    oneoffs/bigquery_moves_schema.json
+
+'''
+import collections
+import itertools
+import json
+import os
+import re
+import sys; sys.path.insert(0, '.')
+import sgf
+from tensorflow import gfile
+
+import coords
+import shipname
+import sgf_wrapper
+import utils
+
+DebugRow = collections.namedtuple('DebugRow', [
+    'move', 'action', 'Q', 'U', 'prior', 'orig_prior', 'N', 'soft_N', 'p_delta', 'p_rel'])
+
+BUCKET_NAME = os.environ['BUCKET_NAME']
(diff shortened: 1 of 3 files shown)


oneoffs/prepare_bigquery.py
```diff
@@ -0,0 +1,185 @@
...
+    komi = float(sgf_wrapper.sgf_prop(props.get('KM')))
+    result = sgf_wrapper.sgf_prop(props.get('RE', ''))
+    board_size = int(sgf_wrapper.sgf_prop(props.get('SZ')))
+    value = utils.parse_game_result(result)
+    was_resign = '+R' in result
+    
+    filename = os.path.basename(gcs_path)
+    filename_no_ext, _ = os.path.splitext(filename)
+    completion_time = int(filename_no_ext.split('-')[0])
+    worker_id = filename_no_ext.split('-')[-1]
```
sethtroisi: BTW completion_time & worker_id aren't 100% unique over the self-play games.
I know at least one of the overlaps was a v3 game overlapping with a v5 game

oneoffs/prepare_bigquery.py
```diff
@@ -0,0 +1,185 @@
...
+    sgf_url = gcs_path
+    first_comment_node_lines = root_node.next.properties['C'][0].split('\n')
+    # in-place edit to comment node so that first move's comment looks
+    # the same as all the other moves.
+    root_node.next.properties['C'][0] = '\n'.join(first_comment_node_lines[1:])
+    resign_threshold = float(first_comment_node_lines[0].split()[-1])
+
+    return {
+        'worker_id': worker_id,
+        'completed_time': completion_time * 1000, # BigQuery's TIMESTAMP() takes in unix millis.
```
sethtroisi: I think this is messing up the big query results. I can't select completed_time right now

oneoffs/prepare_bigquery.py
```diff
@@ -0,0 +1,185 @@
...
+        move_data.append({
+            'worker_id': worker_id,
+            'completed_time': completed_time,
+            'move_num': move_num,
+            'turn_to_play': to_play,
+            'move': move_played,
+            'move_kgs': coords.to_kgs(coords.from_flat(move_played)),
+            'prior_Q': None,
+            'post_Q': post_Q, 
+            'policy_prior': policy_prior,
```
sethtroisi: This seems to result in each Move row being 11,000 Bytes (according to BigQuery).
Consider either not including this data or including the index of the move so that you only have to store 10 policy_priors (instead of 361)

oneoffs/prepare_bigquery.py
brilee: Wow. Damn birthday paradox =/

Well worst thing that can happen is that those games are counted 2x in the bulk statistics, but it should be super rare, and so it shouldn't affect them significantly.

oneoffs/prepare_bigquery.py
brilee: Yeah, this was originally 1000000 instead of 1000. I can do a oneoff fix to cast back to ms, divide by 1000, and then recast to timestamp.

oneoffs/prepare_bigquery.py
brilee: I want to keep the schema general because I intend for the selfplay workers to emit full debug info, not just info on the top 10.

PR #165, opened 2018-04-19
Adds GTP & Minigui support to C++ code

tommadams:


cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+bool GtpPlayer::HandleCmd(const std::string& line) {
+  std::vector<absl::string_view> args =
+      absl::StrSplit(line, absl::ByAnyChar(" \t\r\n"), absl::SkipWhitespace());
+  if (args.empty()) {
+    std::cout << "=" << std::endl;
+    return true;
+  }
+
+  auto cmd = args[0];
+  args.erase(args.begin());
```
sethtroisi: maybe a quick comment `// erase the gtp command from arg list`

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+
+GtpPlayer::GtpPlayer(std::unique_ptr<DualNet> network, const Options& options)
+    : MctsPlayer(std::move(network), options),
+      num_readouts_(options.num_readouts) {}
+
+bool GtpPlayer::HandleCmd(const std::string& line) {
+  std::vector<absl::string_view> args =
+      absl::StrSplit(line, absl::ByAnyChar(" \t\r\n"), absl::SkipWhitespace());
+  if (args.empty()) {
+    std::cout << "=" << std::endl;
```
sethtroisi: GTP spec calls for an "id" see 2.6
http://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html

Other implementations I've seen all add a print_gtp(int id, {var args, string}); method
see http://www.lysator.liu.se/~gunnar/gtp/gtp.c

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+  args.erase(args.begin());
+
+  if (cmd == "quit") {
+    std::cout << "=" << std::endl;
+    return false;
+  }
+
+  auto result = DispatchCmd(cmd, args);
+  if (result == "") {
+    std::cout << "=" << std::endl;
```
sethtroisi: Seems like this is the failure case which is notated with "? "

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+  } else if (cmd == "genmove" || cmd == "mg_genmove") {
+    return HandleGenmove(cmd, args);
+  } else if (cmd == "play") {
+    return HandlePlay(cmd, args);
+  } else if (cmd == "readouts") {
+    return HandleReadouts(cmd, args);
+  } else if (cmd == "report_search_interval") {
+    return HandleReportSearchInterval(cmd, args);
+  }
+  std::cerr << "ERROR: unrecognized GTP command \"" << cmd << "\"\n";
```
sethtroisi: http://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html

2.6 err should start with "?[Id] "

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+  if (cmd == "quit") {
+    std::cout << "=" << std::endl;
+    return false;
+  }
+
+  auto result = DispatchCmd(cmd, args);
+  if (result == "") {
+    std::cout << "=" << std::endl;
+  } else {
+    std::cout << "= " << result << std::endl;
```
sethtroisi: http://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html
2.6
Here '=' indicates success, id is the identity number given in the command, and result is a piece of text ending with **two consecutive newlines**.

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+                                  const std::vector<absl::string_view>& args) {
+  return absl::StrJoin(args, " ");
+}
+
+std::string GtpPlayer::HandleFinalScore(
+    absl::string_view cmd, const std::vector<absl::string_view>& args) {
+  if (!CheckArgsExact(cmd, 0, args)) {
+    return "";
+  }
+  return absl::StrJoin(args, " ");
```
sethtroisi: This doesn't seem finished

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+
+std::string GtpPlayer::HandleGenmove(
+    absl::string_view cmd, const std::vector<absl::string_view>& args) {
+  if (!CheckArgsRange(cmd, 0, 1, args)) {
+    return "";
+  }
+
+  auto c = SuggestMove(num_readouts_);
+  std::cerr << root()->Describe() << std::endl;
+  if (ShouldResign()) {
```
sethtroisi: out of ignorance does pass also need special handling

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+}
+
+std::string GtpPlayer::HandlePlay(absl::string_view cmd,
+                                  const std::vector<absl::string_view>& args) {
+  if (!CheckArgsExact(cmd, 2, args)) {
+    return "";
+  }
+
+  Color color;
+  if (args[0] == "b" || args[0] == "B") {
```
sethtroisi: I'm realizing that maybe you don't want a pedantic review so I'll stop after this comment

http://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html
3.2.1 says color can be any of {"b", "w", "black", "white"} case insensitive

cc/main.cc
```diff
@@ -142,55 +151,83 @@ void WriteSgf(const std::string& output_dir, const std::string& output_name,
...
-    options.resign_threshold = FLAGS_resign_threshold;
   }
+  auto player = absl::make_unique<MctsPlayer>(std::move(dual_net), options);
 
-  MctsPlayer player(&dual_net, options);
-
-  // Play the game.
-  player.SelfPlay(FLAGS_num_readouts);
+  while (!player->game_over()) {
+    auto move = player->SuggestMove(FLAGS_num_readouts);
```
sethtroisi: consider options.num_readouts

cc/main.cc
```diff
@@ -142,55 +151,83 @@ void WriteSgf(const std::string& output_dir, const std::string& output_name,
...
   MctsPlayer::Options options;
-  options.random_seed = FLAGS_seed;
-  options.komi = FLAGS_komi;
-  options.inject_noise = FLAGS_inject_noise;
-  options.soft_pick = FLAGS_soft_pick;
-  options.random_symmetry = FLAGS_random_symmetry;
-  options.batch_size = FLAGS_batch_size;
-  if (rnd() < FLAGS_holdout_pct) {
+  ParseMctsPlayerOptionsFromFlags(&options);
+  Random rnd;
```
sethtroisi: should rnd be seeded with option.random_seed?

cc/mcts_node.h
```diff
@@ -34,6 +35,7 @@ class MctsNode {
   static constexpr float kPuct = 1.38;
```
sethtroisi: should this be moved to constants.h?

cc/mcts_player.cc
```diff
@@ -36,70 +35,61 @@ MctsPlayer::MctsPlayer(DualNet* network, const Options& options)
...
-    if (ShouldResign()) {
-      SetResult(root_->position.to_play() == Color::kBlack ? -1 : 1,
-                root_->position.CalculateScore(),
-                GameOverReason::kOpponentResigned);
-      break;
-    }
+    rnd_.Dirichlet(kDirichletAlpha, &noise);
+    root_->InjectNoise(noise);
+  }
+  int current_readouts = root_->N();
```
sethtroisi: consider `auto goal_readouts = root_->N() + num_readouts;`

cc/mcts_player.cc
```diff
@@ -36,70 +35,61 @@ MctsPlayer::MctsPlayer(DualNet* network, const Options& options)
...
-      float result = score < 0 ? -1 : score > 0 ? 1 : 0;
-      SetResult(result, score, GameOverReason::kMoveLimitReached);
-      break;
-    }
+  auto start = absl::Now();
+  while (root_->N() < current_readouts + num_readouts) {
+    TreeSearch(options_.batch_size);
+  }
+  auto elapsed = absl::Now() - start;
+  std::cerr << "Seconds per 100 reads: " << elapsed * 100 / num_readouts
```
sethtroisi: does this need set_precision?

cc/mcts_player.cc
```diff
@@ -138,46 +128,84 @@ void MctsPlayer::TreeSearch(int batch_size) {
...
+  if (!leaves_.empty()) {
+    features_.clear();
+    features_.reserve(leaves_.size());
+    for (auto* leaf : leaves_) {
+      features_.push_back(&leaf->features);
     }
 
-    std::vector<DualNet::Output> outputs;
-    outputs.resize(leaves.size());
+    outputs_.resize(leaves_.size());
```
sethtroisi: can you get away with `reserve` as the values don't need to be initialized?
This might be premature optimization but I think this is on the critical/hot path

cc/mcts_player.h
```diff
@@ -37,11 +38,8 @@ class MctsPlayer {
     bool inject_noise = true;
     bool soft_pick = true;
     bool random_symmetry = true;
-
-    float resign_threshold = -0.9;
-
+    float resign_threshold = -0.95;
```
sethtroisi: consider moving to constants.h

cc/gtp_player.cc
```diff
@@ -0,0 +1,323 @@
...
+  }
+  return true;
+}
+
+absl::Span<MctsNode* const> GtpPlayer::TreeSearch(int batch_size) {
+  auto leaves = MctsPlayer::TreeSearch(batch_size);
+  if (!leaves.empty() && report_search_interval_ != absl::ZeroDuration()) {
+    auto now = absl::Now();
+    if (now - last_report_time_ > report_search_interval_) {
+      last_report_time_ = now;
```
sethtroisi: consider moving this after the reportsearchstatus. 
On windows some terminals are slow (100 lines per / second) and if half of the duration was taken up by printing the stats might be confusing. but I can also see an arguement that you want output every 1 second and if it only prints every 1.2s that might be weird

minigui/serve.py
```diff
@@ -17,14 +17,23 @@
...
-               "-v", "2"]
+# GTP_COMMAND = ["python",  "-u",  # turn off buffering
+#                "main.py", "gtp",
+#                "--load-file", MODEL_PATH,
+#                "--readouts", "1000",
+#                "-v", "2"]
+
+GTP_COMMAND = [
+    "bazel-bin/cc/main",
+    "--model=saved_models/000483-indus-upgrade.pb",
```
sethtroisi: "--model=" + MODEL_PATH + ".pb"

minigui/serve.py
```diff
@@ -48,6 +57,7 @@ def std_bg_thread(stream):
 
     for line in p.__getattribute__(stream):
         line = line.decode()
+        # print("###", stream, line[:-1])
```
sethtroisi: remove

cc/gtp_player.cc
tommadams: The id is optional according to 2.5 in the spec and I'm following the observed behaviour of our Python implementation.

Happy to add id support later along with more GTP features that I'm missing. This CL is just getting enough support in place to run Minigui (which does not expect IDs).

cc/gtp_player.cc
tommadams: done

cc/gtp_player.cc
tommadams: This is the case where there's output (e.g. play), in which case we want to avoid a trailing whitespace :)
I've added "?" support for errors though.

cc/gtp_player.cc
tommadams: done

cc/gtp_player.cc
tommadams: indeed it wasn't

cc/gtp_player.cc
tommadams: well that's just silly

cc/gtp_player.cc
tommadams: no. resign doesn't need special handling now either :)

cc/gtp_player.cc
tommadams: updated to accept anything that begins with 'b' or 'w' :P

cc/main.cc
tommadams: done

cc/main.cc
tommadams: I don't think so. random_seed controls the determinism of the moves played, not whether the game is selected for holdout etc. Updated the --seed flag's help text.

cc/mcts_node.h
tommadams: done

cc/mcts_player.cc
tommadams: no, you can't legally access the elements of a vector if you only call reserve. This isn't an issue because in practice outputs_ is almost always the same size.

minigui/serve.py
tommadams: done

cc/gtp_player.cc
```diff
@@ -0,0 +1,324 @@
...
+}
+
+GtpPlayer::Response GtpPlayer::HandleFinalScore(
+    absl::string_view cmd, const std::vector<absl::string_view>& args) {
+  auto response = CheckArgsExact(cmd, 0, args);
+  if (!response.ok) {
+    return response;
+  }
+  if (!game_over()) {
+    return Response::Error("game isn't over");
```
amj: For unfinished games, i think correct behavior is to just return the score of the board.  Various GTP harnesses may depend on this behavior.

(Correct failure behavior per spec is to fail with the message "cannot score")

cc/gtp_player.cc
```diff
@@ -0,0 +1,324 @@
...
+    return response;
+  }
+
+  Color color;
+  if (std::tolower(args[0][0]) == 'b') {
+    color = Color::kBlack;
+  } else if (std::tolower(args[0][0]) == 'w') {
+    color = Color::kWhite;
+  } else {
+    return Response::Error("expected b or w for player color, got ", args[0]);
```
amj: You might think it is handy to be able to fail with useful, descriptive error messages.  but alas, the spec constrains the error messages.  For these other cases, printing something actually useful on stderr and failing on stdout with "illegal move" is correct behavior.

cc/gtp_player.cc
tommadams: done

cc/gtp_player.cc
tommadams: done

PR #166, opened 2018-04-19
Rotate dataset with py_func

sethtroisi:
I'm not sure how to test this but this implementation is much cleaner than https://github.com/tensorflow/minigo/pull/163


preprocessing.py
```diff
@@ -77,8 +74,6 @@ def write_tf_examples(filename, tf_examples, serialize=True):
             else:
                 writer.write(ex)
 
-# Read tf.Example from files
```
amj: why are you removing these?

preprocessing.py
```diff
@@ -150,6 +144,17 @@ def read_tf_records(batch_size, tf_records, num_repeats=None,
     return dataset
 
 
+def _rotate_pyfunct(x, pi, value):
```
amj: Can we get a test for this?

preprocessing.py
sethtroisi: They are no longer adjacent to the functions they describe.

sethtroisi:
I have tested the new code: with local_rl_loop.py and with some qualityish tests

tests/test_preprocessing.py
```diff
@@ -146,3 +150,42 @@ def test_make_dataset_from_sgf(self):
...
+        def test_symmetry(s, i, r1, r2):
+            # features
+            x = (r1[0][i] == symmetries.apply_symmetry_feat(sym, r2[0][i])).all()
+            # pi
+            pi = (r1[1][i] == symmetries.apply_symmetry_pi(sym, r2[1][i])).all()
+            # value
+            v = (r1[2][i] == r2[2][i]).all()
+            return x and pi and v
+
+        num_records = 200
```
sethtroisi: I mention this only because I want to link to the math problem but this code has ~1/1e10 chance of failing
see https://en.wikipedia.org/wiki/Coupon_collector%27s_problem

amj: can you use tf.set_random_seed to make the test deterministic?

tests/test_preprocessing.py
```diff
@@ -146,3 +150,42 @@ def test_make_dataset_from_sgf(self):
...
+        raw_data = self.create_random_data(num_records)
+        tfexamples = list(map(preprocessing.make_tf_example, *zip(*raw_data)))
+
+        with tempfile.NamedTemporaryFile() as f:
+            preprocessing.write_tf_examples(f.name, tfexamples)
+            run_one = self.extract_data(
+                f.name, filter_amount=1, random_rotation=False)
+            run_two = self.extract_data(
+                f.name, filter_amount=1, random_rotation=True)
+
```
amj: This code will pass even if random_rotation does nothing; nowhere do you assert that run_one's data and run_two's data are actually different.  Just add an assert that run_one != run_two

tests/test_preprocessing.py
amj: (tf.set_random_seed and in conjunction with random.seed() )

sethtroisi: I'll try but ugh.

tests/test_preprocessing.py
sethtroisi: Below I verify that I find all eight rotations which wouldn't pass but it's easy to add.

amj: That's not quite what it does -- it just finds *a* rotation that matches, then it falls out of the loop.  If the identity rotation passed (because run_one == run_two), the loop would still be exited w/o failing the test.

amj: you're not quite verifying the behavior you think you're verifying, is what i'm sayin ;)

sethtroisi: Disagree, look at line 191

```
difference = set(syms) ^ set(symmetries.SYMMETRIES)
assert not difference, "Didn't find {}".format(difference)
```

brilee:
BTW, this may just be being overly cute, but I had the thought that if you put this in the core model_fn rather than in the input_fn part, then you could reuse this code for dual_net.run_many(use_random_symmetry=True)

PR #167, opened 2018-04-19
Add flags infrastructure to main.py

brilee:
This is a deceptively simple PR that took me like 5 hours of researching the interaction between argparse, arghparse, and absl.flags and trying to figure out how to create a flag file.

Anyway, this is the first commit. This is backwards compatible with the way we've been calling everything, and allows us to gradually swap out hyperparameters one by one by replacing them with flags.DEFINE declarations. I'll submit a few of these as examples on how to swap them out.

The only downside I've found so far is that this approach doesn't allow directly loading flagfiles from GCS; they must be gsutil cp'd to the host before passing --flagfile=local/path/to/flagfile as a command line flag.

### DIFF
--- a/main.py
+++ b/main.py
@@ -23,6 +23,7 @@ import socket
 import sys
 import tempfile
 import time
+from absl import flags
 import cloud_logging
 from tqdm import tqdm
 import gzip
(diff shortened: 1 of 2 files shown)


amj:
I'm very intrigued how this will work w.r.t. our different commands (main.py, rl_loop.py, oneoffs, etc).  Do they all have different flag files?  Do we check them in?

I mean LGTM and let's not hold up the party but i'm curious to see what the working example looks like.

brilee:
The idea is to have one flag file for each experiment type. So ideally the flags are all configured with sensible defaults, and then we can override with --num_mcts_readouts=100 or --num_trunk_layers=10.

I have to think more about rl_loop, local_rl_loop, rl_runner and how we might simplify those. It's a bit annoying to parse flags and then repeatedly pass them down a matroska doll.

PR #169, opened 2018-04-20
Swap out hyperparams in strategies, mcts and set up unit test flag co…

brilee:
…mpatibility

Check it out: some incremental switching out of magic constants for flags.


strategies.py
```diff
@@ -74,7 +78,7 @@ def __init__(self, network, seconds_per_move=5, simulations_per_move=0,
         if two_player_mode:
             self.temp_threshold = -1
         else:
-            self.temp_threshold = TEMPERATURE_CUTOFF
+            self.temp_threshold = FLAGS.softpick_move_cutoff
```
sethtroisi: unrelated temp_threshold appears to be used inconsistently

https://github.com/tensorflow/minigo/blob/master/strategies.py#L135
https://github.com/tensorflow/minigo/blob/master/strategies.py#L156

I think that L156 should be >=

tests/test_strategies.py
```diff
@@ -16,11 +16,13 @@
 import unittest.mock as mock
 import numpy as np
 
+from absl import flags
+
 import coords
 import go
 from go import Position
+import mcts
```
sethtroisi: unused

strategies.py
amj: it was fixed in the v5-running head.

amj:
ok2merge by me

strategies.py
brilee: done

sethtroisi:
I'm pretty sure this broke tests.

I'm getting errors in master about "absl.flags._exceptions.UnparsedFlagAccessError: Trying to access flag --softpick_move_cutoff before flags were parsed."

brilee:
Which entrypoint are you using? All the entry points now need to have a flags.FLAGS() call somewhere in order to initialize the flag default values and potentially parse cmd line flags. I added them to the entry points I could think of, but I probably missed one.

sethtroisi:
`BOARD_SIZE=9 python -m unittest`

brilee:
Ah, okay. You should use ./test.sh to execute the test runner w/ flag support.

sethtroisi:
Okay, will do from now on thanks,
We should probably update the README file

amj:
+1 let's update the README file asap

strategies.py
```diff
@@ -18,18 +18,22 @@
...
-from mcts import MCTSNode, MAX_DEPTH
 
+import coords
 import go
+import mcts
+import sgf_wrapper
 
-# When to do deterministic move selection.  ~30 moves on a 19x19, ~8 on 9x9
-TEMPERATURE_CUTOFF = int((go.N * go.N) / 12)
+flags.DEFINE_integer('softpick_move_cutoff', (go.N * go.N // 12) // 2 * 2,
```
brilee: cc @tommadams to fix this in C++ player

PR #170, opened 2018-04-20
cleanup imports + autopep8

sethtroisi:
Low priority cleanup


sethtroisi:
I cleaned this up so that it's only the autopep8 part. I have a small refactor to follow

Feel free to "squash and merge" me

PR #172, opened 2018-04-21
Replace num_readouts and resign_threshold with flag versions

brilee:


strategies.py
```diff
@@ -86,7 +93,8 @@ def __init__(self, network, seconds_per_move=5, simulations_per_move=0,
         self.root = None
         self.result = 0
         self.result_string = None
-        self.resign_threshold = -abs(resign_threshold)
+        self.resign_threshold = resign_threshold or FLAGS.resign_threshold
+        self.timed_match = timed_match
```
sethtroisi: consider timed_match or self.num_readouts == 0

cluster/selfplay/player_wrapper.sh
```diff
@@ -23,8 +23,7 @@ echo bucket: $BUCKET_NAME
 echo board_size: $BOARD_SIZE
 
 python3 rl_loop.py selfplay \
-  --resign-threshold=0.99 \
-  --readouts=600
+  --num_readouts=600
```
amj: Resign threshold still needs to go here.

strategies.py
```diff
@@ -86,7 +93,8 @@ def __init__(self, network, seconds_per_move=5, simulations_per_move=0,
         self.root = None
         self.result = 0
         self.result_string = None
-        self.resign_threshold = -abs(resign_threshold)
+        self.resign_threshold = resign_threshold or FLAGS.resign_threshold
```
amj: it seems lke the flag value should take priority over the passed parameter, no?

brilee: fixed this by making it clearer that sometimes we pass in -1 to disable resign threshold. (in which case we override the flag)

strategies.py
brilee: Added an assert to prevent cases where neither time / readout settings are configured.

strategies.py
amj: ok, LGTM :)

amj:
LGTM modulo the merge conflict

PR #173, opened 2018-04-22
Add base_dir flag to prepare_bigquery

sethtroisi:
This is approximately 4x faster (cProfile now says 90% of time is in sgf.parse)


oneoffs/prepare_bigquery.py
```diff
@@ -21,65 +21,81 @@
...
-            for sgf_name in sgf_names:
-                game_data, move_data = extract_data(sgf_name)
-                game_f.write(json.dumps(game_data))
-                game_f.write('\n')
-                for move_datum in move_data:
-                    move_f.write(json.dumps(move_datum))
-                    move_f.write('\n')
-                print('processed {}'.format(sgf_name))
+BASE_GS_DIR = 'gs://{}'.format(BUCKET_NAME)
+HOLDOUT_PATH = '{}/data/holdout/{}'
```
amj: should these be os.path.join?

sethtroisi: Done

brilee:
Wait, i'm curious - what was the exact change that made it 4x faster?

sethtroisi:
I'm now getting 25games/s which is 20x what I get right now.
4x from not reading SGFs from the bucket
5x from [this patch to sgf](https://github.com/jtauber/sgf/compare/master...sethtroisi:master?expand=1#diff-84c2feb09a1cea380f8d9c0f01a71568R184)

the last profile I ran had JSON + disk reading being 60% of the execution time

brilee:
👏 that's pretty awesome.

amj:
LGTM ok2merge

Issue #129, opened 2018-03-13
Offer data as a BigQuery table


brilee:
This will probably start as a private BQ dataset; will have to consult to figure out how to offer data publicly/how to allow the general public to query over a ~1TB dataset for free-ish.

General schema of dataset:
- A table with one row per move in a game (so something like 1 billion moves in total by the end of a run). That row will describe things like worker_id[string], completed_time[time], model_num[int], result[float], was_resign[bool], move_num[int], turn_to_play[int], played_move[int], policy_prior[array<float>], mcts_visit_counts[array<int>], q_prior [float], average_q[float], resign_threshold[float].

This allows queries like " select model_num, first 5 moves of game, count(*) group by model_number, moves", which would show the most popular opening by model number. (And by bigquery magic, that query would probably complete in a few minutes, instead of the hour-long analyses we've been running so far...)

brilee:
How to upload data - can retroactively parse all the full-debug log SGFs we have, and also update selfplay workers to stream data into the table. Might consider separate tables for evaluation zoo games vs. selfplay workers; for 20x less data we could also just do the holdout games.

amj:
The dataset of evaluation games should be much smaller and much easier to make available in its entirety.

tommadams:
It seems like we're duplicating of per-game info (worker_id, completed_time, model_num, result, was_resign, resign_threshold) in the per-move table. Could we instead generate a unique key for each game and store that in the move table?

brilee:
Agreed, we're duplicating some of the info. I think (worker_id, completed_time) should form a primary key, and model_num is too useful; may as well denormalize that. Result/resign/resign threshold could be normalized, but I suspect those are too useful to want to join all the time.

brilee:
@tommadams what do you think this should look like, given that you would have to reimplement some form of this in the C++ version?

tommadams:
Well at that point there's probably not much to gain by normalizing the remaining data because it sounds like the primary game key would be almost as large as the data we're normalizing. So storing everything per-move does make sense.

Your schema sounds good to me. I wonder what the most reasonable import format is. I'm somewhat surprised/disappointed that BigQuery doesn't support proto as far as I can tell.

brilee:
Right, BQ doesn't support protos because they are not self-describing. Avro is a proto-like binary format which is self-describing, and BQ does support Avro. Other import options I'm familiar with are JSON and CSV. (CSV will be utterly incapable of handling the arrays we'd be generating)

[...]

PR #177, opened 2018-04-26
New flag for prepare_bigquery to only include top move

sethtroisi:
run time 14s vs 55s

local json file is 28.1 MB vs 486 MB (6%)
bigquery table is 8 MB vs 1.09 GB (0.7%)

upload is 48s vs ~200s


sethtroisi:
Pros for merging this and using this data in bigquery (https://github.com/tensorflow/minigo/issues/129)
* Smaller data footprint means we can upload all games today (instead of just the 5% holdout games)
* Data about the top 10-15 moves isn't used by any current scripts
* Runs fast enough (5 minutes every ~1.7 hours) that I can put it on my rsync

Cons
* We'll have two tables (moves and top_move_only)

oneoffs/prepare_bigquery.py
```diff
@@ -139,16 +142,22 @@ def extract_move_data(root_node, worker_id, completed_time, board_size):
...
+
+        def get_row_data(debug_row):
+            column_names = ["prior", "orig_prior", "N", "soft_N"]
+            return [getattr(debug_row, field) for field in column_names]
+
+        if FLAGS.only_top_move:
+            assert len(debug_rows) <= 1
+            row_data = list(map(get_row_data, debug_rows))
+        else:
+            row_data = [[0,0,0,0],] * (board_size * board_size + 1)
```
brilee: i'm confused about this trailing comma.
Also, leave a comment that yes, this list is shared between all zero'd row data entries, but since they don't get mutated, it doesn't matter?

brilee:
will merge after comment is addressed

PR #181, opened 2018-04-27
Makes Minigui query the Minigo engine for the board size

### LINKED ISSUE #175: Allow Minigui to work with both board_size=9, 19
Currently I have to toggle the line that says `const N = board.BoardSize.Nine;` and then recompile the TS in order to switch the GUI from board size 9 / 19.

Would like some way to switch in the GUI or even autodetect which size the attached GTP process is using.

tommadams:
Fixes #175 

Some other changes:

- merges the HeatMapBoard class into the Board class, since there's not much point having them separate.
- minigui/serve.py now echos the Minigo subprocess' stdout & stderr until it sees the "GTP engine ready" message, which helps debug problems with the engine configuration.
- add a bunch of command line options to minigo/serve.py.
- update minigui/README.md.


minigui/minigui.ts
```diff
@@ -75,34 +75,77 @@ consoleElem.addEventListener('keypress', (e) => {
...
+
+function tryBoardSize(sizes: board.BoardSize[]) {
+  if (sizes.length == 0) {
+    throw new Error('Couldn\'t find an acceptable board size');
+  }
+  gtp.send(`boardsize ${sizes[0]}`, (cmd, result, ok) => {
+    if (ok) {
+      init(sizes[0]);
+    } else {
+      tryBoardSize(sizes.slice(1));
```
brilee: did you mean (car sizes) and (cdr sizes)? :P

tommadams: i wish :)

PR #184, opened 2018-04-28
Replace num_parallel with flag

brilee:
Flagify parallel readouts in MCTS


tests/test_strategies.py
```diff
@@ -228,7 +228,7 @@ def test_long_game_tree_search(self):
 
         # Test that an almost complete game
```
amj: this comment got truncated somewhere?

PR #190, opened 2018-04-30
train with an incremental number of steps instead of max_steps

amj:
defaulting to POSITIONS_PER_GENERATION // TRAINING_BATCH_SIZE

### DIFF
--- a/preprocessing.py
+++ b/preprocessing.py
@@ -108,7 +108,7 @@ def read_tf_records(batch_size, tf_records, num_repeats=None,
     Args:
         batch_size: batch size to return
         tf_records: a list of tf_record filenames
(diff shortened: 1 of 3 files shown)


preprocessing.py
```diff
@@ -138,7 +138,7 @@ def read_tf_records(batch_size, tf_records, num_repeats=None,
     if num_repeats is not None:
         dataset = dataset.repeat(num_repeats)
     else:
-        dataset = dataset.repeat()
+        dataset = dataset.repeat(1)
```
brilee: just set num_repeats=1 in the default kwargs?

amj: XD d'oh, of course.

amj:
Done; local_rl_loop.py worked fine.

PR #192, opened 2018-04-30
Dockerize Minigui

Kashomon:
This PR makes a Dockerfile for Minigui and a bunch of helper scripts. Plus some small random changes.

Some things that took some time.

1. Getting the cloud-sdk installed was quite the pain, but is working now.
2. Some cleverness in the makefile to help prevent errors.
3. `fetch-and-run.sh` should clear up some of the instructions (maybe it should do `tsc` too).
4. Had some nested-python issues. Had to pass the correct python through to the subprocess.
5. Also some networking issues. At least on OSX, the default listen-ip is 0.0.0.0 for docker. And the default for flask is 127.0.0.1. 127.0.0.1 is private, so I think I can't connect to it if Flask is listening on it. In theory, I think I could do 127.0.0.1 from the outside container and 0.0.0.0 from the inside container.
6. Bash hackery
7. Mounted models as a path-volume in the container to speed up iterations.


cc/README.md
```diff
@@ -14,6 +14,16 @@ library:
...
 ```
 
+If you want to compile for CPU and not GPU, then change `TF_NEED_CUDA` to 0.
+
+```shell
+echo "Configuring tensorflow"
+...
+TF_NEED_CUDA=0 \
+...
+./configure
```
amj: ./configure_tensorflow.sh?

cluster/minigui/Makefile
```diff
@@ -0,0 +1,54 @@
...
+build-clean: check-env build clean
+
+check-env:
+ifndef PROJECT
+	$(error PROJECT is undefined)
+endif
+ifndef MINIGUI_PY_CPU_CONTAINER
+	$(error MINIGUI_PY_CPU_CONTAINER is undefined)
+endif
+ifndef VERSION_TAG
```
amj: checks for MODEL/MODEL_DIR/BOARD_SIZE as per the docs above?

minigui/README.md
```diff
@@ -72,15 +38,18 @@
...
 
-
 1. By default, Minigui will use the Python Minigo engine. You can use the C++
    engine by first compiling it (see the
    [README](https://github.com/tensorflow/minigo/tree/master/cc/README.md)),
    then passing `--engine=cc` when starting `minigui/serve.py`. You will need
    to pass a frozen GraphDef proto as the `--model` command line argument
    instead of the saved parameter data that the Python backend requires.
 
+   **Note:** Compiling tensorflow from scratch can take 2+ hours. So you might
```
amj: depends strongly on number of cores in the machine.  It's about 15-30 minutes on my corpstation

minigui/fetch-and-run.sh
```diff
@@ -0,0 +1,71 @@
...
+echo "Running Minigui!"
+echo "--------------------------------------------------"
+echo "Model: $MINIGUI_MODEL_TMPDIR/$MINIGUI_MODEL.converted"
+echo "Size:  $MINIGUI_BOARD_SIZE"
+
+$MINIGUI_PYTHON minigui/serve.py \
+--model="$MINIGUI_MODEL_TMPDIR/$MINIGUI_MODEL.converted" \
+--board_size="$MINIGUI_BOARD_SIZE" \
+--port=$MINIGUI_PORT \
+--host=0.0.0.0 \
```
amj: I assume the fact this is in docker is what makes global serving ok?

minigui/fetch-and-run.sh
```diff
@@ -0,0 +1,71 @@
...
+for suffix in "${MODEL_SUFFIXES[@]}"
+do
+  file_to_check="${MINIGUI_MODEL_TMPDIR}/${MINIGUI_MODEL}.${suffix}"
+  echo "Checking for: ${file_to_check}"
+  if [[ ! -f "${file_to_check}" ]]; then
+    gsutil cp gs://${MINIGUI_BUCKET_NAME}/${MINIGUI_GCS_DIR}/${MINIGUI_MODEL}.${suffix} $MINIGUI_MODEL_TMPDIR/
+  fi
+done
+
+# Assume models need to be converted if .converted.meta doesn't exist.
```
amj: The thing we can do is actually convert and backfill all the models in our bucket.

(we should keep this but end up removing it once we've done the backfill)

minigui/serve.py
```diff
@@ -43,6 +49,13 @@
          "\"cc\" for the C++ implementation.")
 
 
+parser.add_argument(
```
amj: why can't this just be the python on your PATH?

Kashomon: If you're running in a virtualenv, sure, but if you're installing python inside of Ubuntu, then you need python2 as python for google-cloud-sdk and python3 has to then remain python3. And so needs to be passed into container-land.

PR #193, opened 2018-05-01
Lots of minigui scripty-cleanup.

Kashomon:
- Exported minigui to minigo-pub and set ACLS to allUsers+R.
- Made it so you can pass in a model to minigui via an env variable.
- Some cleanup of cluster scripts
- Add a dummy minigui-pod and service yaml. Haven't actually tried them out yet.

You can now run minigui with:

```shell
git clone git@github.com:tensorflow/minigo.git
export MINIGUI_MODEL=000330-audacious
export MINIGUI_BOARD_SIZE=19 # default is 19, so not necessary here.
export MINIGUI_GCS_PATH=v5-19x19/models
minigo/cluster/minigui/run-rocal.sh
```

And it'll pull gcr.io/minigo-pub/minigui-py-cpu:0.16 and run minigui.


Kashomon:
Given this is just shell scripty things and given tight-timelines and timezones, I'm going to just merge this.

PR #208, opened 2018-05-04
Change C++ to handle virtual losses the same way as the Python code

tommadams:
This way they both play the exact same game with all randomization disabled.


brilee:
Oh, that's interesting - you managed to replicate the exact same game / search in your C++ rewrite...

Depending on how you did the C++ rewrite, that could be strong evidence that MCTS is bug-free.

tommadams:
I basically copied the Python code & tests, so I wouldn't assume about correctness based on my code :)

PR #209, opened 2018-05-08
fix wrapper script to correctly count games.  Fix output of duration …

amj:
…to always have the same unit in mcts_player.cc

### DIFF
--- a/cc/mcts_player.cc
+++ b/cc/mcts_player.cc
@@ -85,7 +85,8 @@ Coord MctsPlayer::SuggestMove(int num_readouts) {
     TreeSearch(options_.batch_size);
   }
   auto elapsed = absl::Now() - start;
-  std::cerr << "Seconds per 100 reads: " << elapsed * 100 / num_readouts
+  elapsed = elapsed * 100 / num_readouts;
+  std::cerr << "Seconds per 100 reads: " << absl::ToInt64Milliseconds(elapsed)
             << std::endl;
 
   if (ShouldResign()) {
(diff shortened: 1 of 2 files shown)


cc/mcts_player.cc
```diff
@@ -85,8 +85,9 @@ Coord MctsPlayer::SuggestMove(int num_readouts) {
     TreeSearch(options_.batch_size);
   }
   auto elapsed = absl::Now() - start;
-  std::cerr << "Seconds per 100 reads: " << elapsed * 100 / num_readouts
-            << std::endl;
+  elapsed = elapsed * 100 / num_readouts;
```
sethtroisi: this isn't measuring elapsed can you update the name to be `timing` or `time_per` or something

cluster/selfplay/cc_wrapper.sh
```diff
@@ -24,18 +24,19 @@ echo board_size: $BOARD_SIZE
...
+echo "gsutil ls 'gs://$BUCKET_NAME/data/selfplay/$BASENAME/*.zz'"
+GAMES=`gsutil ls "gs://$BUCKET_NAME/data/selfplay/$BASENAME/*.zz" | wc -l`
+
+gsutil cp $MODEL_FILE .
 
 mkdir -p data/selfplay
 mkdir -p sgf
 
-if [ $GAMES -lt 15000 ];
+if [ $GAMES -lt 25000 ];
```
sethtroisi: I'd suggest making this a constant at the top of the file and printing it on line 51

cc/mcts_player.cc
amj: it is the time elapsed per 100 readouts?

PR #212, opened 2018-05-08
Misc fixes from trainer box

amj:
Remove unused generation_num param from train (using incremental steps instead of max_steps)  
Fix typos in rl_runner.  
Copy golden chunk locally before training, and clean up.  
Fix freeze-graph to use gfile to accept gs:// paths

### DIFF
--- a/rl_loop.py
+++ b/rl_loop.py
@@ -91,12 +91,17 @@ def train(working_dir):
         time.sleep(1*60)
     print("Using Golden File:", training_file)
 
+    local_copy = os.path.join(working_dir, str(new_model_num) + '.tfrecord.zz')
+    print("Copying locally to ", local_copy)
+    gfile.Copy(training_file, local_copy, overwrite=True)
+
     save_file = os.path.join(fsdb.models_dir(), new_model_name)
     try:
-        main.train(working_dir, [training_file], save_file,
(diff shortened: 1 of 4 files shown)


rl_loop.py
```diff
@@ -91,12 +91,17 @@ def train(working_dir):
         time.sleep(1*60)
     print("Using Golden File:", training_file)
 
+    local_copy = os.path.join(working_dir, str(new_model_num) + '.tfrecord.zz')
+    print("Copying locally to ", local_copy)
+    gfile.Copy(training_file, local_copy, overwrite=True)
```
brilee: does this end up being a net speedup?

rl_loop.py
```diff
@@ -91,12 +91,17 @@ def train(working_dir):
...
+
     save_file = os.path.join(fsdb.models_dir(), new_model_name)
     try:
```
brilee: move this try up so that if the save_file line fails, we still clean up

rl_runner.py
```diff
@@ -30,9 +30,7 @@ def loop(working_dir='estimator_working_dir'):
     """Run train and validate as subprocesses."""
     flags = [
         '--working_dir',
```
brilee: did you mean to take this line out too?

PR #214, opened 2018-05-09
Replaces pygtp with our own simple GTP engine.

### LINKED ISSUE #171: refactor minigui extensions
I stumbled on some of the copy-pasted code in gtp_extensions and it's a bit tricky to work around with the new hyperparam work.

Can you at some point refactor this so that the search logic code isn't copy-pasted?

One suggestion for how this might work is for strategies.MCTSPlayerMixin.suggest_move can have callbacks registered, and by default those callbacks do nothing, but gtp_extensions can hook into those callbacks to retrieve debug stats.

### LINKED ISSUE #202: q_heatmap gtp command seems broken
At least in v5-running-head q_heatmap assumes board is 9x9 even when board_size is set and exported.

### LINKED ISSUE #204: "genmove" with no color results in crash
expected behavior not a crash

tommadams:
Fixes #204 
Fixes #202 
Fixes #171 

Apologies for the size of this PR, I ended up having to move a lot of code around :/

The new GTP engine works by allowing the client to register any arbitrary object with cmd_* methods that implement the GTP commands, which results in simpler easier code (no more mixins!)

The GTP engine uses a *little* bit of reflection magic to automatically convert the raw GTP command arguments into function arguments of the expected type for each command, which means that the command handlers don't need to mess around splitting & parsing strings.


cc/gtp_player.cc
```diff
@@ -103,9 +103,9 @@ GtpPlayer::Response GtpPlayer::DispatchCmd(
     return HandleEcho(cmd, args);
   } else if (cmd == "final_score") {
     return HandleFinalScore(cmd, args);
-  } else if (cmd == "gamestate" || cmd == "mg_gamestate") {
+  } else if (cmd == "gamestate" || cmd == "gamestate") {
```
amj: are these different in some exciting and subtle way that i cannot detect with my mortal eyes?

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+                pass
+            except:
+                print("Error saving sgf", file=sys.stderr, flush=True)
+        self._player.initialize_game(go.Position(komi=self._komi))
+
+    def cmd_komi(self, komi: float):
+        self._komi = komi
+        self._player.get_position().komi = komi
+
+    def cmd_play(self, arg0: str, arg1=None):
```
amj: did you mean to leave this as arg0/arg1?

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+        position = self._player.get_position()
+        if translate_gtp_color(color) != position.to_play:
+            position.flip_playerturn(mutate=True)
+
+
+class KgsCmdHandler(object):
+    def __init__(self, player):
+        self._player = player
+
+    def cmd_time_left(self):
```
amj: i think this is called with arguments -- its the servers way of informing the engine how much time is left.  There are extended time commands on CGOS that we may want to support for slightly better time management.

cc/gtp_player.cc
tommadams: no, no, I'm just stupid

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+            coords.to_kgs(coords.from_flat(key)),
+            node.__dict__.get(prop)[key])
+            for key in sort_order if node.child_N[key] > 0][: 20])
+
+
+class MiniguiCmdHandler(BasicCmdHandler):
+    def __init__(self, player):
+        super().__init__(player)
+
+        # Wrap the game's tree_search method, which allows us to update stream
```
amj: nit: sentence does not parse

gtp_engine.py
```diff
@@ -0,0 +1,197 @@
...
+#   https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html
+
+import inspect
+import re
+import sys
+import traceback
+
+
+# List of canonical error messages as required by 6.3 in the specification.
+_CANONICAL_ERRORS = {
```
amj: Should these be referenced in e.g the ValueError raised in `cmd_boardsize`?  might be DRY-er

gtp_engine.py
```diff
@@ -0,0 +1,197 @@
...
+            _print_error(msg_id, "unknown command")
+            return True
+
+        try:
+            handler = self.cmds[sanitized_cmd]
+            args = _convert_args(handler, args)
+            _print_success(msg_id, handler(*args))
+        except Exception as e:
+            traceback.print_exc(file=sys.stderr)
+            if cmd in _CANONICAL_ERRORS:
```
amj: alternatively, if you're going to catch the error and re-emit the canonical error code, you could make the errors elsewhere be more informative

.pylintrc
```diff
@@ -76,8 +77,10 @@ confidence=
 # --disable=W"
 
 disable=bare-except,
-        too-many-locals,
-        too-many-arguments
+	invalid-name,
```
brilee: nit: indentation

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+    if gtp_color.lower() in ["w", "white"]:
+        return go.WHITE
+    raise ValueError("invalid color {}".format(gtp_color))
+
+
+class BasicCmdHandler(object):
+    """GTP command handler for basic play commands."""
+
+    def __init__(self, player, courtesy_pass=False):
+        self._komi = 6.5
```
brilee: probably want to default to 7.5

gtp_wrapper.py
```diff
@@ -12,147 +12,32 @@
...
     name = "Minigo-" + os.path.basename(read_file)
-    gtp_engine = gtp_extensions.GTPDeluxe(instance, name=name)
-    return gtp_engine
+    version = "0.2"
+
+    engine = gtp_engine.Engine()
+    engine.add_cmd_handler(
+        gtp_engine.EngineCmdHandler(engine, name, version))
+
+    engine.add_cmd_handler(KgsCmdHandler(player))
```
brilee: Is it a good idea to register these handlers in the `if cgos_mode` blocks, so that we don't accidentally handle commands we didn't intend to handle in certain modes?

gtp_engine.py
```diff
@@ -0,0 +1,197 @@
...
+    msg = msg.split("#", 1)[0]
+    msg = msg.replace("\t", " ")
+    return msg
+
+
+def _parse(msg):
+    msg = _preprocess(msg).strip()
+    if not msg:
+        return None, None, None
+    parts = [x for x in msg.split(" ") if x]
```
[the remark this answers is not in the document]
brilee: i use this trick a lot, but I understand why you don't necessarily recommend it :D

gtp_engine.py
```diff
@@ -0,0 +1,197 @@
...
+        be invoked when the engine receives a genmove command.
+
+        Args:
+          handler_obj: the handler object to register.
+        """
+        for field in dir(handler_obj):
+            if field.startswith("cmd_"):
+                cmd = field[4:]
+                fn = getattr(handler_obj, field)
+                self.cmds[cmd] = fn
```
brilee: maybe print a warning if a cmd is defined twice? The override behavior seems to be done by later cmd handlers overwrite earlier ones.

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+class GoGuiCmdHandler(object):
+    """GTP extensions of 'analysis commands' for gogui."""
+
+    def __init__(self, player):
+        self._player = player
+
+    def cmd_gogui_analyze_commands(self):
+        return "\n".join(["var/Most Read Variation/nextplay",
```
brilee: are these parsed / interpreted somewhere?

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+            msg["lastMove"] = None
+        msg["n"] = position.n
+        if root.parent and root.parent.parent:
+            msg["q"] = root.parent.Q
+        else:
+            msg["q"] = 0
+        dbg("mg-gamestate:%s", json.dumps(msg, sort_keys=True))
+
+    def cmd_genmove(self, color=None):
+        start = datetime.now()
```
brilee: first time i've seen someone use this instead of time.time() - does this guard against some weird corner case?

gtp_cmd_handlers.py
amj: yeah, go gui adds the output of this command to its UI, giving you buttons or whatever

gtp_engine.py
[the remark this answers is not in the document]
amj: How is it a trick?  Is it doing something different?  'Cuz it just looks less legible to me

Also, the other way (`[x for x if x]`) is actually fewer characters...

gtp_cmd_handlers.py
tommadams: Yes, unfortunately.

If the two argument variant of `play` was `play move color` instead of `play color move`, I could just make `color` optional. But as it is, I don't see a way of sensibly naming the parameters. I guess I could have gone with `color_or_move` and `maybe_move` or something, but I don't think that's any more readable than just `arg0` and `arg1`.

gtp_cmd_handlers.py
tommadams: done.

gtp_cmd_handlers.py
tommadams: done.

gtp_engine.py
tommadams: I've updated the exceptions to contain more meaningful information

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+
+        root = self._player.get_root()
+        default_response = "Supported commands are 'winrate', 'nextplay', 'fortune', and 'help'."
+        if root is None or root.position.n == 0:
+            return "I'm not playing right now.  " + default_response
+
+        if 'winrate' in text.lower():
+            wr = (abs(root.Q) + 1.0) / 2.0
+            color = "Black" if root.Q > 0 else "White"
+            return "{:s} {:.2f}%".format(color, wr * 100.0)
```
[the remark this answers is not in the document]
tommadams: {:.2f} means the number will be formatted with two decimal places, not that it's two characters wide, e.g.:

>>> '{:.2f}%'.format(100)
'100.00%'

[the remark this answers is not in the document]
tommadams: done.

gtp_cmd_handlers.py
```diff
@@ -0,0 +1,340 @@
...
+            # 'play move' on its internal position objects, but we really
+            # want to advance the engine along with us rather than try to
+            # push in some finished Position object.
+            for idx, p in enumerate(sgf_wrapper.replay_sgf(contents)):
+                print("playing #", idx, p.next_move, file=sys.stderr)
+                self._player.play_move(p.next_move)
+                if movenum and idx == movenum:
+                    break
+        except:
+            raise
```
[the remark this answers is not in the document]
tommadams: done. fwiw, that was in the original code

gtp_engine.py
```diff
@@ -0,0 +1,197 @@
...
+    if len(parts) > 1 and parts[0].isdigit():
+        msg_id = parts[0]
+        parts = parts[1:]
+    else:
+        msg_id = None
+    return msg_id, parts[0], parts[1:]
+
+
+def _print_msg(result, msg_id, msg):
+    if msg_id:
```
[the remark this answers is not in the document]
tommadams: done.

.pylintrc
tommadams: thanks for catching that. vim had inserted tabs for some reason

gtp_wrapper.py
tommadams: done.

gtp_engine.py
tommadams: done.
good suggestion, thanks

gtp_cmd_handlers.py
tommadams: no particular reason, just habit I guess
switched to `time.time()`, since it's a bit more readable

gtp_cmd_handlers.py
tommadams: Agreed, but the 6.5 comes from the original code so I'm going to leave it unchanged in this PR, since in theory it shouldn't change any behaviour.

PR #219, opened 2018-05-10
update cc_wrapper.sh to set CUDA_VISIBLE_DEVICES

amj:
I'm not super proud of this but it does work.  See attached picture showing fleetwide speed improvements on k80s.

image


amj:
this is a little brittle and will probably break whenever we update nvidia stuff.  i'll try and do something more reliable with nvml tomorrow.

cluster/selfplay/cc-Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04
 
-RUN apt-get update
-RUN apt-get install -y --no-install-recommends \
+RUN apt-get update && apt-get install -y --no-install-recommends \
         build-essential \
+        bc \
```
Kashomon: what's this?

cluster/selfplay/cc_wrapper.sh
```diff
@@ -33,6 +33,12 @@ GAMES=`gsutil ls "gs://$BUCKET_NAME/data/selfplay/$BASENAME/*.zz" | wc -l`
 
 gsutil cp $MODEL_FILE .
 
+GPU_TO_USE=`nvidia-smi | grep -o [0-9]*% | cut -d% -f1 | paste -s -d">" - | bc`
+
+export CUDA_VISIBLE_DEVICES=$GPU_TO_USE
```
Kashomon: but why...? Is this some weirdness with the new NVidia Image that we're only nows seeing?

cluster/selfplay/cc-Dockerfile
amj: basic calculator :)

cluster/selfplay/cc_wrapper.sh
amj: Not sure why now.  In short, the cc-player can "see" both GPUs, and it doesn't automatically choose which GPU to put its ops on -- it always uses gpu0.  Setting the visible devices to either 0/1 makes that the only visible GPU to the binary.

the GKE docs suggest that a pod should only see the GPUs it requested in its spec, which is clearly not happening.  So i'd like to followup with vish about this.

amj:
Updated this to actually do this in a not horrible way if anyone is curious

cluster/selfplay/cc-Dockerfile
Kashomon: do you still need it?

amj: no, but i didn't want to rebuild tensorflow to get rid of it XD.

PR #224, opened 2018-05-11
Replace dual_net hyperparams with flags

brilee:


dual_net.py
```diff
@@ -305,30 +293,31 @@ def export_model(working_dir, model_path):
...
     if steps is None:
-        steps = EXAMPLES_PER_GENERATION // TRAIN_BATCH_SIZE
+        steps = EXAMPLES_PER_GENERATION // FLAGS.train_batch_size
     estimator.train(input_fn, hooks=[
                         update_ratio_hook, step_counter_hook], steps=steps)
 
 
-def validate(working_dir, tf_records, checkpoint_name=None, **hparams):
-    estimator = get_estimator(working_dir, **hparams)
+def validate(working_dir, tf_records, checkpoint_name=None, validate_name=None):
```
brilee: apparently the **kwargs was swallowing the fact that we were passing in invalid kwarg names.

amj: which invalid kwarg names were we passing?

dual_net.py
```diff
@@ -279,7 +268,7 @@ def bootstrap(working_dir, **hparams):
     sess = tf.Session(graph=tf.Graph())
     with sess.graph.as_default():
         features, labels = get_inference_input()
-        model_fn(features, labels, tf.estimator.ModeKeys.PREDICT, hparams)
+        model_fn(features, labels, tf.estimator.ModeKeys.PREDICT)
```
amj: so we're sure nothing was actually being set by the hparams here?

local_rl_loop.py
```diff
@@ -40,11 +40,10 @@ def rl_loop():
     the reinforcement learning.
     """
     # TODO(brilee): move these all into appropriate local_flags file.
```
amj: keep the monkeypatch comment explaining the small net

tests/test_flags
```diff
@@ -1,2 +1,5 @@
 # One line per flag
-# For example --someflag=value
\ No newline at end of file
+# For example --someflag=value
```
amj: is this actually operational?  will this run if i run `python -m unittest` etc ?

brilee:
Hm. If the network hparams changed, then it will just fail to run, which is safe. The change to worry about is if we accidentally change a shuffler queue size param or something of that nature.

dual_net.py
brilee: Well, they were but now it's being set by flags.

dual_net.py
brilee: we were passing name=validate_name in main.py

dual_net.py
```diff
@@ -305,30 +293,31 @@ def export_model(working_dir, model_path):
...
-def validate(working_dir, tf_records, checkpoint_name=None, **hparams):
-    estimator = get_estimator(working_dir, **hparams)
+def validate(working_dir, tf_records, checkpoint_name=None, validate_name=None):
+    estimator = get_estimator(working_dir)
     if checkpoint_name is None:
         checkpoint_name = estimator.latest_checkpoint()
 
     def input_fn():
-        return preprocessing.get_input_tensors(VALIDATE_BATCH_SIZE, tf_records,
+        return preprocessing.get_input_tensors(FLAGS.train_batch_size, tf_records,
```
brilee: This is an intentional change but I think it should not affect things. Validation will now execute at the same batch size as training; since validation doesn't cause variable updates it should be irrelevant. It might actually be slower because it seems like we're doing steps=500, so we'd do 8x more validation.

dual_net.py
```diff
@@ -305,30 +293,31 @@ def export_model(working_dir, model_path):
...
-    estimator = get_estimator(working_dir, **hparams)
+def train(working_dir, tf_records, steps=None):
+    estimator = get_estimator(working_dir)
 
     def input_fn():
-        return preprocessing.get_input_tensors(TRAIN_BATCH_SIZE, tf_records,
-                                               filter_amount=1.0)
+        return preprocessing.get_input_tensors(
+            FLAGS.train_batch_size, tf_records, filter_amount=1.0,
+            shuffle_buffer_size=FLAGS.shuffle_buffer_size)
```
brilee: This is an intentional change from 2,000,000 to 10,000. Given that you are now using the example buffer it should not affect training. But if you'd be more comfortable keeping it at 2M, I can do that.

tests/test_flags
brilee: It only gets pulled in if you use `./test.sh`

dual_net.py
brilee: I'll change to 2M since this is merely intended to be a refactor, not change in functionality. Please feel free to drop this to a faster-filling number as you see fit; it should not affect the shuffling characteristics.

dual_net.py
amj: i'm fine with it, but validation takes a couple minutes right now.  i don't want it to be > 5min.

dual_net.py
amj: ok, thanks!

PR #228, opened 2018-05-14
add list_commands support.  Support parsing KGS coords that are sent …

amj:
…lowercase.  Print "gtp ready" string to stderr


cc/gtp_player.cc
```diff
@@ -98,6 +98,8 @@ GtpPlayer::Response GtpPlayer::DispatchCmd(
     absl::string_view cmd, const std::vector<absl::string_view>& args) {
   if (cmd == "boardsize") {
     return HandleBoardsize(cmd, args);
+  } else if (cmd == "list_commands") {
+    return HandleListCommands(cmd, args);
```
tommadams: nit: please sort the Handle* commands in sorted order

cc/gtp_player.h
```diff
@@ -73,6 +73,9 @@ class GtpPlayer : public MctsPlayer {
   Response DispatchCmd(absl::string_view cmd,
                        const std::vector<absl::string_view>& args);
 
+  Response HandleListCommands(absl::string_view cmd,
```
tommadams: nit: ordering

cc/gtp_player.cc
```diff
@@ -122,6 +124,21 @@ GtpPlayer::Response GtpPlayer::DispatchCmd(
   return Response::Error("unknown command");
 }
 
+GtpPlayer::Response GtpPlayer::HandleListCommands(
+    absl::string_view cmd, const std::vector<absl::string_view>& args) {
+  auto response = CheckArgsExact(cmd, 0, args);
+  if (!response.ok) {
+    return response;
+  }
+  absl::string_view commands[] = {
```
tommadams: can you add a TODO to put the command handlers in a map<string, std::function> so we can avoid having to list the commands separately here?

PR #229, opened 2018-05-14
updates from training box: validate on last 30.  adjust freq of updat…

amj:
…e_ratio_hook.  Adjust games/generation constant.  Add notes.  Default shuffle buffer size is 20k


dual_net.py
```diff
@@ -298,17 +297,18 @@ def input_fn():
     if steps is None:
         steps = EXAMPLES_PER_GENERATION // FLAGS.train_batch_size
     estimator.train(input_fn, hooks=[
-                        update_ratio_hook, step_counter_hook], steps=steps)
+        update_ratio_hook, step_counter_hook], steps=steps)
 
 
 def validate(working_dir, tf_records, checkpoint_name=None, validate_name=None):
     estimator = get_estimator(working_dir)
+    validate_name = validate_name or "selfplay"
```
sethtroisi: it's weird that this has different style than the next line.

can you unify them to
A = A or default
or
if B is none: B = default

notes.txt
```diff
@@ -24,3 +24,20 @@ buffer down to 100k
 
 1. v360ish -- entered experimental mode; freely adjusted learning rate up and
 down, adjusted batch size, etc, to see if valnet divergence could be fixed.
+
+
+== v7
+
+Started training with model 174.  Moved minimum games before starting training
```
sethtroisi: Rolled back to v5 model 173 and continue from there with first v7 model being 174

notes.txt
```diff
@@ -24,3 +24,20 @@ buffer down to 100k
...
+from 5k up to 10k.
+
+1. for v215, adjusted validation set to be last 30 models instead of last 50.
```
sethtroisi: sudden use of v to refer to model

notes.txt
```diff
@@ -24,3 +24,20 @@ buffer down to 100k
 
 1. v360ish -- entered experimental mode; freely adjusted learning rate up and
 down, adjusted batch size, etc, to see if valnet divergence could be fixed.
+
+
+== v7
+
+Started training with model 174.  Moved minimum games before starting training
+from 5k up to 10k.
```
sethtroisi: seems like it was 500

rl_loop.py
```diff
@@ -29,10 +29,10 @@
 import shipname
 
 # How many games before the selfplay workers will stop trying to play more.
-MAX_GAMES_PER_GENERATION = 10000
+MAX_GAMES_PER_GENERATION = 20000
```
sethtroisi: we have several models with more than this so we may have self-play pauses if you choose this number

rl_loop.py
```diff
@@ -122,7 +122,7 @@ def validate(working_dir, model_num=None, validate_name=None):
     # Run on the most recent 50 generations,
```
sethtroisi: s/50/30

notes.txt
amj: Done

dual_net.py

notes.txt
amj: It wasn't :P

rl_loop.py
amj: Ack.  We're not using py for selfplay anymore so /shruggie

rl_loop.py
amj: Done.

notes.txt
amj: Done.

PR #232, opened 2018-05-15
Log random seed & MctsPlayer options to stderr on startup.

tommadams:
Switch some asserts to MG_CHECK.

Example output:
MctsPlayer options: inject_noise:1 soft_pick:1 random_symmetry:1 resign_threshold:-0.95 batch_size:8 komi:7.5 random_seed:0
Random seed: 1526409134701039


amj:
is there a different tf random seed?

PR #240, opened 2018-05-20
Added more tensorboard metrics

sethtroisi:
Added accuracy, accuracy_top_3, policy_probability when top_1 matches, and value_confidence
I find these metrics to be useful during supervised training, not sure how much this will contribute to overly large tf event logs.


sethtroisi:
Example screenshot screenshot from 2018-05-20 10-43-04

dual_net.py
```diff
@@ -306,8 +315,10 @@ def validate(working_dir, tf_records, checkpoint_name=None, validate_name=None):
     checkpoint_name = checkpoint_name or estimator.latest_checkpoint()
 
     def input_fn():
-        return preprocessing.get_input_tensors(FLAGS.train_batch_size, tf_records,
-                                               shuffle_buffer_size=20000, filter_amount=0.05)
+        return preprocessing.get_input_tensors(
+            FLAGS.train_batch_size, tf_records, filter_amount=0.05,
```
sethtroisi: This should probably be updated to filter_amount=1.0 to match train() (see https://github.com/tensorflow/minigo/blob/master/dual_net.py#L292)

preprocessing.py
```diff
@@ -169,6 +169,7 @@ def get_input_tensors(batch_size, tf_records, num_repeats=None,
 
     Returns a dict of tensors (see return value of batch_parse_tf_example)
     '''
+    print ("Reading tf_records from {} inputs".format(len(tf_records)))
```
sethtroisi: useful sanity check when training not with chunks to make sure your holdout games have been excluded.

dual_net.py
```diff
@@ -210,10 +210,17 @@ def my_res_layer(inputs):
...
+    policy_top1 = tf.argmax(policy_output, 1)
+    policy_top3 = tf.to_int64(tf.nn.top_k(policy_output, 3).indices)
+    top3_match = tf.to_float(tf.reduce_any(tf.equal(tf.expand_dims(policy_label, [1]), policy_top3), 1))
+
     metric_ops = {
-        'accuracy': tf.metrics.accuracy(labels=labels['pi_tensor'],
-                                        predictions=policy_output,
-                                        name='accuracy_op'),
+        'accuracy_top_1': tf.metrics.accuracy(labels=policy_label, predictions=policy_top1),
+        'accuracy_top_3': tf.metrics.mean(top3_match),
```
brilee: See here for the official way to do top_N statistics

https://github.com/tensorflow/tpu/blob/master/models/official/resnet/resnet_main.py#L346

dual_net.py
```diff
@@ -210,10 +210,17 @@ def my_res_layer(inputs):
...
+        'policy_prop_of_move': tf.metrics.mean(tf.reduce_sum(labels['pi_tensor'] * policy_output, 1)),
```
brilee: This seems to be cosine similarity of the two distributions. Technically, the distribution similarity is already being captured by the policy loss function.

dual_net.py
```diff
@@ -238,7 +255,11 @@ def my_res_layer(inputs):
 
 
 def get_estimator(working_dir):
-    return tf.estimator.Estimator(model_fn, model_dir=working_dir)
+    run_config = tf.estimator.RunConfig(save_summary_steps=500)
```
sethtroisi: This reduces the logging by 5x (but update ratios still represent 90% of the log)

sethtroisi:
screenshot from 2018-05-22 01-48-07

dual_net.py
```diff
@@ -210,16 +210,33 @@ def my_res_layer(inputs):
...
                 combined_cost, global_step=global_step)
 
+    policy_target_top_1 = tf.argmax(labels['pi_tensor'], axis=1)
+    policy_output_top_1 = tf.argmax(policy_output, axis=1)
+
+    policy_output_in_top3 = tf.to_float(
+        tf.nn.in_top_k(policy_output, policy_target_top_1, k=3))
+
+    policy_top_1_confidence = tf.reduce_max(policy_output, axis=1)
+    policy_target_top_1_confidence = tf.boolean_mask(
```
brilee: tf.reduce_max(policy_output, axis=1)

sethtroisi: discussed offline.

policy_top_1 is how much weight is going to the top policy output (human interpretations is: how confident is policy) policy_target_top_1 measures how much weight is going to the top TARGET policy output (i.e. human interpretations is: how much did we look at the move that was ultimately chosen)

PR #245, opened 2018-05-22
Adds 'eval' mode to cc/main, to play a game between different networks 

amj:
- adds eval mode and options helper
- Move name_ from gtp_player up to mcts_player.
- Add a path helper for extracting the basename prefix.  
- Extend WriteSgf to take two players


amj:
the likelihood that i'm misusing strings here is high.  i'm going to run with asan and verify i'm not doing anything too silly

cc/main.cc
```diff
@@ -204,6 +219,52 @@ void SelfPlay() {
...
+  options->random_symmetry = true;
+  options->resign_threshold = FLAGS_resign_threshold;
+  options->batch_size = FLAGS_batch_size;
+  options->komi = FLAGS_komi;
+  options->random_seed = FLAGS_seed;
+}
+
+void Eval() {
+  MctsPlayer::Options options;
+  SetMctsPlayerOptionsForEval(&options);
```
tommadams: I'd be tempted to just call ParseMctsPlayerOptionsFromFlags here, then override inject_noise, soft_pick and random_symmetry explicitly.

amj:
The reason not to is because you want the selfplay job to do the tree re-use; with two separate players you dilute your reads across two trees.

tommadams:
Fair enough :)

cc/file/path.h
```diff
@@ -44,6 +44,15 @@ inline absl::string_view Basename(absl::string_view path) {
   return SplitPath(path).second;
 }
 
+inline absl::string_view BasePrefix(absl::string_view path) {
```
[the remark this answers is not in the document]
amj: hey look at that term of art.  Done.

cc/main.cc
amj: Done.

Issue #242, opened 2018-05-21
cc-Dockerfile image breaks if built in a dir with existing tensorflow .so's 


amj:
steps to repro:
1. run cc/configure_tensorflow.sh
2. build cc-selfplay docker image from cluster/selfplay/cc-Dockerfile
3. cc/main will not run

PR #247, opened 2018-05-25
Cc-eval docker refactor

### LINKED ISSUE #243: extract cc-dockerfile to common base image.

amj:
- A bit of redundancy eliminated by refactoring our cc images into a common base docker image.  
- A bit of redundancy created in the wrapper scripts that call our `main` binaries

Fixes #243 and #242


cluster/base/Makefile
```diff
@@ -0,0 +1,12 @@
+base-image:
+		mkdir -p staging/cc
+		cp ../../WORKSPACE staging/
+		cp -r -p ../../cc/ staging/
+		rm -rfd staging/cc/tensorflow
```
amj: this bit fixes #242 by preventing any host-compiled .so's from getting copied into the staging directory.

cluster/evaluator/cc-evaluator.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  parallelism: 10
+  completions: 10
+  template:
+    metadata:
+      name: gpu-evaluator
+    spec:
+      containers:
+      - name: evaluator
+        image: gcr.io/$PROJECT/minigo-cc-evaluator:$VERSION_TAG
```
amj: i didn't extract the 'minigo-cc-evaluator' out to an env var because i'm not sure why we parameterized the container names to environment variables in the first place

cluster/base/Makefile
```diff
@@ -0,0 +1,12 @@
...
+	  docker build -t "gcr.io/tensor-go/cc-base:latest" .
```
Kashomon: latest? that seems pretty dangerous

amj: Fixed, thanks :)

PR #251, opened 2018-05-31
Docker images for CGOS

amj:
we can build off this to make some images for fuego/pachi/etc

### DIFF
--- a/cc/gtp_player.cc
+++ b/cc/gtp_player.cc
@@ -42,6 +42,7 @@ GtpPlayer::GtpPlayer(std::unique_ptr<DualNet> network, const Options& options)
   RegisterCmd("gamestate", &GtpPlayer::HandleGamestate);
   RegisterCmd("genmove", &GtpPlayer::HandleGenmove);
   RegisterCmd("info", &GtpPlayer::HandleInfo);
+  RegisterCmd("komi", &GtpPlayer::HandleKomiCommand);
   RegisterCmd("known_command", &GtpPlayer::HandleKnownCommand);
   RegisterCmd("list_commands", &GtpPlayer::HandleListCommands);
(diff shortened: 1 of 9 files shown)


cc/gtp_player.cc
```diff
@@ -312,6 +313,20 @@ GtpPlayer::Response GtpPlayer::HandleInfo(
   return Response::Ok(oss.str());
 }
 
+GtpPlayer::Response GtpPlayer::HandleKomiCommand(
```
tommadams: ordering nit: Komi comes after Known

cc/gtp_player.h
```diff
@@ -110,6 +110,9 @@ class GtpPlayer : public MctsPlayer {
   Response HandleInfo(absl::string_view cmd,
                       const std::vector<absl::string_view>& args);
 
+  Response HandleKomiCommand(absl::string_view cmd,
```
tommadams: same here

cc/gtp_player.cc
amj: err, right...

cjwagner:
/test pull-tf-minigo-cc
(should report a status now)

amj:
whoa!  @cjwagner that looks like a real failure -- anyway to see the test failure in the stuff at the "details" link?

cjwagner:
Yeah, bazel doesn't output logs for failed tests by default 😞 
I made https://github.com/kubernetes/test-infra/pull/8216 to fix this.

cjwagner:
The test passed this time so it must be flaky.
I'll run it again to see if we can get a failure now that the logs will be useful.
/test pull-tf-minigo-cc

amj:
woohoo!  looks like tom tracked down the flaky test in #255 .  Merging!

PR #255, opened 2018-06-04
Fix flaky position_test

tommadams:
Do I really have more bugs in the position code? No, it's the tests that are wrong.


amj:
haha, well that's a relief.

amj:
LGTM ok2merge

PR #256, opened 2018-06-04
Basic support for running inference on a separate process

tommadams:
Apologies for this size of this thing :/

Running inference on a separate process does not yet support applying random symmetries to the input features, that will require some non-trivial refactoring so I'll add it in a followup PR.

High level changes:

- I've updated the required version of TensorFlow to one that has RPC & proto op support.
- inference_worker.py: wraps the minigo model in a TensorFlow loop that runs indefinitely. The top of the loop requests input features via an RPC op (GetFeatures). The bottom of the loop sends back the inference output via a send RPC op (PutOutputs)
- cc/dual_net/inference_server.cc: a gRPC server. The MCTS code pushes requests onto the server's request queue. The server sends those requests to the inference worker in response to the GetFeatures RPCs and populates the outputs in response to the PutOutputs requests).
- dual_net.py: I've moved the code that generates the inference-only parts of the minigo model out of model_fn and into a new model_inference_fn, so that it can be reused by inference_worker.py.
- proto/gen_inference_service_py_srcs.sh: a script that builds the Python proto & gRPC runtime. This is only needed because we don't (yet?) build our Python code with Bazel. I'm checking in the generated code for now because we haven't required extra build steps for Python before.


amj:
Cool!  If i wanted to pull this in and try it, what sort of env prep work do i need?  Particular versions of TF?  (i notice the unittests can't run due to imports in tf.contrib)

tommadams:
You'll need TF 1.8 so:
pip3 install "tensorflow-gpu>=1.8,<1.9"

Even though I've updated cc/configure_tensorflow.sh in this PR to grab v1.8, you shouldn't need to rerun it because I use a Python harness to run the model with RPC ops.

tommadams:
Oh, and to run it...

```
python inference_worker.py  --model=$MODEL_PATH
```

It should print out "Waiting for server" every second.

Then run `cc:main` with `--model=remote` instead of a model .pb path.

FWIW, it doesn't actually matter which order you launch the two processes, they should wait for each other.

cc/mcts_player.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "cc/mcts_player.h"
 
+#include <algorithm>
```
amj: eh?

cc/thread_safe_queue.h
```diff
@@ -0,0 +1,86 @@
...
+    absl::MutexLock lock(&m_);
+    queue_.push(x);
+  }
+
+  void Push(T&& x) {
+    absl::MutexLock lock(&m_);
+    queue_.push(std::move(x));
+  }
+
+  bool TryPop(T* x) {
```
amj: is it possible to replace this with a call to PopWithTimeout with a timeout of $REALLY_LONG_DURATION ?

dual_net.py
```diff
@@ -123,30 +123,21 @@ def get_inference_input():
...
-        predictions: dict of tensors
-            'policy': [BATCH_SIZE, go.N * go.N + 1]
-            'value': [BATCH_SIZE]
-        loss: a single value tensor
-        train_op: train op
-        eval_metric_ops
-    return dict of tensors
-        logits: [BATCH_SIZE, go.N * go.N + 1]
-    '''
+        features: input features tensor.
```
amj: keep the shape comments OR move the def of `model_fn` above `model_inference_fn` in the file?

inference_worker.py
```diff
@@ -0,0 +1,210 @@
...
+            descriptor_source=config.descriptor_path,
+            name="decode_raw_features")
+
+        # Reshape flat features.
+        features = tf.reshape(
+            flat_features, [-1, go.N, go.N, features_lib.NEW_FEATURES_PLANES],
+            name="unflatten_features")
+
+        # Run inference.
+        policy_output, value_output, _ = dual_net.model_inference_fn(
```
amj: if i understand this right, this means that the model checkpoints loaded now need to match the hyperparams defined here -- i.e., we've lost the ability to load arbitrary depth/width nets that had been frozen as .pbs, correct?

proto/inference_service_pb2_grpc.py
```diff
@@ -0,0 +1,84 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
amj: should these generated files be in the .gitignore?  Should they be checked in?  I don't know :)

cc/mcts_player.cc
amj: actually, now that i'm here i notice i didn't see anything that takes care of ignoring the padded-out features

inference_worker.py
tommadams: Sadly yes, if you want to run with the InferenceServer. At least until I figure out how we can build C++ TensorFlow with the RPC & proto ops linked in.

dual_net.py
tommadams: moved `model_inference_fn` down

cc/thread_safe_queue.h
tommadams: No, TryPop returns immediately if the queue is empty.

cc/mcts_player.cc
tommadams: algorithm is required for std::max.

the padded-out features are ignored because InferenceServer::PutOutputs only iterates over the items in the *requested* batch, not the actual batch sent to the model. See: https://github.com/tensorflow/minigo/pull/256/files#diff-5ff231d14f937b5c55b58e7a87281ee1R149

cc/dual_net/inference_server.cc
```diff
@@ -0,0 +1,247 @@
...
+InferenceServer::~InferenceServer() {
+  // Passing gpr_inf_past to Shutdown makes it shutdown immediately.
+  server_->Shutdown(gpr_inf_past(GPR_CLOCK_REALTIME));
+  thread_.join();
+}
+
+std::unique_ptr<DualNet> InferenceServer::NewDualNet() {
+  return absl::make_unique<InferenceClient>(
+      [this](const RemoteInference& inference) {
+        request_queue_->Push(inference);
```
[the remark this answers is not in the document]
tommadams: oops, that was a rough edge left over from my first attempt at writing the inference server. cleaned up now

cc/dual_net/dual_net.h
```diff
@@ -28,6 +28,8 @@
 
 namespace minigo {
 
+// TODO(tommadams): Figure out how best to handle random symmetries.
```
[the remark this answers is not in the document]
tommadams: done.

cc/configure_tensorflow.sh
```diff
@@ -5,17 +5,25 @@ set -e
...
+tmp_pkg_dir="/tmp/tensorflow_pkg"
 rm -rfd ${tmp_dir}
+rm -rfd ${tmp_pkg_dir}
 mkdir -p ${tmp_dir}
-version_tag="v1.6.0-rc0"
+
+# The TensorFlow 1.8.0 release doesn't compile with gcc6+, so checkout at the
+# commit that fixed the build issue.
+# See https://github.com/tensorflow/tensorflow/issues/18402 for more details.
+# TODO(tommadams): switch to v.18.1 when that's released.
```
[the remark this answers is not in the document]
tommadams: done.

README.md
```diff
@@ -88,11 +88,11 @@ pip3 install -r requirements.txt
 
 Then, you'll need to choose to install the GPU or CPU tensorflow requirements:
 
-- GPU: `pip3 install "tensorflow-gpu>=1.7,<1.8"`.
+- GPU: `pip3 install "tensorflow-gpu>=1.8,<1.9"`.
   - *Note*: You must install [CUDA
     9.0].(https://developer.nvidia.com/cuda-90-download-archive) for Tensorflow
```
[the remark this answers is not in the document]
tommadams: yep, I'm still using CUDA 9.0

cc/mcts_player.cc
amj: nice, got it

cc/thread_safe_queue.h
amj: oh i see, right.

PR #258, opened 2018-06-06
Tpu experiments

brilee:
Code is in working state (both --use_tpu=True/False work).

Invocation I used while testing on TPU-enabled VM:
export STORAGE_BUCKET=gs://....
export DATA_DIR=$STORAGE_BUCKET/data/golden_chunks
export TPU_NAME=....
export BOARD_SIZE=19
python3 dual_net.py train $STORAGE_BUCKET/models/initial_run $DATA_DIR/200.tfrecord.zz $DATA_DIR/201.tfrecord.zz --steps=1000 --use_tpu=True --tpu_name=$TPU_NAME

TBD: throughput for different numbers of parallel files read. (We currently just have 1 golden chunk file per gen which may or may not limit throughput.)
TBD: implement random rotations as tf ops. (@sethtroisi your original rotation impl would have been better here... my bad on having you use py_func...)
TBD: whether this impl is bug free


dual_net.py
```diff
@@ -387,3 +469,12 @@ def after_run(self, run_context, run_values):
             self.summary_writer.add_summary(
                 weight_update_summaries, global_step)
             self.before_weights = None
+
+parser = argparse.ArgumentParser()
+argh.add_commands(parser, [train])
```
amj: can you use the single-form here?  i think it's argh.dispatch_command

preprocessing.py
```diff
@@ -184,6 +191,20 @@ def get_input_tensors(batch_size, tf_records, num_repeats=None,
     return dataset.make_one_shot_iterator().get_next()
 
 
+def get_tpu_input_tensors(batch_size, tf_records):
+    dataset = read_tf_records(batch_size, tf_records, shuffle_records=True,
+        shuffle_buffer_size=1024, num_repeats=None)
+    # TODO take a bfloat16 param and cast stuff to bfloat16
```
amj: stuff?

main.py
```diff
@@ -89,7 +89,7 @@ def train(
         model_save_path: 'Where to export the completed generation.'):
     print("Training on:", tf_records[0], "to", tf_records[-1])
     with utils.logged_timer("Training"):
-        dual_net.train(working_dir, tf_records)
+        dual_net.train(working_dir, *tf_records)
```
amj: you'll need to update rl_loop for this as well.

Does local_rl_loop run?  that's no longer in tests.sh or run_tests.py after the flags work...

preprocessing.py
```diff
@@ -123,12 +124,18 @@ def read_tf_records(batch_size, tf_records, num_repeats=1,
...
     # and the examples being read from the files.
-    dataset = record_list.interleave(lambda x:
-                                     tf.data.TFRecordDataset(
-                                         x, compression_type='ZLIB'),
-                                     cycle_length=64, block_length=16)
-    dataset = dataset.filter(lambda x: tf.less(
-        tf.random_uniform([1]), filter_amount)[0])
+    if sloppy_interleave:
+        dataset = record_list.apply(tf.contrib.data.parallel_interleave(
+            functools.partial(tf.data.TFRecordDataset, buffer_size=8*1024*1024, compression_type='ZLIB'),
```
amj: how does this perform when using a golden chunk?

dual_net.py
```diff
@@ -52,7 +59,38 @@
...
+    'The Cloud TPU to use for training. This should be either the name used'
+    'when creating the Cloud TPU, or a grpc://ip.address.of.tpu:8470 url.')
+
+flags.register_multi_flags_validator(
+    ['use_tpu', 'tpu_name'],
+    lambda flags: bool(flags['use_tpu']) == bool(flags['tpu_name']),
+    'If use_tpu is set, tpu_name must also be set.')
+
+flags.DEFINE_integer(
+    'iterations_per_loop', 100,
```
amj: does this mean metrics every 100 * BATCH_SIZE positions?  Should train_batch_size change with use_tpu?

dual_net.py
```diff
@@ -52,7 +59,38 @@
...
+    help=('Number of steps to run on TPU before outfeeding metrics to the CPU.'
+          ' If the number of iterations in the loop would exceed the number of'
+          ' train steps, the loop will exit before reaching'
+          ' --iterations_per_loop. The larger this value is, the higher the'
+          ' utilization on the TPU.'))
+
+flags.DEFINE_integer(
+    'num_cores', default=8,
```
amj: num_tpu_cores?

dual_net.py
```diff
@@ -52,7 +59,38 @@
...
+    help=('Number of TPU cores. For a single TPU device, this is 8 because each'
+          ' TPU has 4 chips each with 2 cores.'))
+
+flags.DEFINE_enum(
+    'precision', 'float32', ['bfloat16', 'float32'],
```
amj: so then what's this for?  Would we run GPU with bfloat16?  what happens if i set float32 with usetpu = true?

dual_net.py
```diff
@@ -198,62 +237,83 @@ def my_res_layer(inputs):
...
-            'policy_output': policy_output,
-            'value_output': value_output,
-        },
-        loss=combined_cost,
-        train_op=train_op,
-        eval_metric_ops=metric_ops,
-    )
+        train_op = optimizer.minimize(combined_cost, global_step=global_step)
+
+    def host_call_fn(policy_output, value_output, pi_tensor, policy_cost,
```
amj: a comment here might be nice.

dual_net.py
```diff
@@ -307,22 +367,44 @@ def export_model(working_dir, model_path):
...
-    step_counter_hook = EchoStepCounterHook(output_dir=working_dir)
+def train(working_dir, *tf_records, steps=None):
+    if FLAGS.use_tpu:
+        tpu_cluster_resolver = tf.contrib.cluster_resolver.TPUClusterResolver(
+            FLAGS.tpu_name, zone=None, project=None)
+
+        config = tpu_config.RunConfig(
+            cluster=tpu_cluster_resolver,
+            model_dir=working_dir,
+            save_checkpoints_steps=max(600, FLAGS.iterations_per_loop),
```
amj: should this be a new flag?

dual_net.py
amj: (maybe just update the comment in train_batch_size's usage instructions to suggest a better default value for tpus)

dual_net.py
```diff
@@ -307,22 +367,44 @@ def export_model(working_dir, model_path):
...
+
+        estimator = tpu_estimator.TPUEstimator(
+            use_tpu=FLAGS.use_tpu,
+            model_fn=model_fn,
+            config=config,
+            train_batch_size=FLAGS.train_batch_size,
+            eval_batch_size=FLAGS.train_batch_size)
+        def input_fn(params):
+            return preprocessing.get_tpu_input_tensors(params['batch_size'], tf_records)
+        hooks = []
```
amj: no step counter output?

dual_net.py
brilee: I want to leave as multi-dispatch so that I can move all the main.py dual_net-specific stuff back into dual_net.py. Do you think that's a good idea?

main.py
brilee: rl_loop should be fine for now.

brilee: local_rl_loop still runs fine, there's a lot of deferred flag-fixup work... :(

preprocessing.py
brilee: That's TBD. The tpu resnet demo has like 64 parallel reads going on in order to satisfy the TPU's demands. OTOH, our stuff compresses very well, so 1 parallel read with the same bandwidth as each of the 64 parallel reads might end up decompressing to enough MB/s of data to satisfy the cloud TPU. We'll have to see

dual_net.py
brilee: fixed

dual_net.py
brilee: fixed

dual_net.py
brilee: gpu would probably fail unless it had bfloat16 support. TPU supports both float32 and bfloat16 mode.

preprocessing.py
amj: we've also got a PR out to provide ZLIB compression options -- should speed up the throughput substantially.

dual_net.py
brilee: added.

dual_net.py
brilee: Yeah, it broke tpu stuff, will leave a todo to reintegrate.

dual_net.py
amj: which stuff is that?
my mental model is:
main.py: 
basic operations on models.  setup for training (bootstrap, train, validate) or setup to run inference (gtp/selfplay/eval), push models around (freeze, update-global-step).  All the entrypoints you'd want to use in scripts, basically.
rl_loop.py
main.py ops with the boilerplate filled in around cloud-based paths, and codifying the idea of a sequential series of models.

so running 'dual_net.py' seems like a new category of stuff -- how are you thinking about this category?  Is there a different taxonomy that makes more sense?

main.py
amj: rl_loop calls main.train, and the *expansion shouldn't work as it looks right now, IIUC

PR #261, opened 2018-06-12
change -march=ivybridge to native

sethtroisi:
I'm not sure if this is the right way to pass COPT around.

Alternatively I can run 's/ivybridge/native/'


amj:
I think this needs to be a CC_OPTS environment variable, and the variable needs to be set in the Dockerfiles for cc-base/cc-player/cc-evaluator.  The packages we build for our clusters need to be --ivybridge, otherwise you get a package made with whatever your dev machine is, which then doesn't match the arch in GCP.

sethtroisi:
I tried verifying the dockerfile but this [line](https://github.com/tensorflow/minigo/blob/master/cluster/base/Dockerfile#L65) failed, I suspect it's wrong but I think that's out of scope for this PR.

amj:
how did it fail?  did you run it via the makefile?

sethtroisi:
I ran `sudo docker build cluster/base/` and that failed because a staging directory wasn't set.

I see now that if I used the makefile it would probably setup a staging directory.

sethtroisi:
I succedded in building this and verified that --copt=ivybridge was passed to both configure_tensorflow and bazel I didn't upload or run the image but I'm considering it 95% verified.

PR #262, opened 2018-06-12
Set gpu_options.allow_growth = true for c++ engine

sethtroisi:
Alternatively I can set "allow_growth = true" but I think this is a good first step (and something now is better than perfect later)


amj:
`allow_growth = true` seems to be more robust, no?

amj:
\lgtm

PR #263, opened 2018-06-12
Quality of Life improvements to GTP engines

sethtroisi:
* Fix cc engine naming
* Change Komi to be more robust to future changes
* Move "showboard" to default python gtp engine
* Remove hardcoded 100 playout config from python gtp engine


dual_net.py
```diff
@@ -29,7 +29,7 @@
 import numpy as np
 import tensorflow as tf
 from tensorflow.python.training.summary_io import SummaryWriterCache
-from tensorflow.contrib.tpu.python.tpu import bfloat16
+#from tensorflow.contrib.tpu.python.tpu import bfloat16
```
sethtroisi: requires TF 1.8 while project README.md lists 1.7

amj: So this isn't even used in the file?  Weird.

sethtroisi: I think Brian was coping the layout of
https://github.com/tensorflow/tpu/blob/5f44b6141cea377a46c912c6b557047922ecdce9/models/official/resnet/resnet_main.py#L151

But it doesn't appear that the precision flag in this file is used

amj: LGTM ok2merge

PR #266, opened 2018-06-14
Move symmetry handling code up into the MctsPlayer code

tommadams:
Note that this PR swaps the order in which the MctsPlayer calls `RevertVirtualLoss` and `IncorporateResults` on selected leaves. Aside from some potential differences in float rounding, this shouldn't cause any change in behaviour.

Testing performed:
* unit tests.
* manually verified using `compare_examples.py` that the examples generated with `--inject_noise=false --soft_pick=false --random_symmetry=false` match head.
* manually verified that it plays the same game with `--inject_noise=false --soft_pick=false --random_symmetry=false`.


cc/main.cc
```diff
@@ -133,9 +133,14 @@ void WriteExample(const std::string& output_dir, const std::string& output_name,
...
+  DualNet::BoardFeatures features;
+  std::vector<const Position::Stones*> recent_positions;
   for (const auto& h : player.history()) {
-    examples.push_back(tf_utils::MakeTfExample(h.node->features, h.search_pi,
-                                               player.result()));
+    h.node->GetMoveHistory(DualNet::kMoveHistory, &recent_positions);
+    DualNet::SetFeatures(recent_positions, h.node->position.to_play(),
+                         &features);
+    examples.push_back(
+        tf_utils::MakeTfExample(features, h.search_pi, player.result()));
```
amj: i assume all the tf-example files generated with this change are the same?

tommadams: Yep, I verified with `compare_examples.py`.

PR #267, opened 2018-06-14
Speedup random.shuffle by converting deque to list

sethtroisi:
I can break out the important change `self.examples = list(self.examples)` if you'd like that

This is a 100% speedup!


example_buffer.py
```diff
@@ -78,24 +82,26 @@ def update(self, new_games, samples_per_game=4):
...
-                game, samples_per_game)]
-            self.examples.extend(choices)
+                print("Found {}/{} new games".format(
+                    len(new_games) - idx, len(new_games)))
+            self.examples.extend(self.func(game))
 
     def flush(self, path):
+        # random.shuffle on deque is O(n^2) convert to list for O(n)
+        self.examples = list(self.examples)
+        random.shuffle(self.examples)
```
brilee: https://bugs.python.org/issue4123
Nice catch. I had always assumed deques were just implemented as arrays with overallocated space on both ends, rather than as a linked list.

example_buffer.py
```diff
@@ -78,24 +82,26 @@ def update(self, new_games, samples_per_game=4):
...
         with timer("Writing examples to " + path):
-            random.shuffle(self.examples)
             preprocessing.write_tf_examples(
                 path, [ex[1] for ex in self.examples])
+        self.examples.clear()
```
brilee: after flush() is called, self.examples used to be of type Deque, but is now of type List. Subsequent fill/flush operations will have degraded complexity.

amj:
is there pending work here?

sethtroisi:
I need to address Brian's comment and do some perf testing

sethtroisi:
Perf data from a 16 core VM with HDD

Run #1
sampling: 464 seconds
*shuffle: 673 seconds*
writing tfexamples: 286 seconds
misc: 26 seconds
*total: 1446 seconds*

Run #2

example_buffer.py
sethtroisi: Done

brilee: still seems to be an issue - see `self.examples = list(self.examples)` a few lines up

brilee: oh, okay, didn't see that.

PR #269, opened 2018-06-15
update dual_net for use with tpus

amj:
Extends @brilee 's TPU work


amj:
run with `python dual_net.py train $STORAGE_BUCKET/models/work_dir_big_batch $DATA_DIR/{300..329}.tfrecord.zz --steps=250000 --use_tpu=True --tpu_name=grpc://10.240.5.2:8470`

dual_net.py
```diff
@@ -39,7 +40,7 @@
 import preprocessing
 import symmetries
 
-flags.DEFINE_integer('train_batch_size', 256,
+flags.DEFINE_integer('train_batch_size', 2048,
                      'Batch size to use for train/eval evaluation. For GPU '
                      'this is batch size as expected. For TPU, this is the '
                      'batch size used for each TPU core (so that effective '
```
sethtroisi: update comment

dual_net.py
```diff
@@ -200,8 +201,8 @@ def model_fn(features, labels, mode, params=None):
         tf.nn.l2_loss(v)
         for v in tf.trainable_variables() if not 'bias' in v.name])
     combined_cost = policy_cost + value_cost + l2_cost
-    boundaries = [40 * int(1e6), 80 * int(1e6)]
-    values = [1e-3, 1e-4, 1e-5]
+    boundaries = [2 * int(1e4), 8 * int(1e4)]
```
sethtroisi: revert theses are place them under a supervised settings flag

dual_net.py
```diff
@@ -389,14 +391,18 @@ def export_model(working_dir, model_path):
...
-            cluster=tpu_cluster_resolver,
+            master=tpu_grpc_url,
+            evaluation_master=tpu_grpc_url,
             model_dir=working_dir,
-            save_checkpoints_steps=max(600, FLAGS.iterations_per_loop),
+            save_checkpoints_steps=max(200, FLAGS.iterations_per_loop),
+            session_config=tf.ConfigProto(allow_soft_placement=True, log_device_placement=True),
             tpu_config=tpu_config.TPUConfig(
                 iterations_per_loop=FLAGS.iterations_per_loop,
                 num_shards=FLAGS.num_tpu_cores,
```
sethtroisi: This is appearently ignored by TPUEstimator

https://www.tensorflow.org/api_docs/python/tf/contrib/tpu/TPUConfig

dual_net.py
```diff
@@ -389,14 +391,18 @@ def export_model(working_dir, model_path):
...
             FLAGS.tpu_name, zone=None, project=None)
+        tpu_grpc_url = tpu_cluster_resolver.get_master()
 
         config = tpu_config.RunConfig(
-            cluster=tpu_cluster_resolver,
+            master=tpu_grpc_url,
+            evaluation_master=tpu_grpc_url,
             model_dir=working_dir,
-            save_checkpoints_steps=max(600, FLAGS.iterations_per_loop),
+            save_checkpoints_steps=max(200, FLAGS.iterations_per_loop),
```
sethtroisi: why are you saving checkpoints so often? seems like this probably takes some amount of time.

sethtroisi: per https://www.tensorflow.org/api_docs/python/tf/contrib/tpu/TPUConfig

we should probably set iterations_per_loop = save_checkpoints_steps = large number
"It is recommended to be set as number of global steps for next checkpoint."

dual_net.py
amj: Which is?

dual_net.py
amj: it didn't seem to make much difference; probably because our model is (relatively) small  If we bump the batch size further (4096?) then we'll have under 1000 steps/golden chunk, which makes this seem relatively infrequent, actually.

dual_net.py
```diff
@@ -54,12 +53,21 @@
...
+flags.DEFINE_multi_float('lr_rates', [2e-2, 1e-3, 1e-4],
+                     'The different learning rates')
+
 flags.DEFINE_float('l2_strength', 1e-4,
                    'The L2 regularization parameter applied to weights.')
 
 flags.DEFINE_float('sgd_momentum', 0.9,
                    'Momentum parameter for learning rate.')
 
+flags.DEFINE_string('model_dir', None,
```
amj: hey @brilee i'm not real psyched about adding a global 'model_dir' flag (which would then be superfluous with a bunch of main.py commands, no?), but i'm not sure what a better way to plumb it into the host_call_fn is.  Ideas?

dual_net.py

Ttl:
Does the current batch normalization momentum work with this large batch size?

Currently the momentum is 0.997 which already seemed slow with the old batch size. It takes 1000 steps to converge to within 5% of the true value (0.997**1000 ~= 0.05). Default momentum in tensorflow is 0.99 and 0.9 in pytorch.

amj:
@Ttl thanks for the comment.  our early tests with this batch size did show that it indeed learned *something* -- but that's by accident.  Honestly i've no idea if this batch-norm momentum value is anything like adequate and i'll start doing some tests today.  Thanks for suggesting other values!

amj:
@brilee the integration test is failing here because of the discrepancies between dual_net.train and main.train -- I'm not really thrilled that the working_dir flag is now so far from the declaration of the command.  Ideas?

brilee:
I had been thinking that train/validate/bootstrap/export should all be top-level methods in dual_net.py

WDYT?

amj:
SGTM

amj:
refactor in the next PR?

dual_net.py
brilee: Another thing that's kind of weird about using this as a flag is that anything importing dual_net (selfplay, etc.) then gets this flag defined. It's sort of harmless in that it's not required, but still makes me feel funny.

dual_net.py
```diff
@@ -247,9 +258,15 @@ def host_call_fn(policy_output, value_output, pi_tensor, policy_cost,
         }
         # Create summary ops so that they show up in SUMMARIES collection
         # That way, they get logged automatically during training
-        for metric_name, metric_op in metric_ops.items():
-            tf.summary.scalar(metric_name, metric_op[1])
-        return metric_ops
+        if FLAGS.use_tpu:
+            with summary.create_file_writer(FLAGS.model_dir).as_default(), summary.record_summaries_every_n_global_steps(FLAGS.iterations_per_loop * 2):
```
brilee: FWIW, the estimator model_fn interface allows config/params as args, which will get passed through. This would be the official way to do it sans FLAGS.

main.py
```diff
@@ -74,24 +74,21 @@ def bootstrap(
...
-        tf_records: 'list of files of tf_records to train on',
+def train(tf_records: 'list of files of tf_records to train on',
         model_save_path: 'Where to export the completed generation.'):
     print("Training on:", tf_records[0], "to", tf_records[-1])
     with utils.logged_timer("Training"):
-        dual_net.train(working_dir, *tf_records)
+        dual_net.train(*tf_records)
     print("== Training done.  Exporting model to ", model_save_path)
-    dual_net.export_model(working_dir, model_save_path)
+    dual_net.export_model(flags.FLAGS.model_dir, model_save_path)
```
brilee: you could update the export_model() function to use flags instead of a function arg

tests/test_dual_net.py
```diff
@@ -20,19 +20,22 @@
...
 import preprocessing
 from tests import test_utils
+from absl import flags
 
 
 class TestDualNet(test_utils.MiniGoUnitTest):
     def test_train(self):
         with tempfile.TemporaryDirectory() as working_dir, \
                 tempfile.NamedTemporaryFile() as tf_record:
+            flags.FLAGS.model_dir = working_dir
```
brilee: I see you can't use the test flagfile because you need to create a tmp directory...

Not sure I like putting model_dir as a FLAG.

amj:
ok, filed https://github.com/tensorflow/minigo/issues/276 to track the work fixing the model_dir flag.

PR #274, opened 2018-06-18
Clean up C++ code

tommadams:
Performs a cleanup pass over the C++ code.

* C++ now builds for a 19x19 board by default (you no longer have to specify --define=board_size=19).
* Specify license types in BUILD files.
* Clean up includes & BUILD dependencies.
* Fix a use after free bug in TryParseKgs.
* Allow inference server to be disabled at compile time by defining MINIGO_DISABLE_INFERENCE_SERVER=1 (for use on platforms that don't support gRPC).
* Reformat some BUILD.


cc/README.md
```diff
@@ -54,15 +45,15 @@ enabled for a particular board size. Consequently, you must run the tests
 multiple times:
 
 ```shell
-bazel test --define=board_size=9 cc/...  &&  bazel test --define=board_size=19 cc/...
+bazel test --define=board_size=9 cc/...  &&  bazel test cc/...
```
amj: do we need to update the test dockerfiles?

tommadams: The tests will continue to work as the --define=board_size=19 will just be ignored. The test dockerfiles should probably be updated to avoid confusion in the future though. I'll create a new issue.

PR #278, opened 2018-06-19
Rotate in chunker

sethtroisi:
not to be sumbitted only for preview / review


sethtroisi:
screenshot from 2018-07-07 15-04-42

PR #283, opened 2018-06-22
Oneoff SWA script

sethtroisi:
This shows good strength so I think we should check it in.

Run with: `BOARD_SIZE=19 python3 oneoffs/swa.py --bucket_name="minigo-pub/v7-19x19"`
http://cloudygo.com/cross-eval/eval-model/2515?sorted=True
screenshot from 2018-06-22 00-26-31


oneoffs/swa.py
```diff
@@ -0,0 +1,96 @@
...
+import tensorflow as tf
+from tensorflow.python.framework import meta_graph
+from absl import flags
+
+import dual_net
+import fsdb
+from utils import ensure_dir_exists
+
+flags.DEFINE_string("data_dir", "data/swa/", "Where to save swa models")
+flags.DEFINE_integer("count", 2, "number of models to save")
```
amj: to save?

amj: consider a better docstring

oneoffs/swa.py
```diff
@@ -0,0 +1,96 @@
...
+    print(stored_var_names)
+    print(len(stored_var_names), len(var_list))
+
+    sessions = [tf.Session() for _ in model_paths]
+    saver = tf.train.Saver(var_list=var_list)
+    for sess, model_path in zip(sessions, model_paths):
+        saver.restore(sess, model_path)
+
+    values = [sess.run(var_list) for sess in sessions]
+
```
amj: consider adding a comment or two here

PR #284, opened 2018-06-23
Bootstrap function fix

josephch405:
Bootstrapping directly seems to not work without this change


amj:
looks good, thanks for the fix!

PR #285, opened 2018-06-24
Adjust c_puct to 0.96

sethtroisi:
See analysis in #281


cc/constants.h
```diff
@@ -34,7 +34,7 @@ constexpr float kDefaultKomi = 7.5;
 
 constexpr float kDirichletAlpha = 0.03f * 361 / (kN * kN);
 
-static constexpr float kPuct = 1.38;
+static constexpr float kPuct = 0.96;
```
amj: this should probably be extracted to a flag in both places.

sethtroisi: I'll file an issue but IMO it's not inscope to change code in a small PR like this.

PR #291, opened 2018-06-26
Make dual_net.py model setup more verbose

sethtroisi:
Model setup has numerous ambiguities this change aim to make our decisions more obvious. A future PR will likely modify some of these parameters.

This is a functional no-op but helps clear the way for changing center and scale on my_batchn


dual_net.py
```diff
@@ -331,19 +343,18 @@ def my_res_layer(inputs):
         shared_output = my_res_layer(shared_output)
 
     # policy head
-    policy_conv = tf.nn.relu(my_batchn(
-        my_conv2d(shared_output, filters=2, kernel_size=[1, 1]),
-        center=False, scale=False))
+    policy_conv = my_conv2d(shared_output, filters=2, kernel_size=1)
+    policy_conv = tf.nn.relu(my_batchn(policy_conv, center=True, scale=True))
```
brilee: this is changing current semantics, right? I'm okay with the change, just highlighting that this isn't a no-op PR.

amj: yes, this is a change -- deliberate?

PR #292, opened 2018-06-26
Small model tweaks

sethtroisi:
- Remove bias on [conv2d](https://www.tensorflow.org/api_docs/python/tf/layers/conv2d) layers, this is redundant because all my_conv2d are followed by my_batchn with `scale=True`
- stop regularizating [my_batchn bias](https://www.tensorflow.org/api_docs/python/tf/layers/batch_normalization) which is named `beta`
- ~~Turn on `center=True` for final value conv2d~~

This builds on https://github.com/tensorflow/minigo/pull/291

### DIFF
--- a/dual_net.py
+++ b/dual_net.py
@@ -209,11 +209,11 @@ def model_fn(features, labels, mode, params=None):
             logits=logits, labels=tf.stop_gradient(labels['pi_tensor'])))
     value_cost = tf.reduce_mean(
         tf.square(value_output - labels['value_tensor']))
-    # TODO(sethtroisi) Add 'beta' to reg_vars
     reg_vars = [v for v in tf.trainable_variables()
-        if not '/bias:' in v.name]
+                if not '/bias:' in v.name and not '/beta:' in v.name]
 
-    l2_cost = FLAGS.l2_strength * tf.add_n(tf.nn.l2_loss(v) for v in reg_vars)
+    l2_cost = FLAGS.l2_strength * \
+        tf.add_n([tf.nn.l2_loss(v) for v in reg_vars])
     combined_cost = policy_cost + value_cost + l2_cost
     learning_rate = tf.train.piecewise_constant(
         global_step, FLAGS.lr_boundaries, FLAGS.lr_rates)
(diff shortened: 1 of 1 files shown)


dual_net.py
```diff
@@ -209,9 +209,11 @@ def model_fn(features, labels, mode, params=None):
             logits=logits, labels=tf.stop_gradient(labels['pi_tensor'])))
     value_cost = tf.reduce_mean(
         tf.square(value_output - labels['value_tensor']))
-    l2_cost = FLAGS.l2_strength * tf.add_n([
-        tf.nn.l2_loss(v)
-        for v in tf.trainable_variables() if not 'bias' in v.name])
+    reg_vars = [v for v in tf.trainable_variables()
+                if not '/bias:' in v.name and not '/beta:' in v.name]
```
brilee: i'm having trouble parsing this boolean logic. So the variables you want to regularize are:
- All variables
- except for bias variables
- but actually, regularize bias variables if they're part of batchnorm

I think this is slightly less complicated, although it's still a bit hard to read.
`v for v in trainable_variables() if 'batch' in v.name or not 'bias' in v.name`

I think the right solution is to abstract out a function `should_regularize(v)` and write unit tests for it to sanity check that we're doing what we think we're doing..

sethtroisi: You can see the list of variables in https://github.com/tensorflow/minigo/pull/294#issuecomment-400506907

currently `tf.trainable_variables` is:
batch_normalization_<X>/beta:0
batch_normalization_<X>/gamma:0
conv2d_<X>/bias:0
conv2d_<X>/kernel:0
dense_<X>/bias:0
dense_<X>/kernel:0

But we exclude the conv2d and dense bias. I thought it made sense that we should also exclude batch_normalization beta (which I thought was essentially a bias)

sethtroisi:
This change is backwards compatible; The frozen pb models don't change, and loading the models with python works except that the bias weights are ignored (which results in a slightly weaker model).

```
minigo-cc-303 v minigo-cc-303-no-bias (100/100 games)
board size: 19   komi: 7.5
minigo-cc-303             56 56.00%       30 60.00%     26 52.00%    110.95
minigo-cc-303-no-bias     44 44.00%       24 48.00%     20 40.00%    106.45
                                          54 54.00%     46 46.00%
```

amj:
I think this addresses my concerns about backwards compatibility; i don't think we should go back and re-freeze the old models though, as they were trained with the extra bias terms.

I'm not confident enough to sign off on the 'regularize our batchnorm bias' part of the change, though.  @brilee ?

sethtroisi:
I updated the desc, technically this change is to stop regularizating of my_batch bias

PR #295, opened 2018-06-27
Only add D-noise to legal moves

sethtroisi:
Per [Tom's CC tests](https://github.com/tensorflow/minigo/blob/master/cc/random_test.cc#L82), 61% of the weight is on a single weight and 21% is in the 2nd weight. for 361 it's more like 19% and 12%

Given that later in the game \~most\~ positions are illegal (already contain a stone) this means that we when we add policy + noise, policy is mostly unchanged (50% of the time the largest change from noise will be less than 0.25 *  0.12 = 0.04)


amj:
can you explain "50% of the time the largest change from noise will be less than 0.25 * 0.12 = 0.04" a little further?  Where does that "50% of the time" come from?

mcts.py
```diff
@@ -238,10 +238,12 @@ def is_done(self):
         return self.position.is_game_over() or self.position.n >= FLAGS.max_game_length
 
     def inject_noise(self):
-        dirch = np.random.dirichlet(
-            [FLAGS.dirichlet_noise_alpha] * ((go.N * go.N) + 1))
+        epsilon = 1e-5
+        legal_moves = (1 - (self.illegal_moves / 1000)) + epsilon
```
brilee: this is sort of weird to recompute the legal moves from the illegal moves, but I guess I don't care about inefficiency since d-noise injection only happens once per move.

Can you abstract out the 1000 as a constant? I chose 1000 because in the worst case, if an illegal move is given prior of 1.0, then it will take ~1,000,000 searches at a node before the action score brings up an illegal move as the top action score (due to the sqrt(parent-N) term in the U calculation), and I figured 1,000,000 searches was probably not going to happen with the python impl, given memory constraints. 

(this reminds me that the C++ impl might actually be able to hit 1,000,000 searches and so 1000 is no longer a suitable value for this constant) FYI @tommadams

mcts.py
```diff
@@ -238,10 +238,12 @@ def is_done(self):
...
+        a = legal_moves * ([FLAGS.dirichlet_noise_alpha] * (go.N * go.N + 1))
```
brilee: I think this works too and is cleaner: `dirichlet = np.random.dirichlet(legal_moves * FLAGS.dirichlet_noise_alpha)`

sethtroisi:
@amj simulations show that the d-noise averages 19% on the top weight, and 12% on the 2nd largest weight (for 362 weights). I'm assuming that 1/2 of the board is illegal to play in (filled with stone or is self-capture).

We transform weights `0.75 * policy + 0.25 * d-noise`
1/2 of the time the index of the largest d-noise weight will be an illegal move and adding 0.25 * 0.19 = 0.05 to it's policy value doesn't matter (because it will never be expanded). The second largest weight get's modified by adding `0.25 * 0.12 = 0.03` to it's policy value which isn't a big change.

PR #297, opened 2018-06-30
Handle 'tt' as pass on 19x19 boards

sethtroisi:


amj:
[tt] as pass was a hack that's kinda-sorta deprecated.  They mention in the spec that it's still there for backwards compatibility.

i guess the natural question is -- where are you getting sgf's that use [tt] for pass?

PR #298, opened 2018-07-01
tf.metrics.mean != tf.reduce_mean

sethtroisi:
See discussion in https://github.com/tensorflow/tensorflow/issues/15115

TL;DR: tf.metrics.mean is averaged over the entire tf session where tf.reduce_mean is over a single batch.

We would optimally use some running average to reduce noise on the metric but keep it close to the current value.

Currently if a value (policy_accuracy) linearly increases over the run tensorboard will show approximately 1/2 of it's current value because the sum of all it's values (0.01, 0.02, 0.03 ... 0.50 = .025) is about half it's most recent batch.


dual_net.py
```diff
@@ -244,20 +244,27 @@ def host_call_fn(policy_output, value_output, pi_tensor, policy_cost,
...
+            'combined_cost': tf.reduce_mean(combined_cost),
 
             'policy_accuracy_top_1': tf.metrics.accuracy(
                 labels=policy_target_top_1, predictions=policy_output_top_1),
-            'policy_accuracy_top_3': tf.metrics.mean(policy_output_in_top3),
-            'policy_top_1_confidence': tf.metrics.mean(policy_top_1_confidence),
-            'policy_target_top_1_confidence': tf.metrics.mean(
+            'policy_accuracy_top_3': tf.reduce_mean(policy_output_in_top3),
+            'policy_top_1_confidence': tf.reduce_mean(policy_top_1_confidence),
+            'policy_target_top_1_confidence': tf.mreduce_mean
```
amj: typo

amj:
ahh, looks good.  Have you tried this on a SL run or a running start?

amj:
is this obsolete now?

PR #299, opened 2018-07-01
Zero out illegal moves

sethtroisi:
I loaded a self-play sgf with 130 moves and ran "genmove b" to print the sum of policy of legal moves which was ~0.98 meaning 2% of policy mass is on illegal moves (looking at that board I suspect these are occupied coords not self-capturing moves)

I think a human explanation for the current behavior would be "a dynamic c_puct that values policy less as more stones are filled on the board" which is interesting (and maybe something we should test) but not what the paper suggests.

I'm running 100 games right now comparing current (default) behavior and new zeroing behavior.
Do you think I should use `-inject-noise` to better assess how that flag will interact with this change or should I not add that extra noise to the eval games?

In DM's literature I found
Alpha Go paper: "A final softmax layer outputs a probability distribution over all legal moves a"
AlphaGoZero paper: "In addition, the player is provided with the set of legal moves in each position"


cc/mcts_node.h
```diff
@@ -126,7 +126,7 @@ class MctsNode {
   std::array<EdgeStats, kNumMoves> edges;
 
   // TODO(tommadams): a more compact representation.
```
amj: this would fix that 'compactness' TODO, i think ;)

mcts.py
```diff
@@ -207,7 +209,24 @@ def incorporate_results(self, move_probabilities, value, up_to):
             self.revert_visits(up_to=up_to)
             return
         self.is_expanded = True
-        self.original_prior = self.child_prior = move_probabilities
+
+        '''
+        temp = move_probabilities * (1 - self.illegal_moves)
```
amj: is this ready for review?

mcts.py
```diff
@@ -207,7 +209,24 @@ def incorporate_results(self, move_probabilities, value, up_to):
...
+        new_s = sum(temp)
+
+        print ("{:.4f}   {:.4f}  {:.4f}".format(original - new_s, original, new_s))
+        assert abs(1 - original) < 1e-5
+        '''
+
+        # Zero out illegal moves.
+        move_probs = move_probabilities * (1 - self.illegal_moves)
+        scale = sum(move_probs)
+        if scale > 0:
```
amj: why would scale be <= 0?  No legal moves?

sethtroisi: Yes or a broken network file

cc/mcts_node.h
sethtroisi: This would still be 3Gigs of memory overhead at 1M playouts.
it can be compacted by a factor of 8 with clever boolean logic (or vector<bool>) or removed entirely by integrating it into U

mcts.py
sethtroisi: this was here to demonstrate my methodology for determining rough impact I should have commented "WILL BE REMOVED BEFORE SUBMIT"

amj:
this is a real broken test at the moment, fyi.

cc/mcts_node.cc
```diff
@@ -186,9 +186,24 @@ void MctsNode::IncorporateResults(absl::Span<const float> move_probabilities,
...
+  for (int i = 0; i < kNumMoves; ++i) {
+    if (!illegal_moves[i]) {
+        policy_scalar += move_probabilities[i];
+    }
+  }
+  if (policy_scalar > std::numeric_limits<float>::min()) {
+    policy_scalar = 1 / policy_scalar;
+  }
+
+  std::cout << "What what " << policy_scalar << std::endl;
```
josephch405: Assuming this was for debugging? :D

sethtroisi: Yes, I need to re-push the new version with tests.

sethtroisi:
Updated with tests, ready for a new look.

cc/mcts_node.h
tommadams: Isn't it more like 362MB?
I don't think moving to bit twiddling is really necessary at this point but if it were, we should be able to just drop in std::bitset.

cc/position.h
```diff
@@ -30,6 +30,10 @@
 #include "cc/stone.h"
 
 namespace minigo {
+ namespace position {
```
tommadams: Can you remove this namespace please?
I'm actually surprised the test compiles, since it doesn't look like the function definition in the .cc file is in the same namespace. Did you run it with `--define=board_size=9`?

cc/position.h
```diff
@@ -30,6 +30,10 @@
...
+  inline_vector<Coord, 4>& GetNeighborCoords(Coord c);
```
tommadams: Return a const reference

cc/position.cc
```diff
@@ -51,12 +51,13 @@ std::array<inline_vector<Coord, 4>, kN* kN> kNeighborCoords = []() {
   return result;
 }();
 
-inline const inline_vector<Coord, 4>& GetNeighborCoords(Coord c) {
+}  // namespace
+
+
+inline_vector<Coord, 4>& position::GetNeighborCoords(Coord c) {
```
tommadams: It'd be nice to check the generated code to verify that the compiler still inlines the GetNeighborCoords calls:

```
bazel build -c opt --save_temps cc:position
view bazel-bin/cc/_objs/position/cc/position.pic.s
```

cc/position.h
sethtroisi: Done.

cc/position.h
sethtroisi: the cc file defines in in minigo::position namespace.
I was trying to keep it out of the global minigo namespace, is this not the right choice?

cc/position.cc
sethtroisi: Didn't know I could do this :) and it's good you mentioned to check because they are no longer inline. I'll follow up tomorrow.

cc/position.h
tommadams: We don't really use inner namespaces on minigo, the project really isn't big enough to warrant them. I think the only one we have is minigo::sgf and that's because it felt weird to inject a name like Ast into the global namespace. In retrospect I should have just called the class SgfAst for consistency and not added the sgf inner namespace.

cc/position.h
tommadams: I don't see the change, do you need to push or something?

cc/position.h
```diff
@@ -30,6 +30,8 @@
 #include "cc/stone.h"
 
 namespace minigo {
+  // Defined only for tests.
+  const inline_vector<Coord, 4>& GetNeighborCoords(Coord c);
```
tommadams: remove indentation

cc/position.cc
```diff
@@ -51,12 +51,12 @@ std::array<inline_vector<Coord, 4>, kN* kN> kNeighborCoords = []() {
   return result;
 }();
 
-inline const inline_vector<Coord, 4>& GetNeighborCoords(Coord c) {
+}  // namespace
+
+const inline_vector<Coord, 4>& GetNeighborCoords(Coord c) {
```
tommadams: I thought these calls weren't getting inlined now? Or is the compiler able to inline the calls from position.cc?

PR #304, opened 2018-07-02
Update cc-base default tag

sethtroisi:
"latest" is just a tag and it's out of date.
Alternatively I should update the "latest" tag and point everything at that


cluster/evaluator/Dockerfile-cc
```diff
@@ -1,4 +1,4 @@
-FROM gcr.io/$PROJECT/cc-base:latest
+FROM gcr.io/$PROJECT/cc-base:0.17
```
amj: this is fine, you're welcome to commit as-is; you could also have changed just this one to use 0.1.

PR #306, opened 2018-07-06
Macro for MG_GTP_CMD

sethtroisi:
I don't know how people on this project feel about macros.

This is okay, so I thought I'd put it out there and see what people say.


tommadams:
I was expecting something a lot more complicated after just reading the PR title ;)

I'm not sure how I feel about this. On the one hand, removing the line wrapping does make the declarations a lot more readable. On the other, well... it's macros.

How about providing an alias for the arguments type:
  using Args = std::vector<absl::string_view>;

Then we'd have:
  Response HandleReportSearchInterval(absl::string_view cmd, const Args& args)

sethtroisi:
I like the suggestion, getting the signature to fit on a single line would be a big improvement and I can reuse Args in the cc file as well.

amj:
this looks ready?

PR #307, opened 2018-07-06
Add "benchmark" gtp command

sethtroisi:
Allows much faster testing of various batch_sizes

format is "benchmark [num_readouts] [batch_size]"

image


amj:
Is it useful to log the batch size (8/16/32), esp if the log string is going to be part of genmove?

amj:
(Also, try this on CGOS mode)

sethtroisi:
cc doesn't have CGOS mode but I tried with `--seconds_per_move 10`

`Milliseconds per 100 reads: 208ms over 4792 readouts (batched: 8)`

PR #308, opened 2018-07-10
Cleanups

sethtroisi:
It's a little weird that nothing was depending on self.losses_applied but I guess self.W manages the "lock"


amj:
the tests fail for real: the local_rl_loop test fails.

PR #310, opened 2018-07-11
Create dockerfiles and deployments for tpu selfplay

amj:


cc/mcts_player.cc
```diff
@@ -34,8 +34,7 @@ std::ostream& operator<<(std::ostream& os, const MctsPlayer::Options& options) {
      << " soft_pick:" << options.soft_pick
      << " random_symmetry:" << options.random_symmetry
      << " resign_threshold:" << options.resign_threshold
-     << " batch_size:" << options.batch_size
-     << " komi:" << options.komi
+     << " batch_size:" << options.batch_size << " komi:" << options.komi
```
amj: So this is driving me bananas but some part of the fancy google magic vim thing is insisting on autoformatting on save.

cluster/selfplay/tpu_wrapper.sh
```diff
@@ -0,0 +1,36 @@
...
+  --model=gs://tensor-go-minigo-v7-19/models/000485-onslaught \
+  --inject_noise=true \
+  --soft_pick=true \
+  --random_symmetry=true \
+  --virtual_losses=8 \
+  --games_per_inference=16 \
+  --parallel_games=32 \
+  --num_readouts=800 \
+  --resign_threshold=-0.95 \
+  --output_dir=gs://tmadams-sandbox/data/selfplay \
```
tommadams: do you want to change these directories?

amj: yep, definitely; i figure similar to how we won't be hardcoding the model either, we'll want to rework and parameterize the model paths we're looking at as well.  probably should extract them into env vars that are set in the docker config, which are then also set via the 'launch deployment' script.

cc/dual_net/inference_server.cc
```diff
@@ -161,6 +160,8 @@ class InferenceServiceImpl final : public InferenceService::Service {
       game.notification->Notify();
     }
 
+    // std::cerr << absl::Now() << " DONE  PutOutputs\n";
```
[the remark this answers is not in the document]
amj: I removed the other ones too.  Done.

cluster/selfplay/Dockerfile.tpu
```diff
@@ -0,0 +1,38 @@
...
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+
+ADD staging/requirements.txt /app/requirements.txt
+
+RUN pip3 install --upgrade pip
+RUN pip3 install -r /app/requirements.txt
+RUN pip3 install "tensorflow==1.9.0rc2"
```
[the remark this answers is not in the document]
amj: nope, our docs aren't yet tested on tf-gpu 1.9 with whatever version of cuda; it will probably not work.  As it is, this is just the TPU dockerfile that will use TF 1.9.

PR #311, opened 2018-07-11
Always use tf.contrib.summary

sethtroisi:
tf.contrib.summary is not as well supported for tf.estimator as tf.summary but TPU requires tf.contrib.summary

Old: tf.summary + runconfig
```
tf.summary.scalar("name", tensor)
run_config = tf.estimator.RunConfig(save_summary_steps=500)
tf.estimator.Estimator(config=run_config)
```
and the op to save summary is done automagically by estimator

New: tf.contrib.summary
```
with tf.contrib.summary.create_file_writer(...).as_default:
    with tf.contrib.summary.always_record_summaries():
        tf.contrib.summary("name", tensor)
```

but nothing executes tf.contrib.summary.all_summary_ops()
so now we add a host_call that returns all_summary_ops() so that summary gets evaled occasionally.
To make sure this keeps working for CPU/GPU we use as_estimator_spec which converts the host_call into a hook when not onnot on TPU.


sethtroisi:
I tested with both TPU and GPU training and verified everything (metrics + hooks) work in both environments.

amj:
So this will write out summaries when it writes a checkpoint?

PR #314, opened 2018-07-12
Optionally monitor directory for updated models

tommadams:
- Adds a new checkpoint_dir flag to inference_worker.py. If set, the worker spawns a new thread that monitors a directory for new models and updates the model used for subsequent inferences.
- Updates the C++ code to keep track of which model was used for each TreeSearch step, and add that info to each move comment in the full SGF.


inference_worker.py
```diff
@@ -60,21 +65,39 @@
...
+NUM_WORKER_THREADS = 2
+
+# Print to stderr and flush by default.
+print = functools.partial(print, file=sys.stderr, flush=True)
+
+
+class RwLock(object):
+    """A simple read/write mutex.
+
+    I'm kind of surprised Python doesn't provide one of these by default.
```
amj: nit: drop "kind of" for more impact ;)

inference_worker.py
```diff
@@ -94,85 +117,174 @@ def custom_getter(getter, name, *args, **kwargs):
...
+            self.feature_placeholders = tuple(self.feature_placeholders)
+
+        if FLAGS.model:
+            self._load_model(FLAGS.model)
+
+    def _run_threads(self):
+        """Run inference threads and optionally a thread that updates the model.
+
+        Synchronization between the inference threads and the model update
+        thread is performed using a RwLock that protections access to self.sess.
```
amj: nit protects

Issue #317, opened 2018-07-13
Undefined names 'good_size' and 'pi1'


cclauss:
Undefined names can potentially raise NameError at runtime.

flake8 testing of https://github.com/tensorflow/minigo on Python 3.6.3

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./oneoffs/rotate_examples.py:71:25: F821 undefined name 'good_size'
        error = (size - good_size) / (in_size + 1)
                        ^
./tests/test_preprocessing.py:136:17: F821 undefined name 'pi1'
                pi1.reshape((go.N, go.N)), pi2((go.N, go.N)))
                ^
2       F821 undefined name 'good_size'
2
```

sethtroisi:
these look like they are my mistakes, thanks for finding them :)

PR #318, opened 2018-07-13
Typo in Variable name: pi1 --> pi

cclauss:
__pi1__ is an undefined name in this context.    Discovered via #317


amj:
heh, nice find!

cclauss:
It would be super cool if we could teach @k8s-ci-robot a new command, __/flake8-core__ which would execute __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__ which is a really nice sanity check of Python code.

And maybe another new command __/flake8-strict__ which would execute __flake8 . --count --show-source --statistics__

PR #322, opened 2018-07-17
TPU random_rotation

sethtroisi:


symmetries.py
```diff
@@ -82,3 +83,46 @@ def randomize_symmetries_feat(features):
 def invert_symmetries_pi(symmetries, pis):
     return [apply_symmetry_pi(invert_symmetry(s), pi)
             for s, pi in zip(symmetries, pis)]
+
+
+def rotate_train(x, pi):
+    sym = tf.random_uniform(
```
sethtroisi: I used tf.Print to verify this took on multiple values during tests but I couldn't verify on the TPU (because op happens on the TPU)

preprocessing.py
```diff
@@ -210,7 +220,12 @@ def get_tpu_input_tensors(batch_size, tf_records, num_repeats=1,
         functools.partial(batch_parse_tf_example, batch_size))
 
     if random_rotation:
-        raise ValueError("random_rotation not yet supported on TPU")
+        # Unbatch the dataset so we can rotate it
```
brilee: I don't mind if a minibatch uses exactly the same rotation - all that I care about is that the 4 positions pulled from 1 game get assigned different rotations. And since the 4 positions are highly unlikely to be in the same minibatch, they're likely to be assigned randomly different rotations. Unbatching might have undesirable performance side effects.

symmetries.py
```diff
@@ -82,3 +83,47 @@ def randomize_symmetries_feat(features):
...
+        [],
+        minval=0,
+        maxval=len(SYMMETRIES),
+        dtype=tf.int32,
+        seed=123)
+
+    def rotate(tensor):
```
brilee: This is fricking cool.

tests/test_preprocessing.py
```diff
@@ -217,5 +219,6 @@ def test_tpu_rotate(self):
             syms.extend(
                 map(lambda r: self.find_symmetry(*r), zip(x, pi, x2, pi2)))
 
+        difference = set(symmetries.SYMMETRIES) - set(syms)
```
brilee: I'm not 100% sure I understand how this test works - could you explain?

sethtroisi: run_one, run_two, and run_three read 100 examples from a tf_examples file.
run_two and run_three are the same because I call reset_random() (and because the random_uniform has a seed) this is deterministic.
run_one should have no rotation.

I then use "find_symmetry" to find which sym makes `apply_sym(sym, run_two[i]) == run_one[i]`
This is nice because it uses a second, well tested, implementation to verify `symmetries._rotate_train()` code path 

This final line is verifying that some position in run_two was rotated (or technically rotated inversely) with each symmetries so that I know I'm doing all rotations (and not forgetting some symmetries or wrongly implementing them)

symmetries.py
sethtroisi: Yes, much better than what I had before :)

preprocessing.py
sethtroisi: Andrew ran this last night and I there's no difference in performance.

PR #326, opened 2018-07-18
Add value_head_weight for supervised training

sethtroisi:


dual_net.py
```diff
@@ -62,6 +62,10 @@
 flags.DEFINE_float('l2_strength', 1e-4,
                    'The L2 regularization parameter applied to weights.')
 
+flags.DEFINE_float('value_cost_scalar', 1.0,
```
brilee: `value_cost_weight`?

PR #328, opened 2018-07-18
Add tie breaker to fix num_readouts=1

### LINKED ISSUE #260: num_readouts=1 seems broken
genmove b
A19
genmove w
Illegal move (tries to play on A19 again?)

sethtroisi:
Fixes #260 

Fix an off by one difference between `play_move` and `suggest_move` in squash


strategies.py
```diff
@@ -175,14 +173,17 @@ def pick_move(self):
...
+                self.root.position.n < self.temp_threshold):
+            cdf = self.root.children_as_pi(squash=True).cumsum()
             cdf /= cdf[-2]  # Prevents passing via softpick.
             selection = random.random()
             fcoord = cdf.searchsorted(selection)
             assert self.root.child_N[fcoord] != 0
+        else:
+            # break ties in N by looking at action_score
+            fcoord = np.argmax(self.root.child_N +
+                               self.root.child_action_score / 1000)
```
sethtroisi: IDK what to do here anything larger than `/ 2` is likely to work in practice but  `/ (1000 + 2 * c_puct)` might be more technically correct

strategies.py
```diff
@@ -175,14 +173,17 @@ def pick_move(self):
 
         Highest N is most robust indicator. In the early stage of the game, pick
         a move weighted by visit count; later on, pick the absolute max.'''
-        if self.root.position.n >= self.temp_threshold:
-            fcoord = np.argmax(self.root.child_N)
-        else:
-            cdf = self.root.child_N.cumsum()
+        if (not self.two_player_mode and
+                self.root.position.n < self.temp_threshold):
+            cdf = self.root.children_as_pi(squash=True).cumsum()
```
sethtroisi: This is a change in behavior (to select from the squashed distribution), This seems like what we SHOULD have been doing all along but maybe I misunderstand the code

brilee: I'm not sure I'm fully on board with the squash thing; it always seemed a bit voodoo to me.

amj: `temp_threshold` (the move at which we switch from softpick to argmax) is set to be -1 if `two_player_mode` is true.  So this doesn't change behavior.

the squash thing is a little weird.

strategies.py
amj: i think this works fine.

Do we need to add this to C++ to keep their behaviors the same?

strategies.py
amj: (this means the check for `self.two_player_mode` is not needed and can be removed)

PR #336, opened 2018-07-21
 Rewrite inference_worker.py to support GPU inference

tommadams:


inference_worker.py
```diff
@@ -116,24 +135,212 @@ def custom_getter(getter, name, *args, **kwargs):
...
+    def __init__(self, tpu_name, parallel_tpus, batch_size):
+        tpu = [tpu_name] if tpu_name else None
+        tpu_grpc_url = tf.contrib.cluster_resolver.TPUClusterResolver(
+            tpu=tpu).get_master()
+        sess = tf.Session(tpu_grpc_url)
+        Session.__init__(self, sess)
+
+        self._parallel_tpus = parallel_tpus
+        self._tpu_init = tf.contrib.tpu.initialize_system()
+        self._tpu_shutdown = tf.contrib.tpu.shutdown_system()
```
amj: oh weird, this must be an op.  You gotta admit, not knowing anything about it, it looks like you're just init'ing and then shutting down the system right here.

tommadams: good point, added a comment

PR #343, opened 2018-07-24
Trainer changes

amj:


dual_net.py
```diff
@@ -265,16 +265,16 @@ def eval_metrics_host_call_fn(policy_output, value_output, pi_tensor, policy_cos
             'value_confidence': tf.metrics.mean(tf.abs(value_output)),
         }
 
+        if est_mode == tf.estimator.ModeKeys.EVAL:
+            return metric_ops
         # Create summary ops so that they show up in SUMMARIES collection
         # That way, they get logged automatically during training
         summary_writer = summary.create_file_writer(FLAGS.model_dir)
         with summary_writer.as_default(), \
-                summary.record_summaries_every_n_global_steps(FLAGS.summary_steps):
```
brilee: right there in the docs:
Sets the should_record_summaries Tensor to true if global_step % n == 0.
*sadface*

sethtroisi: After model ~50 we collect 2048*1024 positions 4 positions * 500,000 games and train for 1024 steps.
But for the first few models we train fewer steps. For the first model it was ~20 = 4 positions * 12,000 games / 2048.
This means that global_step will be some random number and that global step at the host_call won't evenly divide 128 (and won't be recorded which is what happened to our metrics in v8)

dual_net.py
```diff
@@ -265,16 +265,16 @@ def eval_metrics_host_call_fn(policy_output, value_output, pi_tensor, policy_cos
             'value_confidence': tf.metrics.mean(tf.abs(value_output)),
         }
 
+        if est_mode == tf.estimator.ModeKeys.EVAL:
```
sethtroisi: I wonder if moving this will fix the weird bumps on the summaries.
Previously we might have been updating the summary scalars but not writing them to the file or something.

preprocessing.py
```diff
@@ -206,7 +206,7 @@ def get_input_tensors(batch_size, tf_records, num_repeats=None,
 def get_tpu_input_tensors(batch_size, tf_records, num_repeats=1,
                           shuffle_records=True, shuffle_examples=True,
                           shuffle_buffer_size=1024,
-                          filter_amount=1, random_rotation=False):
```
sethtroisi: Should be fixed by #342

PR #344, opened 2018-07-24
changes from live gatherer box

amj:


example_buffer.py
```diff
@@ -68,20 +68,18 @@ def parallel_fill(self, games, threads=8):
         """ games is a list of .tfrecord.zz game records. """
         games.sort(key=os.path.basename)
         # A couple extra in case parsing fails
-        max_games = self.max_size // self.samples_per_game + 10
+        max_games = (self.max_size // self.samples_per_game) + 480
         if len(games) > max_games:
             games = games[-max_games:]
```
brilee: you can do
`games = games[-max_games:]` without the conditional and it will just work.

example_buffer.py
```diff
@@ -25,12 +25,12 @@
...
-MINIMUM_NEW_GAMES = 5000
+MINIMUM_NEW_GAMES = 12000
 
 
 def pick_examples_from_tfrecord(filename, samples_per_game=4):
     protos = list(tf.python_io.tf_record_iterator(filename, READ_OPTS))
-    if len(protos) < 20:  # Filter games with less than 20 moves
+    if len(protos) < 50:  # Filter games with less than 20 moves
         return []
     choices = random.sample(protos, min(len(protos), samples_per_game))
```
brilee: not going to go with the len(protos) // 50 idea?

example_buffer.py
```diff
@@ -68,20 +68,18 @@ def parallel_fill(self, games, threads=8):
         """ games is a list of .tfrecord.zz game records. """
         games.sort(key=os.path.basename)
         # A couple extra in case parsing fails
-        max_games = self.max_size // self.samples_per_game + 10
+        max_games = (self.max_size // self.samples_per_game) + 480
```
brilee: why 480?

example_buffer.py
amj: `// 50` is for when a game is in the window for ~50 generations, correct?  That's probably not the ratio anymore...

example_buffer.py
amj: The consequences of not having enough positions was worse before #329

example_buffer.py
amj: ack.

PR #346, opened 2018-07-25
Reset tf.metrics after summary logs

sethtroisi:
This is a better fix than #298.

The core problem is that in a log supervised run the metrics are about 1/2 of what they current are (e.g. accuracy_top_1 which improves linearly will report 0.15 accuracy but recent batches will be closer to 0.30 accuracy)


dual_net.py
```diff
@@ -273,9 +277,14 @@ def eval_metrics_host_call_fn(policy_output, value_output, pi_tensor, policy_cos
             for metric_name, metric_op in metric_ops.items():
                 summary.scalar(metric_name, metric_op[1])
 
-        if est_mode == tf.estimator.ModeKeys.EVAL:
-            return metric_ops
-        return summary.all_summary_ops()
+        # Reset metrics occasionally so that they are mean of recent batches.
+        reset_op = tf.variables_initializer(tf.local_variables("mean"))
```
sethtroisi: This doesn't work for policy_accuracy_top1 which is tf.metric.accuarcy

I'd suggest resolving the TODO above and wrapping the metric_ops in a variable_scope("metrics") then using that scope here

amj: ok -- do you want to do that in this PR then?

dual_net.py
```diff
@@ -250,21 +252,21 @@ def eval_metrics_host_call_fn(policy_output, value_output, pi_tensor, policy_cos
             tf.one_hot(policy_target_top_1, tf.shape(policy_output)[1]))
 
         # TODO(sethtroisi): For V10 add tf.variable_scope for tf.metrics.mean's
```
amj: is the TODO done now?

PR #347, opened 2018-07-26
Enable configuration of which inference engines are compiled into C++ binary

tommadams:
- I've replaced the previous \*PlayerFactory classes with \*DualNetFactories, which a more sensible abstraction.
- Compilation of TfDualNet and InferenceServer is controlled by two command line defines `--define=tf={0,1}` and `--define=remote={0,1}`. The newly added `.bazelrc` defaults these options to `1` for backwards compatibility.
- `cc/dual_net/factory.cc` is a complete mess of `#ifdef`s but at least it's all in one place and hidden from the rest of the code.
- The `--remote_inference` flag has been replaced with an `--engine` flag that should be set to either `tf` or `remote`. This is done in anticipation of adding a third `tflite` inference engine.

Note that compiling with `--define=tf=0` still doesn't remove our dependency on C++ TensorFlow because we need to use the TF utilities to write the examples and SGFs to GCS, but it's a step in the right direction.


sethtroisi:
Do you want remote=1 in the bazelrc? it seems like that's the exceptional case and remote=0 is the more common case.

tommadams:
There's reasonable arguments to be made each way. My primary motivation for adding remote=1 in this  CL was making sure there was no behavioural change in the refactor.

Having said that, I'd wouldn't mind if remote=1 become the default, because it means people won't have to deal with manually compiling C++ TensorFlow before they can get up and running for the first time. Unfortunately, we're still relying on TensorFlow for GCS access & writing example files. Splitting cc/main.cc into separate binaries will help there.

PR #349, opened 2018-07-27
Dynamic q

sethtroisi:
Building on @killerducky's #87 

```
minigo-default-q v minigo-dynamic-q (500/500 games)
board size: 19   komi: 7.5
                   wins              black          white        avg cpu
minigo-default-q    112 22.40%       50  20.00%     62  24.80%     91.07
minigo-dynamic-q    388 77.60%       188 75.20%     200 80.00%     93.23
                                     238 47.60%     262 52.40%
```


killerducky:
You might get some weird dependence on how many threads you run etc if the dynamic Q includes virtual losses. It will also act as a sort of Q-reduction. For LZChess they wrote a separate function that returns the real current Q instead of Q including virtual losses for this.

sethtroisi:
```
minigo-default-q v minigo-dynamic-q (168/500 games)
board size: 19   komi: 7.5
minigo-default-q     34 20.24%       15 17.86%     19 22.62%     91.69
minigo-dynamic-q    134 79.76%       65 77.38%     69 82.14%     93.70
                                     80 47.62%     88 52.38%
```

brilee:
I think this is a good idea - dynamic parent Q will reduce the impact of a noisy parent Q measurement,  lessening the likelihood of explore1/explore-all behavior. Also agreed with ducky's point about vloss interaction.

The right way to do this might be to use the vloss counter in the calculations, instead of mutating the W and N numbers.

(I.e. instead of Q = W / N where W and N have been mutated, don't mutate and instead do Q = (W - #vlosses) / (N + #vlosses)

sethtroisi:
One thing I don't like about my implementation.
node.child_Q[move] != node.child[move].Q()

Should I maybe move this definition under a new property name? dynamic_Q? dynamic_child_Q?

tommadams:
+1 for keeping child_Q as `child_W(i)) / (1 + child_N(i)` and adding a new property. I don't have a strong preference for either dynamic_Q or dynamic_child_Q.

amj:
what are we doing with this?  It seemed like a win in selfplay but was weaker vs Leela.  Do we add it for v10 or table it for future experimental runs?

Eddh:
weaker vs Leela ? That's weird. Did you do any experiment with fpu reduction btw ?

PR #352, opened 2018-07-30
Add a tf lite inference engine

tommadams:
Also update `cc/configure_tensorflow.sh` to build release v1.9.0


cc/README.md
```diff
@@ -98,6 +98,54 @@ fairly simple (there is no need for a LibertyTracker) and at the time of writing
...
 counterpart.
 
+## Inference engines
+
+C++ Minigo currently supports three separate engines for performing inference:
+
+ - tf: peforms inference using the TensorFlow libraries built by
+   `cc/configure_tensorflow.sh`.
+ - remote: launches a Python subprocess that performs inference using the
+   version of TensorFlow installed on the system. The remote inference
```
amj: maybe mention this this version of tensorflow found on the current PYTHONPATH ?

tommadams: I don't think it loads off PYTHONPATH, mine doesn't actually contain anything useful. I think it uses PATH

amj: but it will launch whatever `python` is active for the environment you launch minigo in, though, correct?  Like if your venv is activated, its in that venv python, etc.

tommadams: Yes. I've updated the doc to explicitly mention that

PR #354, opened 2018-07-31
Slow window

amj:
should be an easy one :)

### DIFF
--- a/example_buffer.py
+++ b/example_buffer.py
@@ -177,6 +177,9 @@ def _determine_chunk_to_make(write_dir):
 
     return chunk_to_make, False
 
+def get_window_size(chunk_num):
+    return min(500000, (chunk_num + 5) * 12000)
+
 def fill_and_wait_time(bufsize=dual_net.EXAMPLES_PER_GENERATION,
                   write_dir=None,
                   threads=32,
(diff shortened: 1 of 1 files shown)


example_buffer.py
```diff
@@ -177,6 +177,9 @@ def _determine_chunk_to_make(write_dir):
 
     return chunk_to_make, False
 
+def get_window_size(chunk_num):
+    return min(500000, (chunk_num + 5) * MINIMUM_NEW_GAMES)
```
brilee: I think this should scale sublinearly with chunk_number, since otherwise it's possible for the window size to just stay exactly at first game ~ latest game during the early phases.

So something like `min(500_000, (chunk_num + 5) * MINIMUM_NEW_GAMES * 0.5)`?

amj: ah but here's a trick, it already is sublinear :)  Games still get finished during the time it takes to write out the chunks, which is ~minutes.  we end up with about 1.5x-2x MINIMUM_NEW_GAMES per model.

brilee: Yeah, I understand that that's what's happening - but it'd be nice to have that implicit assumption be in the code, so that we don't have weird cascading effects due to "training happens to have gotten faster and now our pipeline isn't working for some reason"

sethtroisi: Consider documenting where this idea came from.

amj: Done.

amj: Done.

example_buffer.py
```diff
@@ -178,7 +179,12 @@ def _determine_chunk_to_make(write_dir):
     return chunk_to_make, False
 
 def get_window_size(chunk_num):
-    return min(500000, (chunk_num + 5) * MINIMUM_NEW_GAMES)
+    """ Adjust the window size by how far we are through a run.
+    At the start of the run, there's a benefit to 'expiring' the completely
+    random games a little sooner, and scaling up to the 500k game window
+    specified in the paper.
+    """
+    return min(500000, (chunk_num + 5) * (AVG_GAMES_PER_MODEL/2))
```
brilee: this shuold be integer division `//` to avoid breakage when AVG_GAMES_PER_MODEL is odd.

PR #355, opened 2018-07-31
Reload flags from a file every time we start a new game

tommadams:
- I've moved all the self play code into a new SelfPlayer class, which made it easier to coordinate all the threads safely reloading the flags each time a new game starts.
 - I've also removed the games_per_inference flag and the inference worker always uses ceil(1, parallel_games / 2) instead.


cc/dual_net/factory.cc
```diff
@@ -91,8 +88,9 @@ class RemoteDualNetFactory : public DualNetFactory {
       fputc('\n', stderr);
     });
 
+    int games_per_inference = std::max(1, parallel_games / 2);
```
sethtroisi: I'm commenting blindly and might misunderstand the purpose of this flag.
I found that having running a couple extra parallel_games worked much better (e.g. use parallel_games / 2 -2)
Otherwise you are waiting for the 100% percentile latency over X games V virtual losses (one game on one slow thread with one tricky position keeps the entire inference batch from starting)

cc/main.cc
```diff
@@ -282,36 +319,72 @@ std::thread SelfPlayThread(int thread_id, ThreadSafeRandom* rnd,
...
+      if (line.empty()) {
+        continue;
+      }
+      MG_CHECK(line.length() > 2 && line[0] == '-' && line[1] == '-') << line;
+      std::pair<std::string, std::string> flag_value =
+          absl::StrSplit(line, absl::MaxSplits('=', 1));
+      flag_value.first = flag_value.first.substr(2);
+      std::cerr << "Setting command line flag: --" << flag_value.first << "="
+                << flag_value.second << std::endl;
+      gflags::SetCommandLineOption(flag_value.first.c_str(),
```
sethtroisi: Not even sure if this is their API but Can you use ProcessOptionsFromStringLocked?
https://github.com/gflags/gflags/blob/23348465a02c9fdbdecce01cee24e62b9f1f5bad/src/gflags.cc#L968

sethtroisi: Or maybe even ProcessFlagfileLocked

cc/dual_net/factory.cc
tommadams: This isn't a problem in our case: when a game ends, we start a new one within a couple of seconds.

cc/main.cc
tommadams: Neither of those is part of the gflags public API.

sethtroisi: Cloud you use ParseCommandLineFlags and pass the array of lines from the file as argv?

There's also ReadFromFlagsFile but it's deprecated.

sethtroisi: Kinda hacky but you can just set argv = ["--flagfile", "<COPY_OF_GCS_FILE>"] and call ParseCommandLineFlags

Suggested by author here
https://github.com/gflags/gflags/issues/231#issuecomment-341087615

tommadams: ReadFromFlagsFile was apparently marked deprecated 12 years ago, so I doubt it's going away any time soon. Still, we should probably be good citizens and not use it.

I guess ParseCommandLineFlags would work, but I'm not sure it's much better than what I have.

cc/dual_net/factory.cc
amj: Does everything pause while we do?

if we play 32 games, a game takes 15m, this will happen ~twice a minute, so, it'll happen ~30 times in a game.

if the pause is just a few millis, no big deal though.

tommadams: Setting parallel_games > 2 * game_per_inference will also cause each game to take longer to finish though. I've not spent too much time tuning any of this, so we can most likely improve throughput with some experimentation in this area.

Ideally we'd enable these settings to be configured by the flag file and tune the numbers on GCE while running a few hundred selfplay jobs, so we can get real-world performance numbers. Sadly, that requires more work because we'd need to introduce a second mechanism to pass the config options over to the inference_worker in some coordinated fashion (so that both processes agree on the expected batch sizes). It wouldn't be too difficult, but I don't have time to work on that right now. Maybe in a couple of weeks.

tommadams: FWIW, I have no idea how long the selfplay jobs take to write the SGF and TF examples to GCS when running on GCE. On my local machine it's 1 or 2 seconds but it should be much faster for the VMs since they're colocated.

PR #358, opened 2018-08-01
[RFC] InMemoryEvaluatorHook

sethtroisi:


amj:
looks like an improvement, if it works?  Do you want to write up some more details on this?

sethtroisi:
It doesn't work on TPUs :( and it's broken at head (https://github.com/tensorflow/tensorflow/pull/20623/commits/36bf24bb1452b587448965d92fda066a25d22976)

PR #360, opened 2018-08-03
Update constants

amj:
Changes to constants:
- change max moves to 722

Changes to default flag values
- change resign threshold to -0.999
- change holdout percent from 5% to 3%

Changes to wrapper scripts.
- change resign threshold to -0.999
- change disable-resign amount from 20% to 10%

### DIFF
--- a/cluster/evaluator/evaluator_cc_wrapper.sh
+++ b/cluster/evaluator/evaluator_cc_wrapper.sh
@@ -30,10 +30,9 @@ echo Retrieiving Models
 gsutil cp ${MODEL_BLACK} .
 gsutil cp ${MODEL_WHITE} .
 
-#TODO use `basename` instead
-BASENAME_BLACK=`echo $MODEL_BLACK  | rev | cut -d/ -f1 | rev`
-BASENAME_WHITE=`echo $MODEL_WHITE  | rev | cut -d/ -f1 | rev`
-DATE=`date +%Y-%m-%d` 
+BASENAME_BLACK=`basename $MODEL_BLACK`
+BASENAME_WHITE=`basename $MODEL_WHITE`
+DATE=`date +%Y-%m-%d`
 
 bazel-bin/cc/main --mode=eval \
   --model=$BASENAME_BLACK \
--- a/cluster/selfplay/cc_wrapper.sh
+++ b/cluster/selfplay/cc_wrapper.sh
@@ -40,7 +40,7 @@ then
     --model=$NAME \
     --num_readouts=800 \
     --mode=selfplay \
-    --resign_threshold=0.99 \
+    --resign_threshold=0.999 \
     --disable_resign_pct=0.1 \
     --holdout_dir="gs://$BUCKET_NAME/data/holdout/$BASENAME" \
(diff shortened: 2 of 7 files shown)


cluster/evaluator/evaluator_cc_wrapper.sh
```diff
@@ -30,10 +30,9 @@ echo Retrieiving Models
 gsutil cp ${MODEL_BLACK} .
 gsutil cp ${MODEL_WHITE} .
 
-#TODO use `basename` instead
-BASENAME_BLACK=`echo $MODEL_BLACK  | rev | cut -d/ -f1 | rev`
-BASENAME_WHITE=`echo $MODEL_WHITE  | rev | cut -d/ -f1 | rev`
```
sethtroisi: so much nicer :)

cluster/selfplay/cc_wrapper.sh
```diff
@@ -40,8 +40,8 @@ then
     --model=$NAME \
     --num_readouts=800 \
     --mode=selfplay \
-    --resign_threshold=0.91 \
-    --disable_resign_pct=0.2 \
+    --resign_threshold=0.99 \
+    --disable_resign_pct=0.1 \
```
brilee: is this a percent or fraction?

cluster/selfplay/tpu_wrapper.sh
```diff
@@ -31,8 +31,8 @@ bazel-bin/cc/main \
   --games_per_inference=16 \
   --parallel_games=32 \
   --num_readouts=800 \
```
sethtroisi: did we discuss moving this to 1200?

cluster/selfplay/player_wrapper.sh
```diff
@@ -24,7 +24,7 @@ echo board_size: $BOARD_SIZE
 
 python3 rl_loop.py selfplay \
   --bucket_name=$BUCKET_NAME \
-  --resign-threshold=0.88 \
-  --num_readouts=900
+  --resign-threshold=0.99 \
```
sethtroisi: what are player_wrapper and cc_wrapper used for? and why does their value not agree with the TPU wrapper value (0.999)?

amj: player_wrapper is run by the python selfplay container and cc_wrapper is run by the cc selfplay container.

values harmonized.

cluster/selfplay/tpu_wrapper.sh
amj: we did, but i'm not sure yet that we'll have the extra quota for v10 (and not v11)

cluster/selfplay/cc_wrapper.sh
amj: it's a fraction o.o  And actually, i didn't check that these flags all still exist this way.

amj: actually this is true for all the flags. you want s/pct/amt?  something else?

brilee: s/pct/frac I think

PR #361, opened 2018-08-03
Don't divide steps by NUM_TPUS if passed from command line

sethtroisi:
steps is still divided if calculated by counting the records in the file.


dual_net.py
```diff
@@ -485,6 +483,8 @@ def count_examples(tf_record):
 
         total_examples = sum(map(count_examples, tf_records))
         steps = total_examples // FLAGS.train_batch_size
+        if FLAGS.use_tpu:
```
amj: Can you update the comment on train() with the behavior of steps w.r.t. use_tpu and num_tpu_cores?

sethtroisi: The comment now better matches the behavior

if you pass steps=<S> it will train for <S> steps (instead of <S>/num_tpu steps if use_tpu else <S>) 

if you don't pass steps it will determine steps as num_examples / true_batch_size 
(true_batch_size = batch_size * (num_tpus if use_tpu else 1) which is described in the batch_size flag.

PR #364, opened 2018-08-10
Unify soft_pick temperature cutoff code

sethtroisi:
both codebases now set temp_threshold to -1 if not using softpick.

both now do 0.98 squash (previously only c++ code did)


cc/mcts_player.cc
```diff
@@ -164,7 +165,7 @@ Coord MctsPlayer::SuggestMove() {
 }
 
 Coord MctsPlayer::PickMove() {
-  if (!options_.soft_pick || root_->position.n() >= temperature_cutoff_) {
```
amj: with >=, doesn't that mean 30 is the first argmax?   i.e., black has one extra softpick move?  or is n indexed from 0?

sethtroisi: I 99% know it's indexed from 0 in Python. I assume it is in C++ too.

tommadams: >= is correct, it was originally incorrectly >

See PR #213

amj:
LGTM ok2merge.

PR #365, opened 2018-08-10
Simplify most visited path code

sethtroisi:


mcts.py
```diff
@@ -278,29 +278,22 @@ def most_visited_path_nodes(self):
         return output
 
     def most_visited_path(self):
-        node = self
         output = []
-        while node.children:
-            next_kid = np.argmax(node.child_N)
-            node = node.children.get(next_kid)
-            if node is None:
```
sethtroisi: I don't think "GAME END" works correctly (you could search for an example in the logs and prove me wrong) but node.children would have to be non empty (so that the while enters) but not have a value for the node with the most visits.

amj: i didn't see it ever

amj:
LGTM

mcts.py
brilee: The reason is that node.children being nonempty is synonymous with node.child_n having at least 1 nonzero entry. So the `while node.children` and 

```
             node = node.children.get(next_kid)
             if node is None:
                 break
```
clauses are redundant. (You can remove them from the most_visited_path_node code)

brilee:
LGTM

PR #366, opened 2018-08-11
Improvements to several debug print statements

sethtroisi:


rl_runner.py
```diff
@@ -38,7 +38,7 @@ def loop(working_dir='estimator_working_dir', tpu_name=None):
         '--tpu_name', tpu_name
     ]
     while True:
-        print("==================================")
+        print("=" * 40)
```
sethtroisi: full disclosure this may add 1 or 2 ='s but 40 is a much rounder number

inference_worker.py
```diff
@@ -67,7 +70,7 @@
 NUM_WORKER_THREADS = 2
 
 # Print to stderr and flush by default.
-print = functools.partial(print, file=sys.stderr, flush=True)
+print = dbg
```
amj: can we just s/print/dbg here then?  i'd rather avoid a top-level assign here.

local_rl_loop.py
```diff
@@ -48,20 +48,23 @@ def rl_loop():
     flags.FLAGS.num_readouts = 10
 
     with tempfile.TemporaryDirectory() as base_dir:
+        bootstrap_model = '000000-bootstrap'
+
         flags.FLAGS.base_dir = base_dir
         working_dir = os.path.join(base_dir, 'models_in_training')
         flags.FLAGS.model_dir = working_dir
-        model_save_path = os.path.join(base_dir, 'models', '000000-bootstrap')
+        model_save_path = os.path.join(base_dir, 'models', bootstrap_model)
```
amj: maybe a TODO but i think this should be able to be refactored with fsdb?

Issue #363, opened 2018-08-08
Moves considered logging is broken at times


sethtroisi:
![screenshot from 2018-08-08 02-19-31](https://user-images.githubusercontent.com/10172976/43828537-abbce5f8-9ab1-11e8-9108-894304101754.png)

tommadams:
This is the same issue that PR #328 addresses.

The move list is sorted by (N, action_score, move) while the first move in the principle variation is chosen as by the first move with the highest N, effectively sorting by (N, move).

I can fix this, or if you want to add the fix to #328 you're more than welcome. This is the line that needs fixing: https://github.com/tensorflow/minigo/blob/2dbc02645b1ce4e1f0bd6ad0049627f26fb119cb/cc/mcts_player.cc#L169

Add a call to CalculateChildActionScores, then I guess you'd have to iterate over `root_->edges` manually instead of calling ArgMax since you need to index into two separate arrays.

sethtroisi:
I found the problem.

We describe twice. One when verbose >= 2, one on stderr
three guesses which one is more useful.

amj:
This is not quite accurate.  When running the python engine via gtp.py, we always install the minigui gtp handlers, even when we are not launching the engine via minigui/serve.py

The minigui genmove handler calls `describe` on the board *after* the move has been played (so the scraper picks it up).  This results in the list of moves being the ones for the child move (i.e., only the re-used subtree).

tommadams:
I'm going to fix the problem about move selection I mentioned in my [earlier comment](https://github.com/tensorflow/minigo/issues/363#issuecomment-411536510) over in issue #260

PR #376, opened 2018-08-11
Only load minigui cmd handle in minigui mode

sethtroisi:
Fixes part of  #363


gtp.py
```diff
@@ -63,7 +64,10 @@ def make_gtp_instance(load_file, cgos_mode=False, kgs_mode=False, verbosity=1,
         engine.add_cmd_handler(KgsCmdHandler(player))
     engine.add_cmd_handler(RegressionsCmdHandler(player))
     engine.add_cmd_handler(GoGuiCmdHandler(player))
-    engine.add_cmd_handler(MiniguiCmdHandler(player, courtesy_pass=kgs_mode))
+    if minigui_mode:
```
brilee: Can you rename MiniguiCmdHandler -> MiniguiBasicCmdHandler? It seems fishy to me that it's possible not to include a "basic" command handler, but it took me some digging to realize that MiniguiCmdHandler subclassed the basic handler, rather than adding new commands.

brilee:
lgtm

PR #377, opened 2018-08-11
Upgrade dbg print wrapper

sethtroisi:
dbg doesn't work if passed none string argument I assumed it did in #366, I believe this fixes it


utils.py
```diff
@@ -24,9 +24,9 @@
 import time
 
 
-def dbg(fmt, *args):
+def dbg(*args, **kwargs):
```
brilee: I'm not sure I understand the design of this wrapper. Maybe have impl be `print(*args, **kwargs, file=sys.stderr, flush=True)`? Then the api is same as print function.

sethtroisi: That's quite clever let me change it to that :)

PR #382, opened 2018-08-15
Fix MCTS PUCT formula

### LINKED ISSUE #379: Problem with MCTS
the PUCT formula (From AlphaGo Zero pg 26) is `cpuct * policy[move] * sqrt(sum(child_N)) / (1 +child_N[move])`

Our current implementation ([mcts.py#L106](https://github.com/tensorflow/minigo/blob/master/mcts.py#L106)) uses `sqrt(1 + self.N)`

Problems:
1. `+ 1` not mentioned in AlphaGo
1. `self.N != sum(child_N)`, this cut shortcut is generally off by 1 (as the initial expansion of the root increments self.N but not any of child_N)
1. Minigo [increments edge count](https://github.com/tensorflow/minigo/blob/master/mcts.py#L138) prior to select instead of during backup, so that `self.N = sum(child_N) + 2` when action_scores are calculated.

All of this results in `sqrt(sum(child_N) + 3)` instead of `sqrt(sum(child_N))`, which is clearly wrong.
To muddy this conversation, `sqrt(sum(child_N) +1)` seems to make more sense for because otherwise when there's no visits `sqrt(sum(child_N))` equals 0 and argmax will need a tiebreaker.

sethtroisi:
Fixes #379


sethtroisi:
C++ code will also need to be fixed but that can be a separate PR

mcts.py
```diff
@@ -147,7 +149,10 @@ def select_leaf(self):
                 current = current.maybe_add_child(pass_move)
                 continue
 
+            # NOTE(#379): child_action_score depends on current.N
```
brilee: I don't see why we need this - why not just remove the initial current.N += 1 at the top of the while loop?

mcts.py
```diff
@@ -103,7 +103,9 @@ def child_Q(self):
 
     @property
     def child_U(self):
-        return (FLAGS.c_puct * math.sqrt(1 + self.N) *
+        # NOTE(#379): Expanding self counts as 1, self.N - 1 = sum(child_N).
```
brilee: I think you can revert this change if you don't increment current.N at the top of the while loop?

tests/test_mcts.py
```diff
@@ -177,6 +221,9 @@ def test_never_select_illegal_moves(self):
             self.assertNotEqual(leaf.fmove, 1)
 
     def test_dont_pick_unexpanded_child(self):
+        # CPUCT must be large enough that it can overcome Q = -1/2
```
brilee: I don't understand this change - was this test failing after your changes?

tests/test_mcts.py
```diff
@@ -58,6 +63,45 @@
 
 
 class TestMctsNodes(test_utils.MiniGoUnitTest):
+    def test_puct(self):
```
brilee: can you give this test a more descriptive name? Not sure what it's doing, other than a general integration-ish test.

tests/test_mcts.py
sethtroisi: Yes, previously puct was magnified (`sqrt(N + 3)` is much larger than `sqrt(N)` for small Ns) and now this test fails.

tests/test_mcts.py
sethtroisi: I updated the name.

mcts.py
sethtroisi: I disagree,

The paper calls for sqrt(sum(child_n))
* the +1 is not called for
* self.N != sum(child_n), it's sum(child_n) + 1
* the while loop increments N out of order previously resulting in an additional +1 (now handled in that loop)

brilee: After meditating on it, I think it's hard to argue in favor of either `sqrt(self.N)` or `sqrt(max(1, self.N-1))`. I'd prefer the former for simplicity's sake. AGZ paper doesn't really say one way or another.

mcts.py
amj: that doesn't work when the node hasn't been expanded, though, correct?  We don't detect that we picked it if we never incremented N.

brilee: `node.is_expanded` doesn't refer to `node.N`, so I'm not sure what you mean.

mcts.py
amj: the paper is pretty clear that the thing in the sqrt is the sum of the visits of the siblings (i.e., parent.N - 1)

mcts.py
amj: it means N is not incremented in the situation where virtual losses are not sufficient to prevent the same node from being selected.

brilee:
I don’t understand what this comment is trying to say

sethtroisi:
@brilee We increment N for all edges in the path from root to leaf during "Select (figure 2a)" instead of during backup. So removing the + 1 would stop all nodes from being incremented.

That being said Andrew is trying to move the visit increment to incorporate_results which is what the paper says it does.

amj:
@brilee: you're asking to move the line that increments N from the top of the loop.  I'm pointing out that that means there's a path out of the loop that does not increment N and there are side effects to that

amj:
(LGTM from me btw, but would like to have @brilee's approval plz before merging)

brilee:
@amj ahhhh... that is indeed a problem. Oh man, that's a rather distasteful quantity of current.N manipulation.

I think the right fix is:
- in `select_leaf`, have no mutations whatsoever
- during `add_virtual_loss`, increment `losses_applied`, don't mutate N / W.
- in Q calculations, replace `W/(1 + N)` with `(W-self.position.to_play * self.losses_applied)/(1 + N + self.losses_applied).
- during `revert_virtual_loss`, decrement `losses_applied`, don't mutate N / W
- `revert_visits` can be deleted
- in `backup_value`, add to W and increment N.

I believe this satisfies all of our desiderata?

sethtroisi:
@brilee I want to keep this PR as short, sweet, and simple as possible. Currently it's 3 lines of code. 

I'm totally in agreement with the larger refactor that andrew has planned but let's do that in a separate PR and get this for the short term.

brilee:
Sure-  let's get this merged in. I can make the changes I described on top of this change.

tommadams:
I don't really understand the urgency of committing this PR. Why not just wait and fix the underlying cause of the bug rather than just the symptom?

Andrew and I already have changes to the C++ and Python code that move the update of N into `backup_value` they just require a little more tweaking to match what Brian described and should be ready to go today.

amj:
re: @brilee i think we can clean things up a bit further by having SelectLeaf() apply the virtual losses, and IncorporateResults() removes them.

that means all of the vloss logic is contained inside the mcts node object and none of it.

i haven't completely thought it through but it seems reasonable.

sethtroisi:
@tommadams I'm trying out that small, separate CL advice. I just want child_U to be calculated with sqrt(n) , it seems the change you and andrew are working on is a larger (and separate) change cleaning up our vloss implementation.

amj:
Fixed by #392 ?

PR #385, opened 2018-08-15
Change all unittest.assert() to be assert(expected, actual)

sethtroisi:
https://bugs.python.org/issue10573


tests/test_features.py
```diff
@@ -92,60 +92,61 @@ def test_stone_features(self):
...
+        self.assertEqual(1, f[1, 0, 7])
 
     def test_recent_moves_feature(self):
         f = features.recent_move_feature(TEST_POSITION)
         self.assertEqual(f.shape, (9, 9, features.recent_move_feature.planes))
         # most recent move at (1, 0)
-        self.assertEqual(f[1, 0, 0], 1)
-        self.assertEqual(f[1, 0, 3], 0)
+        self.assertEqual(1, f[1, 0, 0])
+        self.assertEqual(0, f[1, 0, 3])                             # < WHAT IS THIS TESTING?
```
amj: the history planes?

tests/test_features.py
```diff
@@ -92,60 +92,61 @@ def test_stone_features(self):
...
         # second most recent move at (0, 8)
-        self.assertEqual(f[0, 8, 1], 1)
-        self.assertEqual(f[0, 8, 0], 0)
+        self.assertEqual(1, f[0, 8, 1])                             # < WHAT IS THIS TESTING?
+        self.assertEqual(0, f[0, 8, 0])                             # < WHAT ORDER ARE THE PLANES IN?
```
amj: These are the features for the paper #1 features -- namely, onehots for the most recent N moves.

Since it doesn't look like we'll be using or training with the old paper #1 code, we can kill this, right @brilee ?

tests/test_features.py
amj: testing that the most recent move played doesn't exist on history planes before it's been played.

tests/test_features.py
```diff
@@ -92,60 +92,61 @@ def test_stone_features(self):
...
+        self.assertEqual(0, f[0, 1, 1])
         # plane 1 is W
-        self.assertEqual(f[0, 8, 1], 1)
-        self.assertEqual(f[0, 8, 0], 0)
+        self.assertEqual(0, f[0, 8, 0])
+        self.assertEqual(1, f[0, 8, 1])
         # plane 2 is empty
-        self.assertEqual(f[0, 5, 2], 1)
-        self.assertEqual(f[0, 5, 1], 0)
+        self.assertEqual(1, f[0, 5, 2])                                 # < WHAAAAAAATTTT
```
brilee: it's the "empty" plane - i.e. 1-hot if that spot is neither black nor white. AGZ got rid of the empty planes since in theory it's just 1 - B - W.

tests/test_features.py
brilee: i secretly think that AGZ would be stronger with some of the AG feature planes, so I'd like to keep them.

amj:
could we remove the QUESTIONS UNDER DISCUSSION IN ALL CAPITAL COMMENTS before merging tho plz? :pray:

tests/test_features.py
sethtroisi: My understanding is that ordering of the planes is?
```
012345
BWBWBW
```
So you're checking that 1,0 is B most recently, and not W two turns ago?

amj: yeah it looks like it should be s/3/2 to me

tests/test_features.py
sethtroisi: That makes sense.
I failed to differentiate "stone_color_features" and "stone_features" and thought this part of NEW_FEATURES.

tests/test_features.py
amj: hasn't changed since our fork:
https://github.com/brilee/MuGo/blob/master/tests/test_features.py#L77

sethtroisi: Nope, I'm just wrong about how these features are laid out. removing comments

PR #386, opened 2018-08-15
Fix CC MCTS PUCB formula

sethtroisi:
Fixes CC side of #379 

In python I had to fix an additional test, it's weird that it doesn't break here too but I'll leave that investigating to another dev.


cc/mcts_node_test.cc
```diff
@@ -237,7 +292,7 @@ TEST(MctsNodeTest, DontPickUnexpandedChild) {
   leaf1->AddVirtualLoss(&root);
 
   auto* leaf2 = root.SelectLeaf();
-  EXPECT_EQ(leaf1, leaf2);
+  EXPECT_EQ(leaf2, leaf2);
```
amj: lol.  why not leaf1?

cc/mcts_node_test.cc
```diff
@@ -36,6 +38,59 @@ static constexpr char kAlmostDoneBoard[] = R"(
     XXXXXOOOO
     XXXXOOOOO)";
 
+// Test puct and child action score calculation
+TEST(MctsNodeTest, Puct) {
```
amj: very legible test, thank you.

amj:
fixed by #392  ?

PR #389, opened 2018-08-16
test_strategies.py work

sethtroisi:
some refactoring I did while trying to understand these tests.


tests/test_strategies.py
```diff
@@ -281,8 +281,7 @@ def test_only_check_game_end_once(self):
         self.assertEqual(player.root.child_N[pass_move], 1)
 
     def test_extract_data_normal_end(self):
-        player = MCTSPlayer(DummyNet())
-        player.initialize_game()
+        player = initialize_basic_player()
         player.tree_search()
```
amj: re: above, this would have searched twice?

tests/test_strategies.py
```diff
@@ -76,9 +76,9 @@ def run_many(self, positions):
         return [self.fake_priors] * len(positions), [self.fake_value] * len(positions)
 
 
-def initialize_basic_player():
+def initialize_basic_player(position=None):
```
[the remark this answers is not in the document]
amj: (i guess i'm saying i think the behavior should be
```
if position is not None:
    [do search]
```
)

tests/test_strategies.py
```diff
@@ -237,7 +237,10 @@ def test_cold_start_parallel_tree_search(self):
         player.initialize_game()
         self.assertEqual(0, player.root.N)
         self.assertFalse(player.root.is_expanded)
-        player.tree_search(parallel_readouts=4)
+        leaves = player.tree_search(parallel_readouts=4)
+        self.assertEqual(4, len(leaves))
+        self.assertEqual(player.root, leaves[0])
```
brilee: actually all 4 elements of leaves will be the root  - could verify that too :D

tests/test_strategies.py
brilee: yeah, I think this isn't a functionally equivalent substitution.

tests/test_strategies.py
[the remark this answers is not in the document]
sethtroisi: You do the search so that you have priors and a subsequent tree_search actually does something interesting (instead of just returning the root node x 8).

tests/test_strategies.py
sethtroisi: I removed self.assertEqual(1, len(set(leaves))) because It was getting a little silly.

tests/test_strategies.py
sethtroisi: This test doesn't really need to invoke search, and invoking it twice isn't a bad thing.
it's about making sure that is_done() get's set when you call play_move(None) twice.

tests/test_strategies.py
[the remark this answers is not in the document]
amj: I agree that makes sense, and we don't seem to manually push in priors.  maybe just call it init_player_with_search? or something?

PR #396, opened 2018-08-20
Fixes for num_results=1

### LINKED ISSUE #260: num_readouts=1 seems broken

tommadams:
Fixes C++ implementation for issue #260 .

- Change `MdsPlayer::PickMove` and `MctsNode::MostVisitedPath` to break visit count ties using action score.
- Always expand the root node if it has no visit counts: when the root has 0 visits, the first readout will expand the children but not actually visit any of them, so child_N(i) will be 0 for all i. By expanding the root if N == 0, SelectLeaf will actually expand the first child. Note that during self play, this initial expansion of the root only happens once.
- Also fixes `MctsNode::PickMove` to squash probabilities before performing soft pick.


cc/mcts_player.cc
```diff
@@ -179,18 +179,19 @@ Coord MctsPlayer::PickMove() {
...
-  float norm = 1 / cdf[cdf.size() - 1];
+  // For moves before the temperature cutoff, exponentiate the probabilities by
+  // a temperature slightly larger than unity to encourage diversity in early
+  // play and hopefully to move away from 3-3s.
   for (size_t i = 0; i < cdf.size(); ++i) {
-    cdf[i] *= norm;
+    cdf[i] = std::pow(root_->child_N(i), kVisitCountSquash);
+  }
+  for (size_t i = 1; i < cdf.size(); ++i) {
+    cdf[i] += cdf[i - 1];
```
sethtroisi: this is much nicer :)

gtp_cmd_handlers.py
```diff
@@ -277,7 +277,6 @@ def cmd_genmove(self, color=None):
 
         root = self._player.get_root()
         dbg("")
-        dbg(root.describe())
```
sethtroisi: I thought this was required for minigui

tommadams: No, Minigui only actually parses lines that are printed to stderr and start with the prefix `mg-`, everything else is just echoed to it's console. We added the `root.describe()` call just for something interesting to show in the console, it's not needed at all.

PR #403, opened 2018-08-29
Changes to support quantized inference & training

tommadams:
- Change TFLite inference engine to expect a quantized model.
- Add a `quantize` flag to dual_net.py that optionally quantizes the graph.
- Add a `add_quant_ops` to main.py that will add quantize ops to a partially trained model. I haven't actually managed to get this to work properly yet but I'm adding it for reference.


main.py
```diff
@@ -136,6 +136,36 @@ def convert(load_file, dest_file):
     tf.reset_default_graph()
 
 
+def add_quant_ops(load_file, dest_file):
```
amj: i'm a little leery of committing code that's dead-on-arrival; could we mark this with some comments about what's broken, what's expected, etc?

dual_net.py
```diff
@@ -78,6 +78,12 @@
 
 flags.DEFINE_bool('use_tpu', False, 'Whether to use TPU for training.')
 
+flags.DEFINE_bool('quantize', False, 'Whether to quantize the model.')
```
amj: "whether to create a quantized model" ?

(what happens if we set this flag when we load an unquantized model?)

tommadams: Presumably something bad

main.py
tommadams: hopes... deleted

PR #404, opened 2018-08-30
Better sampling

brilee:
@amj i haven't actually run this code.

### DIFF
--- a/example_buffer.py
+++ b/example_buffer.py
@@ -28,11 +28,11 @@ MINIMUM_NEW_GAMES = 12000
 AVG_GAMES_PER_MODEL = 20000
 
 
-def pick_examples_from_tfrecord(filename, samples_per_game=4):
+def pick_examples_from_tfrecord(filename):
     protos = list(tf.python_io.tf_record_iterator(filename, READ_OPTS))
-    if len(protos) < 50:  # Filter games with less than 20 moves
-        return []
-    choices = random.sample(protos, min(len(protos), samples_per_game))
+
+    number_samples = np.random.poisson(len(protos) / 50.0)
+    choices = random.sample(protos, min(len(protos), number_samples))
 
     def make_example(protostring):
         e = tf.train.Example()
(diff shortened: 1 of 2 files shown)


example_buffer.py
```diff
@@ -28,21 +29,16 @@
 AVG_GAMES_PER_MODEL = 20000
 
 
-def pick_examples_from_tfrecord(filename, samples_per_game=4):
+def pick_examples_from_tfrecord(filename, sampling_frac=0.02):
```
amj: maybe extract the .02 to a default at the top?  (where does it come from, btw?)

example_buffer.py
```diff
@@ -56,19 +52,18 @@ def _ts_to_str(timestamp):
...
+        self.sampling_frac = sampling_frac
+        self.func = functools.partial(choose, sampling_frac=sampling_frac)
         self.total_updates = 0
 
     def parallel_fill(self, games, threads=8):
         """ games is a list of .tfrecord.zz game records. """
         games.sort(key=os.path.basename)
         # A couple extra in case parsing fails
-        max_games = (self.max_size // self.samples_per_game) + 480
+        max_games = (self.max_size / self.sampling_frac / 200) + 480
```
amj: 200 = avg moves?

brilee:
Yeah, 0.02 comes from 4 / 200 being the assumption we were making previously in this code. I think it should be factored out more explicitly in terms of the hyperparameters of the run (what is trailing window size, etc.)

Speaking of which, what happened to our slow start thing - I forget if we wrote code for that?

amj:
Got it.  "Slow start" is right here: https://github.com/tensorflow/minigo/blob/master/example_buffer.py#L184

it's been in the v10 run :)

brilee:
Ah, I see. But you can't change the window size without also changing the sampling ratio, because then your chunks won't fill up completely? Or something like that.

*shrug*, I guess it's working so I'm not going to look too closely at this particular file.

brilee:
I also did run the integration tests and fixed some stuff, so it should be good.

amj:
i'm really not sure about this change.  In particular, i think the assumption of "50 generations in the window" may be pretty far off.  We can try it on v11 but i have some real reservations about how this will work...

brilee:
The assumption being made here is identical to the assumption that was
previously being made (4 moves from a game that is 200 long on average)

amj:
Not quite:  4 moves per game * a window of 500k games = the 2M samples required for 1k steps per checkpoint.

PR #405, opened 2018-08-30
Move batching out of remote inference and makes it available for all …

chsigg:
…backends.

Move leaves and output around as vectors instead of pre-allocating it in the MctsPlayer and reference it.

Fix minor warnings.


chsigg:
Hello MiniGo team. I would like to propose this PR to reduce the diff to my local changes (supporting playing multiple games per thread etc, which makes the main.cc code a little more involved).

I think this change is a net benefit because it makes the batching more deterministic. 
It also deallocates features as early as possible and allocates the output as late as possible.

cc/dual_net/dual_net.h
```diff
@@ -52,6 +56,118 @@ class DualNet {
...
+    std::array<float, kNumMoves> policy;
+    float value;
+  };
+
+  struct Result {
+    std::vector<Output> outputs;
+    std::string model;
+  };
+
+  class Service {
```
tommadams: Can you explain why the Service and Client classes are needed? This code structure has pushed a bunch of complexity down into the client classes, which I'd like to avoid. It seems to me that IncrementClientCount, DecrementClientCount and FlushClient are implementation details of batching and shouldn't be part of any public API.

I think you might have made life a little difficult for yourself by eliminating the factory classes: having the factory classes was useful because that was a place where one could put shared state.

When I was originally thinking about cleaning up the batching code, I imagined that there would be a BatchingFactory that wrapped a TfDualNetFactory (or whatever). The BatchingFactory::New would then simply return a client of the batcher that implemented the existing DualNet interface and the client refcounting and flushing would be implementation details of that DualNet subclass. That way none of the calling code would have to change at all: you'd just implement a batcher, batching factory and add the relevant code to factory.cc.

tommadams:
This looks like a good start, thanks for taking this on!

Just a few high-level comments to begin with.

This PR contains multiple different changes that are mixed together, which makes it hard to review:

- Cleaned up batching code. Great!
- Changed Span to vector. I like this too!
- DualNet::RunMany is now async.
- TfDualNet has been substantially changed.
- The DualNet implementations have been moved around, which makes understanding the other changes harder.
- The continuation passing style changes to DualNet and the functors. I'm not clear if that's because of the async changes or the batching. It's hard to follow.

Are you able to separate out _just_ the batching and maybe span/vector changes please?

cc/dual_net/batching_service.cc
```diff
@@ -0,0 +1,203 @@
...
+        continue;
+      }
+
+      auto inference = std::move(inference_queue_.front());
+      inference_queue_.pop();
+      size_t num_features = inference.features.size();
+      DualNet::Functor functor(std::move(inference.features),
+                               std::move(inference.promise));
+
+      // If inference doesn't fit into the batch anymore, move it to shared
```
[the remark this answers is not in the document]
chsigg: Aren't there inferences with <virtual_losses at the end of each game as well? I think that's what I've seen.

The complexity of the SharedFunctor is quite moderate, and it simplfies the MaybeRunBatches() code.

cc/dual_net/dual_net.h
chsigg: The asynchronous playing of multiple games from a single thread requires some this complexity (in particular, that the client is shared between multiple players).

But I fully agree that I did not do a good job at hiding this complexity when it's not really needed. I hope the latest commit improved on it. 

The client count and flushing is still exposed in the service, but now transparently handled by the client. I think going all the way with an abstract client and an abstract factory isn't really warranted, but definitely possible.

tommadams: Playing multiple games from a single thread or multiple threads seems orthogonal to batching to me. In both cases, it seems like each player could continue to just have their own DualNet instance: a lightweight class that pushes requests to the batcher and waits for them to complete, much like how the previous InferenceServer worked. I don't understand why you need Client and Service in addition to DualNet.

Unfortunately, I'm very bad at trying to explain ideas that are in my head, so I resurrected my original attempt at extracting the batcher last night: https://github.com/tommadams/minigo/commit/291abff7061e88ab1039d9d080a4b42db834f9c7
It's just a sketch really but it seems to work and doesn't require the extra Client and Service classes.

I asked this before offline, but do you have benchmarks to demonstrate that playing multiple games from a single thread is significantly faster than just launching one thread per game? The setup you're targeting (a single process running on eight V100s, as I understand) is quite far removed from what we and hobbyist Minigo users run on, I'd like to understand whether the performance wins justify the non-trivial complexity added with async inference.

Even running your master branch code with the number of threads equal to the number of parallel games would be sufficient. I tried this on my desktop machine and didn't see any significant performance difference for 512 parallel games, but then I only have a fairly modest GPU.

cc/dual_net/batching_service.cc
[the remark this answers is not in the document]
tommadams: Apart from the very start of training where the games basically random and the board tends to fill up, the engine will basically always be able to make virtual_losses reads. Even for the resign disabled games, the engine will pass rather than fill in its own eyes once it has learnt the basics of the game.

My primary concern in that of readability: Minigo go is built by and for hobbyists to play with. Prior to this PR, my opinion is that it was relatively easy to understand how to add a new inference backend: subclass DualNet, implement a single method, and create a factory. Now, someone wishing to accomplish the same has to understand the DualNet, Service, Client, Continuation, Functor classes. I'd like to understand which parts of the PR are strictly necessary and what can be simplified.

tommadams:
Morning! I haven't had chance to look at the latest changes but I wanted to add some comments before breakfast :)

cc/dual_net/dual_net.h
tommadams: While out walking my dog, it occurred to me that this comment could be misinterpreted: I didn't mean to suggest that the batching code had to be structured or written a certain way, just that I don't believe that the Service and Client classes are strictly necessary.

This code will probably be read orders of magnitudes more times than it is written, so I'd like to keep it as simple as possible.

chsigg: Thanks Tom for the comments. I ran some experiments on my totally butchered repo. With a fake net, I got 38 sec on 16 threads vs 49 on 512 threads. So there is some improvement but it might not be relevant in the grand scheme. Let's put this aside for now, all code is back to synchronous execution in my latest commit.

Regarding the client count: it might be easier to understand why we need it if you consider the case of playing multiple eval games. Each game has two clients, one for each model. If each dual net waits for all clients to request an inference batch, they will deadlock each other.

With my latest commit. The service is now what the factory was in the original implementation. I didn't really see the value of the extra factory class when the service interface is very small. But you might want to convince me to rename the classes anyway ;-)
The client is an ABC with the client count being a private implementation detail of the batching client.

cc/dual_net/dual_net.h
```diff
@@ -51,6 +55,42 @@ class DualNet {
...
+
+  class Service;
+
+  // Base class to perform inferences on a single batch of features.
+  class Client {
+   public:
+    virtual ~Client();
+
+    // Runs inference on a batch of input features. Thread-safe.
+    virtual Result RunMany(std::vector<BoardFeatures>&& features) = 0;
```
tommadams: Thanks for the refactor, this helped me understand the code structure better. I still don't think the Client class is required and batching would be better implemented as a subclass of DualNet. As far as I can tell DualNet::Service is basically the same as DualNetFactory (if batching were implemented as a DualNetFactory), so I don't think that is required either.

Both DualNet::RunMany and DualNet::Client::RunMany have very similar interfaces: they both operate on batches of games, the difference is just that the DualNet::RunMany takes flattened inputs. The DualNet implementations are going to have to flatten those games into a single inference batch anyway.

The way I see it, batching should be purely an implementation detail of a simple DualNet subclass that wraps a concrete DualNet (e.g. TfDualNet). In the example I gave in https://github.com/tommadams/minigo/commit/291abff7061e88ab1039d9d080a4b42db834f9c7, the BatchingDualNet wrapped the concrete DualNet implementation.

I think this is close to what you have already and we can probably get there with a little more iteration.

tommadams: One of the reasons I'm pushing back quite strongly against adding complexity to this part of the codebase is that we're in the process of porting the C++ code to Raspberry Pi and we hope that will encourage people who are interested in Go and ML to play around with the code. So we're placing a higher than usual emphasis on code simplicity, especially around the inference code where we'd like people to be able to easily drop in a new implementation.

(The existing code isn't as simple as I'd like FWIW)

chsigg:
Abandon this PR, please see PR #446 for the next attempt.

chsigg:
PR #465 is the much much cleaner version of this one. Closing.

PR #410, opened 2018-09-04
Make all dependencies on TensorFlow optional

tommadams:
Move the file IO functions from `cc/tf_utils.cc` into `cc/file/utils.cc` and use stdio implementations if Minigo is compiled without a TensorFlow inference engine, i.e. `--define=tf=0`.


cc/config/BUILD
```diff
@@ -13,3 +13,23 @@ config_setting(
     name = "minigo9",
     define_values = {"board_size": "9"},
 )
+
+# Build condition labels that configure which inference engines are enabled.
+# Additionally, enable_tf is also required in order for the following
+# functionality, which is provided by TensorFlow:
+#  - Google Cloud Storage access.
+#  - Writing TensorFlow examples at the end of self play.
```
amj: i laughed.

Maybe mention that it's not needed to play minigo over GTP?

PR #413, opened 2018-09-05
Remove petname dependency.

briandersn:


brilee:
We'd previously had this dep because it was nice to have a different namelist for 9x9 models vs. 19x19 models. But at this point we have multiple runs and we need to disambiguate v9 (19x19) from v10 (19x19) from v7 (19x19) from v3 (9x9) so I'm okay with dumping the petnames dep.

@amj wdyt?

brilee:
from the lint errors - I0906 14:57:08.027] W: 23, 0: Unused import go (unused-import)

amj:
i'm fine with it :)  lgtm with the lints fixed

briandersn:
Thanks for the quick turn around. Fixed the lint error and it looks like all checks are passing now.

I'm not that familiar with the github merge process: is there any further action required on my part?

amj:
Nope, we got it covered :)  Tom will have to do something to pull it back into g3 though i think

tommadams:
I'll do that now.

PR #414, opened 2018-09-07
Fix first node bootstrap issue

### LINKED ISSUE #393: Tree search should always expand root if it has 0 visits
This is a follow on form issue #260.

At the start of selfplay, the Python code runs:

```
     first_node = player.root.select_leaf()
     prob, val = network.run(first_node.position)
     first_node.incorporate_results(prob, val, first_node)
```

With a comment that this is required for noise injection to work correctly. Actually, this `select_leaf`, `run`, `incorporate_results` sequence should be performed at the start of tree search _every time_ we find that the root node as 0 visits. This happens at the start of the game, and any time the move prior to tree search wasn't itself decided by a tree search (e.g. pondering after loading an SGF, in GTP mode after the opponent's turn).

For an explanation of why this is necessary, consider the `num_readouts=1` case. When the root has 0 visits, the first readout will expand the children but not actually visit any of them, so `child_N(i)` will be 0 for all `i`. Choosing the next leaf will then have to rely on a tie breaker, which is still missing from the Python code (I'm finishing up a C++ implementation now).

The [C++ code](https://github.com/tensorflow/minigo/blob/102f8d39d6eecda8453667dcad23d5b9eb5bf9e9/cc/mcts_player.cc#L123) is almost correct but the check for root node having 0 visits is behind an `inject_noise` enabled check. I'll fix that in my next PR.

brilee:
Fixes #393


brilee:
nm, this doesn't catually fix the problem

PR #415, opened 2018-09-07
Scale center weights of conv2d

sethtroisi:


sethtroisi:
image

amj:
@sethtroisi what is the context for these images?

sethtroisi:
I wanted a record of using tf.summary.image, it was super easy and looked okay.

sethtroisi:
The larger context is they are the conv2d kernels from the center_conv experiment

PR #417, opened 2018-09-07
Add conv width and minigui flags

tommadams:


minigui/serve.py
```diff
@@ -44,9 +44,14 @@
     "--engine",
     default="py",
     type=str,
-    help="Which Minigo engine to use: \"py\" for the Python implementation, "
-         "\"cc\" for the C++ implementation.")
+    help="Which Minigo engine to use: \"py\" for the Python engine, or "
```
amj: Is it worth putting a validator on this?

tommadams: Oof, this help text was pretty wrong.

The problem is that the supported list of engines changes depending on how the C++ binary was compiled. I've changed text to describe how to find out the actually supported list.

PR #418, opened 2018-09-08
Make use_random_rotation (for inference) a flag

sethtroisi:


brilee:
LGTM. I'm surprised we don't have any unit tests that require us to toggle between use_random_symmetries=True/False.

sethtroisi:
I think we only have 3 unit tests that run real inference.

PR #419, opened 2018-09-08
More refactoring

brilee:
@amj
Separate out train.py; refactor --model_dir vs --working_dir; combine export_model/freeze_graph

### DIFF
--- a/dual_net.py
+++ b/dual_net.py
@@ -29,7 +29,6 @@ from tqdm import tqdm
 import numpy as np
 import tensorflow as tf
 from tensorflow.contrib import summary
-from tensorflow.python.training.summary_io import SummaryWriterCache
 from tensorflow.contrib.tpu.python.tpu import tpu_config
 from tensorflow.contrib.tpu.python.tpu import tpu_estimator
 from tensorflow.contrib.tpu.python.tpu import tpu_optimizer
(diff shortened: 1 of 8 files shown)


dual_net.py
```diff
@@ -470,71 +464,31 @@ def bootstrap(working_dir):
...
     latest_checkpoint = estimator.latest_checkpoint()
     all_checkpoint_files = tf.gfile.Glob(latest_checkpoint + '*')
     for filename in all_checkpoint_files:
         suffix = filename.partition(latest_checkpoint)[2]
         destination_path = model_path + suffix
         print("Copying {} to {}".format(filename, destination_path))
         tf.gfile.Copy(filename, destination_path)
+    # also export a .pb for C++ inference
+    n = DualNetwork(latest_checkpoint)
+    out_graph = tf.graph_util.convert_variables_to_constants(
```
amj: Have you tested this with `use_tpu` set to true?  The previous version of freeze did not run on TPUs.

brilee: blargh -.-

I combined the calls because all usage locations of dual_net.export_model were accompanied by main.freeze_graph

I think what you're saying implies that we should have a guard at top saying `if FLAGS.use_tpu: raise Exception("exporting doesn't work on a tpu machine")` or something along those lines.

brilee:
Oh, also - I don't intend to merge any of the refactor PRs until it's all done. I've split out the multiple changes into different PRs so that they can be reviewed in increments.

dual_net.py
```diff
@@ -407,28 +402,32 @@ def my_res_layer(inputs):
...
+
 
+def _get_nontpu_estimator():
     run_config = tf.estimator.RunConfig(
         save_summary_steps=FLAGS.summary_steps,
         keep_checkpoint_max=FLAGS.keep_checkpoint_max)
     return tf.estimator.Estimator(
         model_fn,
-        model_dir=working_dir,
+        model_dir=FLAGS.model_dir,
```
amj: Does this mean that non-tpu training will spew `model.ckpt-123456` into gs://$BUCKET/models/ ....?

amj: Because (here and below) conflating the fsdb 'model_dir' with the 'work_dir' is probably not the right thing...

PR #420, opened 2018-09-08
Factor out validate.py

brilee:
Builds upon #419

### DIFF
--- a/rl_runner.py
+++ b/rl_runner.py
@@ -50,8 +50,9 @@ def loop(working_dir='estimator_working_dir', tpu_name=None):
 
         with timer("validate"):
             subprocess.call(['python', 'rl_loop.py', 'validate-hourly'] + flags)
-            subprocess.call(['python', 'main.py', 'validate',
-                             '--validate-name=pro'] + flags[3:])
+            subprocess.call(['python', 'validate.py',
+                             'gs://jacksona-sandbox/data/validate',
+                             '--validate-name=pro'] + flags[1:])
 
 
 if __name__ == '__main__':
--- a/validate.py
+++ b/validate.py
@@ -1,0 +1,72 @@
+"""Validate a network.
+
+Usage:
+  BOARD_SIZE=19 python validate.py tfrecord_dir/ tfrecord_dir2/
+"""
+import os
+
+from absl import app, flags
+from tensorflow import gfile
+
+import dual_net
+import preprocessing
+import utils
+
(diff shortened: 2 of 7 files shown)


rl_runner.py
```diff
@@ -50,8 +50,9 @@ def loop(working_dir='estimator_working_dir', tpu_name=None):
 
         with timer("validate"):
             subprocess.call(['python', 'rl_loop.py', 'validate-hourly'] + flags)
-            subprocess.call(['python', 'main.py', 'validate',
-                             '--validate-name=pro'] + flags[3:])
+            subprocess.call(['python', 'validate.py',
+                             'gs://jacksona-sandbox/data/validate',
```
sethtroisi: I'm not big on this being checked in. can we move these to minigo-pub or make it a parameter

validate.py
```diff
@@ -0,0 +1,72 @@
...
+        estimator.evaluate(_input_fn, steps=steps, name=FLAGS.validate_name)
+
+
+def main(*validation_paths):
+    """Validate a model's performance on a set of holdout data."""
+    if FLAGS.expand_validation_dirs:
+        tf_records = []
+        with utils.logged_timer("Building lists of holdout files"):
+            for record_dir in validation_paths:
+                tf_records.extend(gfile.Glob(os.path.join(record_dir, '*.zz')))
```
sethtroisi: This will only expand one directory.
consider if multiple directory expansion (with record_dir, '**', '*.tfrecord.zz') would ever be useful.

validate.py
```diff
@@ -0,0 +1,72 @@
...
+flags.declare_key_flag('use_tpu')
+flags.declare_key_flag('num_tpu_cores')
+
+FLAGS = flags.FLAGS
+
+
+def validate(*tf_records):
+    """Validate a model's performance on a set of holdout data."""
+    if FLAGS.use_tpu:
+        def _input_fn(params):
```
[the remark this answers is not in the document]
brilee: I was too lazy to write docstrings to make pylint shut up.
😀

rl_runner.py
brilee: @amj I take it this is being actively used for our validation?

amj: yes.

validate.py
amj: the holdout data directories are gs://$BUCKET/data/holdout/YYYY-MM-DD-HH/, as long as the 'validation paths' is set as all those YYYY-MM-DD-HH paths...

PR #421, opened 2018-09-08
Factor out evaluate.py; unify handling of --verbose flags

brilee:
Factor out evaluate.py; unify handling of --verbose flags @amj 

Builds on #420

### DIFF
--- a/cluster/evaluator/evaluator_wrapper.sh
+++ b/cluster/evaluator/evaluator_wrapper.sh
@@ -26,11 +26,11 @@ echo white:  $MODEL_WHITE
 
 DATE=`date +%Y-%m-%d` 
 
-python3 main.py evaluate \
+python3 evaluate.py \
   $MODEL_BLACK $MODEL_WHITE \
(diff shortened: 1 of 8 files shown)


cluster/evaluator/evaluator_wrapper.sh
```diff
@@ -26,11 +26,11 @@ echo white:  $MODEL_WHITE
 
 DATE=`date +%Y-%m-%d` 
 
-python3 main.py evaluate \
+python3 evaluate.py \
```
amj: does this also need to be changed in the evaluator_cc_wrapper.sh?

main.py
```diff
@@ -94,7 +78,7 @@ def convert(load_file, dest_file):
 
 
 parser = argparse.ArgumentParser()
-argh.add_commands(parser, [bootstrap, evaluate, convert])
+argh.add_commands(parser, [bootstrap, convert])
```
amj: it shrinks!

PR #425, opened 2018-09-10
oneoff script to count number of zero weights

sethtroisi:
~50% of weights appear to be zero (< 1e-10) in v10
Ideally I'd count dead neurons but that seems harder.

All graphs seem to have a constant number of zero weights with very little absolute change over the v10 run
conv2d_ x -kernel

dense_ x -kernel

### DIFF
--- a/oneoffs/l2_cost_by_var.py
+++ b/oneoffs/l2_cost_by_var.py
@@ -1,7 +1,8 @@
 """Plot the l2 cost of various tensors over a training run.
 
 Example usage:
-  python3 oneoffs/l2_cost_by_var.py --base_dir "gs://minigo-pub/v7-19x19/"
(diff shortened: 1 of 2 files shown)


brilee:
I think by definition, a "dead neuron" is something that has zero activation (i.e. pre-ReLU activation is less than zero) over all examples in the empirical distribution. It is termed dead because once the activations are all zero, there is no example in your training data capable of shifting the activation in a direction where it can start learning again.

(that being said, the neurons that feed into it can shift their output, thereby reviving it. So I doubt that neurons are ever really fully dead. Maybe the first layer neurons only.)

amj:
this looks worth revisiting as 50% zeros seems a little high...

amj:
this script seems useful - why did you close the PR with no comment?

PR #430, opened 2018-09-11
Print to dbg to avoid gtp errors

sethtroisi:
using `--verbosity=5` results in gtp errors because of printing to stdout


strategies.py
```diff
@@ -159,7 +159,7 @@ def play_move(self, c):
         try:
             self.root = self.root.maybe_add_child(coords.to_flat(c))
         except go.IllegalMove:
-            print("Illegal move")
+            dbg("Illegal move")
```
sethtroisi: maybe this shouldn't be dbg and should cause gtp to fail but IDK

tommadams: Let's go with `dbg` for now.

It looks like the Python code doesn't handle bad moves properly. The GTP engine should print `? illegal move` with no capitalization in response to a bad move (see section 6.3.3: https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html) but from looking at the code it appears we will currently print `? False`.

I think we need to change [BasicCmdHandler.cmd_play](https://github.com/tensorflow/minigo/blob/a51db4f2d9a13436512c3cbfcb8db47f21ecda2e/gtp_cmd_handlers.py#L71) to raise an exception if `play_move` returns false. Then the Python GTP will report the [correct error message](https://github.com/tensorflow/minigo/blob/a51db4f2d9a13436512c3cbfcb8db47f21ecda2e/gtp_engine.py#L27).

PR #442, opened 2018-09-17
Add performance logging to inference server

sethtroisi:
Spits out some numbers like this
```
(-300) recent: 4.73s/6400 (52.7% in session), total: warming up
(-100) recent: 4.81s/6400 (53.0% in session), total: warming up
(0) recent: 4.71s/6400 (53.0% in session), total: warming up
(100) recent: 4.86s/6400 (53.7% in session), total: 48.60s/64000 inferences
(200) recent: 4.82s/6400 (52.8% in session), total: 48.39s/64000 inferences
(300) recent: 4.92s/6400 (51.6% in session), total: 48.60s/64000 inferences
```
They get interlaced with game debug but they happen often enough to be able to find them.
Existing debug is to find the lines like this in the game debug and multiple by number of games and stuff 
```
Milliseconds per 100 reads: 650ms over 8000 readouts (batched: 8)
```

The code takes a lot of lines which I don't really like but I like the output so I'm for it.


inference_worker.py
```diff
@@ -416,12 +421,21 @@ def _worker_thread(self, thread_id):
             pass
 
         dbg("running worker", thread_id)
+        # Don't start counting till we've warmed up.
+        batches = -500
```
tommadams: Can you remove the special case logic for warming up please? If we're going to start using the inference worker to play via GTP, we really need to perform a warm up inference during initialization, similar to [tf_dual_net.cc](https://github.com/tensorflow/minigo/blob/a3e1df9cc9802b224db6092257245e38e10aa746/cc/dual_net/tf_dual_net.cc#L71), otherwise the time estimation will be all wrong.

sethtroisi: ~I've noticed it takes a couple batches for the whole system to hit steady state, this just delays the initial "start" time by a couple hundred batches so that long term throughput is a little more stable calculation. I can certainly reduce this to -100 (or even -10).~

KISS, I'll remove this.

sethtroisi:
I simplified what I log so it's a much cleaner solution now.

Issue #434, opened 2018-09-11
Check if Python handle illegal move correctly


sethtroisi:
Tom in #430 noted that Python probably handles play move incorrectly when an illegal move is passed.

I'll verify and fix if true.

sethtroisi:
The answer seems to be no

![screenshot from 2018-09-18 22-52-23](https://user-images.githubusercontent.com/10172976/45733421-9048a000-bb95-11e8-8a0b-27e7a44caebc.png)

sethtroisi:
I should verify first

sethtroisi:
verified

PR #445, opened 2018-09-19
Correctly return gtp error for illegal play_move

sethtroisi:
Follow up to #434 

screenshot from 2018-09-18 23-06-09


sethtroisi:
Thanks for noticing this in #430 Tom

amj:
so, why do the tests fail?

strategies.py
```diff
@@ -156,14 +156,10 @@ def play_move(self, c):
...
-            self.root = self.root.maybe_add_child(coords.to_flat(c))
-        except go.IllegalMove:
-            dbg("Illegal move")
-            if not self.two_player_mode:
-                self.searches_pi.pop()
-            self.comments.pop()
-            return False
+        self.root = self.root.maybe_add_child(coords.to_flat(c))
+        if not self.two_player_mode:
+            self.searches_pi.pop()
```
tommadams: These two pops should only be done if maybe_add_child throws, so I think you need to keep the try/catch block and reraise instead of returning false.

strategies.py
```diff
@@ -158,12 +158,13 @@ def play_move(self, c):
...
         try:
             self.root = self.root.maybe_add_child(coords.to_flat(c))
-        except go.IllegalMove:
+        except go.IllegalMove as e:
             dbg("Illegal move")
             if not self.two_player_mode:
                 self.searches_pi.pop()
             self.comments.pop()
-            return False
+            raise e
```
tommadams: `raise e` here will loose the callstack information. Use a plain `raise` to reraise the exception while preserving the callstack. You won't need the `as e` in the `except` line above either.

sethtroisi:
Thanks for the exception help Tom

PR #447, opened 2018-09-19
Replace argh library with fire library.

briandersn:
Fire is slightly more straightforward for minigo's usage
since everything can be passed directly to the fire.Fire()
rather than indirectly through argh.add_commands().

argparse usage is also hidden and doesn't need to be
imported separately and used as a proxy.

### DIFF
--- a/rl_loop.py
+++ b/rl_loop.py
@@ -13,14 +13,13 @@
 # limitations under the License.
 
 """Bookkeeping around other entry points to automate the rl pipeline."""
-import argparse
 import logging
 import os
 import sys
 import time
 
+import fire
 from absl import flags
-import argh
 from tensorflow import gfile
 import itertools
 import random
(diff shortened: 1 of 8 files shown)


rl_loop.py
```diff
@@ -158,13 +157,15 @@ def backfill():
...
 if __name__ == '__main__':
     cloud_logging.configure()
     remaining_argv = flags.FLAGS(sys.argv, known_only=True)
-    argh.dispatch(parser, argv=remaining_argv[1:])
+    fire.Fire({
+        'train': train,
+        'selfplay': selfplay,
+        'backfill': backfill,
+        'bootstrap': bootstrap,
+        'fsdb.game_counts': fsdb.game_counts,
```
brilee: I think this one would be called as `game_counts`, not `fsdb.game_counts`.

brilee:
Looks good other than the one imported method.

rl_loop.py
briandersn: Done.

briandersn:
It looks like the presubmit bots don't have the fire library installed. How can we update them?

brilee:
I think if you add fire to requirements.txt, the docker image will pick up the changes.

briandersn:
I guess updating requirements.txt doesn't work if it's part of the same patch? I'll try pulling in the fire dependency into requirements.txt as a separate patch first.

briandersn:
Added fire to requirements.txt here: https://github.com/tensorflow/minigo/pull/449

PR #448, opened 2018-09-20
Tie break most visited path with action score

sethtroisi:


mcts.py
```diff
@@ -255,11 +255,15 @@ def children_as_pi(self, squash=False):
...
 
+    def _top_child(self):
+        # Sort by child_N tie break with action score.
+        return np.argmax(self.child_N + self.child_action_score / 10000)
+
     def most_visited_path_nodes(self):
         node = self
         output = []
         while node.children:
-            next_kid = np.argmax(node.child_N)
```
brilee: my OCD is complaining about the swapping of "kid" and "child"...

brilee:
LGTM

PR #451, opened 2018-09-21
Moving .bazelrc according to best practices.

chsigg:
See https://docs.bazel.build/versions/master/best-practices.html#bazelrc


tommadams:
Nice, thanks!

PR #452, opened 2018-09-21
Move batching out of remote inference.

chsigg:
Make batching available for all backends.

Move leaves and output around as vectors instead of pre-allocating it in
the MctsPlayer.


chsigg:
All reviewer comments should be addressed (see comments in PR #446).

cc/dual_net/factory.cc
```diff
@@ -168,4 +72,13 @@ std::unique_ptr<DualNetFactory> NewDualNetFactory(std::string model_path,
   return nullptr;
 }
 
+}  // namespace
+
+DualNetFactory::~DualNetFactory() = default;
+
+std::unique_ptr<DualNetFactory> NewDualNetFactory(
```
tommadams: This adds adds a batching wrapper around every DualNet implementation, even for cases where it doesn't make sense (e.g. some accelerators only support a batch size of 1).

PR #454, opened 2018-09-21
Minigui rewrite

tommadams:
Sorry, this is basically a complete rewrite of the whole thing.

Highlights:

- Three separate modes: full demo, lightweight demo, kiosk
- New reactive layout for lightweight demo & kiosk (full demo preserves layout optimized for 1920x1080 displays).
- Shared App base class.
- Boards are composed of separate layers.
- Each layer may be bound to a specific property of incoming GTP messages.
- Boards only request a redraw when one or more of their layers change.
- Boards are redrawn on the next animation frame instead of in response to every change.
- C++ `GtpPlayer` uses `nlohmann::json` to format JSON objects instead of hand written nonsense.


cc/gtp_player.cc
```diff
@@ -290,34 +292,17 @@ GtpPlayer::Response GtpPlayer::HandleGamestate(absl::string_view cmd,
...
     oss << ch;
   }
-  std::string board = oss.str();
-
-  // toPlay field.
-  std::string to_play = position.to_play() == Color::kBlack ? "Black" : "White";
-
-  // lastMove field.
-  std::string last_move;
+  j["toPlay"] = position.to_play() == Color::kBlack ? "B" : "W";
```
sethtroisi: Can you use the format below

nlohmann::json j = {
      {"moveNum", pos.n()},
      {"toPlay", pos.to_play() == Color::kBlack ? "B" : "W"},
    ...
  };

for all the properties (minus lastMove)?

cc/gtp_player.cc
```diff
@@ -521,36 +506,44 @@ GtpPlayer::Response GtpPlayer::HandleReportSearchInterval(absl::string_view cmd,
...
+  };
+
+  auto& search = j["search"];
   std::vector<const MctsNode*> path;
   for (const auto* node = last_read; node != root(); node = node->parent) {
     path.push_back(node);
   }
   for (auto it = path.rbegin(); it != path.rend(); ++it) {
-    std::cerr << " " << (*it)->move.ToKgs();
+    search.push_back((*it)->move.ToKgs());
```
sethtroisi: Can you combine this with the loop above and avoid storting path?

gtp_cmd_handlers.py
```diff
@@ -258,16 +258,17 @@ def cmd_gamestate(self):
...
         msg["board"] = "".join(board)
-        msg["toPlay"] = "Black" if position.to_play == 1 else "White"
         if position.recent:
             msg["lastMove"] = coords.to_kgs(position.recent[-1].move)
         else:
             msg["lastMove"] = None
-        msg["n"] = position.n
+        msg["toPlay"] = "B" if position.to_play == 1 else "W"
+        msg["moveNum"] = position.n
         if root.parent and root.parent.parent:
```
sethtroisi: msg["q"] = root.parent.Q if root.parent and root.parent.parent else 0

gtp_cmd_handlers.py
```diff
@@ -309,23 +310,31 @@ def _minigui_report_search_status(self, leaves):
...
+        msg["dq"] = [int(round(x * 100)) for x in dq]
 
-        n = ['%d' % x for x in root.child_N]
-        dbg("mg-n:%s" % " ".join(n))
+        msg["n"] = [int(n) for n in root.child_N]
 
         nodes = root.most_visited_path_nodes()
-        path = [coords.to_kgs(coords.from_flat(m.fmove)) for m in nodes]
-        dbg("mg-pv:%s" % " ".join(path))
+        msg["pv"] = [coords.to_kgs(coords.from_flat(m.fmove)) for m in nodes]
```
sethtroisi: the c code avoids sending this if it didn't change. is that an easy possibility here?

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+	"target": "es6",
```
sethtroisi: seems to have the wrong indention

minigui/layer.ts
```diff
@@ -0,0 +1,407 @@
...
+  constructor(board: Board, private caption: string) {
+    super(board);
+  }
+
+  draw() {
+    let ctx = this.board.ctx;
+    let size = this.board.size;
+
+    let textHeight = Math.floor(0.4 * this.board.stoneRadius);
+    ctx.font = `${textHeight}px sans-serif`;
```
amj: this syntax is kind of amazing.

minigui/serve.py
```diff
@@ -80,21 +92,23 @@
                    "gtp.py",
                    "--load_file", args.model,
                    "--minigui_mode=true",
-                   "--num_readouts", "1000",
+                   "--num_readouts", "400",
                    "--conv_width", "128",
```
amj: If we're targeting something that's going to use our current models, the default has been 256 at least since v7.  If the default case is gonna be the edge TPU stuff, we can leave it at 128.

amj:
this still has approval

minigui/app.ts
```diff
@@ -0,0 +1,194 @@
...
+          let args = dataset[key];
+          this.gtp.send(`${cmd} ${args}`);
+        }
+      }
+    }
+
+    this.updateBoards(this.activePosition);
+  }
+
+  protected updateBoards(state: any) {
```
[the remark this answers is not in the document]
tommadams: Added a comment.

minigui/app.ts
```diff
@@ -0,0 +1,194 @@
...
+  protected onSearch(msg: SearchMsg) {
+    // Parse move variations.
+    msg.search = util.parseMoves(msg.search as string[], this.size);
+    msg.toPlay = util.parseGtpColor(msg.toPlay as string);
+    if (msg.pv) {
+      msg.pv = util.parseMoves(msg.pv as string[], this.size);
+    }
+
+    // Update the board state with contents of the search.
+    const props = ['n', 'dq', 'pv', 'search'];
```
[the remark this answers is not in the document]
tommadams: Adding more would require updating this list, yes. We'd pretty much have to do that anyway if we want to send extra properties from the backend because the current structures are statically typed.

I guess we could make the SearchMsg and PositionHistory hold arbitrary key/value pairs and rely on the UI to display them correctly. That's an interesting idea, so I've added a TODO for now.

minigui/layer.ts
tommadams: Yeah, I love it. It's one of my favourite things about modern JavaScript

minigui/board.ts
```diff
@@ -12,408 +12,177 @@
...
 
-  clearHeatMap() {
-    this.heatMap.fill(0);
-    this.draw();
-  }
-
-  setHeatMap(map: Array<number>) {
-    for (let i = 0; i < map.length; ++i) {
-      this.heatMap[i] = map[i];
+  update(state: any) {
```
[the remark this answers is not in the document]
tommadams: Sometimes. Sometimes they're a Position.

Added a comment.

PR #458, opened 2018-09-23
Add a oneoff script to generate positions embeddings from the bottom of the shared trunk

sethtroisi:
This was / is used to try and find nearest neighbor positions see
image


oneoffs/embeddings.py
```diff
@@ -0,0 +1,107 @@
...
+
+def main(argv):
+    features, labels = dual_net.get_inference_input()
+    tf_tensors = dual_net.model_inference_fn(features, False)
+    if len(tf_tensors) != 4:
+        print("oneoffs/embeddings.py requires you modify")
+        print("dual_net.model_inference_fn and add a fourth param")
+        sys.exit(1)
+
+    p_out, v_out, logits, shared = tf_tensors
```
amj: It looks like you backed out the changes in dual_net.py -- i don't think this will work without *something* being returned from model_inference_fn

amj: lol

PR #461, opened 2018-09-25
More refactoring7

brilee:
Set up local integration tests


evaluate.py
```diff
@@ -109,8 +109,9 @@ def play_match(black_model, white_model, games, sgf_dir):
                                                                    timeper,
                                                                    dur))
 
-def main(black_model, white_model):
+def main(argv):
     """Play matches between two neural nets."""
+    _, black_model, white_model = argv
     utils.ensure_dir_exists(FLAGS.eval_sgf_dir)
```
amj: worth validating that the models exist too?

validate.py
```diff
@@ -54,19 +54,20 @@ def _input_fn():
...
         with utils.logged_timer("Building lists of holdout files"):
             for record_dir in validation_paths:
-                tf_records.extend(gfile.Glob(os.path.join(record_dir, '**', '*.zz')))
+                tf_records.extend(gfile.Glob(os.path.join(record_dir, '*.zz')))
     else:
         tf_records = validation_paths
 
+    if not tf_records:
+        raise RuntimeError("Did not find any holdout files for validating!")
     validate(*tf_records)
```
amj: interestingly, the number of steps at the beginning of a run is often greater than the number of holdout games that have been played.  (no action required)

PR #463, opened 2018-09-25
Turn embeddings into t-SNE coordinates for graphing

sethtroisi:
screenshot from 2018-09-25 01-33-31


sethtroisi:
builds on #462

oneoffs/embeddings.py
```diff
@@ -75,10 +76,12 @@ def main(argv):
 
     try:
       progress = tqdm(get_files())
-      embeddings = np.empty([len(progress), FLAGS.embedding_size])
+      embeddings = []
       metadata = []
       for i, f in enumerate(progress):
         short_f = os.path.basename(f)
+        short_f = short_f.replace('-minigo-cc-evaluator', '')
```
brilee: shipname.py has utilities for extracting gen numbers from strings

brilee:
BTW, tensorboard allows you to interactively fiddle with tSNE hyperparameters using the same interface you have here (n-dimensional vector associated with a debug image).

brilee:
Would love to get a zoomable version of that screenshot if possible!

sethtroisi:
@brilee https://cloudygo.com/v10-19x19/tsne/value_conv_tsne_small.graph?x=223

oneoffs/embeddings.py
sethtroisi: this is just to make the name short enough that tqdm fits on one line.
it prints something like


15367123-250-320-250-herucles-320-sunset-10.sgf |xxxxx.....| 52% 1234/10000 
and if that line is longer than your terminal it prints for each file.
you can make your terminal wider but removing ~15 characters also helps.

PR #469, opened 2018-09-28
Minigo bt output

gitosaurus:
This pull requests allows the C++ Minigo implementation to optionally link against Cloud Bigtable and write out its game examples to a table in parallel with writing them to GCS, as it does today.

This does not address the input pipeline yet, and support is disabled by default.

Minor fixes to `requirements.txt` and `README.md` are included too.


cc/README.md
```diff
@@ -45,7 +46,7 @@ enabled for a particular board size. Consequently, you must run the tests
 multiple times:
 
 ```shell
-bazel test --define=board_size=9 cc/...  &&  bazel test cc/...
+bazel build --define=board_size=9 cc/...  &&  bazel test cc/...
```
sethtroisi: I don't think this should have changed.

cc/README.md
```diff
@@ -69,7 +70,7 @@ To run Minigo with a 9x9 model:
 
 ```shell
 bazel build --define=board_size=9 -c opt cc:main
-bazel-bin/cc/main --model=$MODEL_PATH
+bazel-bin/cc/main --model=$MODEL_PATH --mode=selfplay
 ```
 
 To run Minigo with a 19x19 model:
```
sethtroisi: line three down from here should also get --mode=selfplay (or gtp)

cc/main.cc
```diff
@@ -145,6 +151,17 @@ std::string GetOutputName(absl::Time now, size_t i) {
   return absl::StrCat(timestamp, "-", hostname, "-", i);
 }
 
+std::string GetRowPrefix(absl::Time now, int id) {
+  std::ostringstream out;
+  absl::TimeZone utc = absl::UTCTimeZone();
+  out << absl::FormatTime("%Y%m%dT%H%M%S", now, utc)
```
sethtroisi: This the T required? could it be changed to a _?

cc/tf_bt_utils.cc
```diff
@@ -0,0 +1,69 @@
...
+
+// Writes a list of tensorflow Example protos to a series of Bigtable rows.
+int WriteTfExamples(Table& table,
+                    const std::string& row_prefix,
+                    int row_cursor,
+                    const std::vector<tensorflow::Example>& examples) {
+  for (const auto& example : examples) {
+    std::string data;
+    example.SerializeToString(&data);
+    char buf[64];
```
sethtroisi: nit: this doesn't need to be 64 if you only write a 6digit number into it

cc/README.md
gitosaurus: Oops, my mistake. Will revert that.

cc/README.md
gitosaurus: Thanks. Will fix.

cc/tf_bt_utils.cc
gitosaurus: True. To be honest, I'm wondering whether 6 digits is overachieving, too. It only needs to hold the number of examples written out by a single thread worker. Is there a more sensible upper bound?

cc/main.cc
gitosaurus: I used the `T` only because that's ISO 8601, but honestly, there's no reason why it has to conform to that. The valuable part is having a timestamp that sorts lexicographically. I'll change it.

cc/tf_bt_utils.cc
sethtroisi: If the number only needs to be unique to a games then 4d would work (max number of moves is 722)
each worker generates ~ 1 moves every 30 seconds but might run for days so <= 100,000

gitosaurus: Ok. I'll leave it as `%06d`, then, in the interest of not over-optimizing.

cc/main.cc
```diff
@@ -118,6 +120,10 @@ DEFINE_string(output_dir, "",
               "Output directory. If empty, no examples are written.");
 DEFINE_string(holdout_dir, "",
               "Holdout directory. If empty, no examples are written.");
+DEFINE_string(output_bigtable, "",
+              "Output Bigtable specification, of the form: "
+	      "project,instance,table. "
```
tommadams: nit: alignment is off here

cc/main.cc
```diff
@@ -145,6 +151,17 @@ std::string GetOutputName(absl::Time now, size_t i) {
   return absl::StrCat(timestamp, "-", hostname, "-", i);
 }
 
+std::string GetRowPrefix(absl::Time now, int id) {
+  std::ostringstream out;
+  absl::TimeZone utc = absl::UTCTimeZone();
+  out << absl::FormatTime("%Y%m%d-%H%M%S", now, utc)
```
tommadams: Can you use absl::StrFormat here? I find it easier to read than iomanip.

something like:

```
absl::StrFormat("%s-w%03d-", absl::FormatTime("%Y%m%d-%H%M%S", now, utc), id)
```

We do have a bunch of ostringstring in the C++ codebase because that was written before absl::StrFormat was released. I'll file an issue to clean those up.

cc/main.cc
```diff
@@ -145,6 +151,17 @@ std::string GetOutputName(absl::Time now, size_t i) {
...
+      << "-w"
+      << std::setw(3)
```
tommadams: Use 4 digit padding (or even higher): Christian Sigg has been experimenting with up to 2048 parallel games.

cc/main.cc
```diff
@@ -331,6 +348,13 @@ class SelfPlayer {
     const bool use_ansi_colors = isatty(fileno(stderr));
 
     GameOptions game_options;
+    int row_cursor = 0;
+    std::vector<std::string> bigtable_spec = absl::StrSplit(FLAGS_output_bigtable, ',');
+    bool use_bigtable = bigtable_spec.size() == 3;
+    if (!bigtable_spec.empty() && !use_bigtable) {
+      std::cerr << "Bigtable output must be project,instance,table" << std::endl;
```
tommadams: Let's make this a fatal error: `MG_FATAL() << ...`

cc/main.cc
```diff
@@ -384,6 +408,25 @@ class SelfPlayer {
         tf_utils::WriteGameExamples(GetOutputDir(now, example_dir), output_name,
                                     *player);
       }
+      if (use_bigtable) {
+        int incoming_cursor = row_cursor;
+        std::string gcp_project_name = bigtable_spec[0];
```
tommadams: nit: `s/std::string/const auto&/` to avoid making copies

cc/tf_utils.cc
```diff
@@ -97,11 +96,7 @@ void WriteTfExamples(const std::string& path,
 
 }  // namespace
 
-void WriteGameExamples(const std::string& output_dir,
-                       const std::string& output_name,
-                       const MctsPlayer& player) {
-  MG_CHECK(file::RecursivelyCreateDir(output_dir));
-
+std::vector<tensorflow::Example> FetchExamples(const MctsPlayer& player) {
```
tommadams: `FetchExamples` sounds like it's reading them from somewhere: the first bit of this pull request I started reviewing was tf_bt_utils.cc and I initially thought that `FetchExamples` was reading from BigTable.

How about something like `MakeExamples` or `CreateExamples`?

cc/main.cc
tommadams: Is the unique ID for a game the combination of timestamp & thread ID? If so, we should probably use more fine grained granularity than one second. It's entirely possible (though admittedly unlikely) for a game to take less than one second with a low number of reads. For example, randomization during selfplay could cause black to open on a corner point, after which black will resign on its next turn (once training has progressed sufficiently).

tommadams: Actually, we might need to update our version of absl before you can use absl::StrFormat. Let me do that now.

cc/main.cc
gitosaurus: Sorry about that! Fixed.

cc/main.cc
gitosaurus: No problem. Will move to `absl::StrFormat`.

cc/main.cc
gitosaurus: Cool! Will use 6-digit padding, because a million is a nice round number.

cc/main.cc
gitosaurus: Good idea. Will return early upon this error as well. Since the Bigtable write was optional and in parallel, I had let it go by with a warning, but I agree that if the specification is given and is wrong, it's better to stop.

cc/main.cc
gitosaurus: At present, the unique ID is definitely timestamp + thread ID, so the point you bring up is an important one. I could add more precision to the time stamp, but based on a discussion today with amj and sethtroisi, a timestamp is less useful than a true (and atomic) game ID.

I may do the work to move to an atomic game ID in a later PR, in the interest of keeping this one manageable in size.

PR #482, opened 2018-10-04
Make DualNet implementations private

chsigg:
Just moving code around, no functional changes.


chsigg:
So many commits...

Should be ready for review now.

PR #490, opened 2018-10-08
Use --define=mgpu=1 for multi-GPU TfDualNet

gitosaurus:
Without this flag, the original `TfDualNet` that uses only
a TF Session without regard to GPU entities will be the
DualNet selected with `--define=tf=1`.  This allows TF+CPU
to continue building and running.


cc/dual_net/tf_mgpu_dual_net.cc
```diff
@@ -0,0 +1,199 @@
...
+      InferenceData inference;
+      if (inference_queue_.PopWithTimeout(&inference, absl::Seconds(1))) {
+        worker.RunMany(std::move(inference.features),
+                       std::move(inference.outputs));
+        inference.notification->Notify();
+      }
+    }
+  };
+
+  TF_CHECK_OK(tensorflow::ValidateGPUMachineManager());
```
chsigg: Sorry about breaking the GPU-less TensorFlow path! I assume that's what happened, because the single-GPU path should not be affected. 

For the GPU-less path, I think it would be sufficient to add a single worker thread here that's not placed on a device. Something like replacing line 162 with:

if (!tensorflow::ValidateGpuMachineManager().ok()) {
  worker_threads_.emplace_back(functor, graph_def);
  return;
}

gitosaurus: Yes, it was the ability to run on the CPU that I noticed; I haven't tried the single-GPU path.

If I take the approach you describe of having a single worker that's not on a device, I'll still have the need to `#ifdef`-out GPU-related code, correct? It was a missing `#include` that led me down this path in the first place.

gitosaurus: Tested #493 and it worked great for me. Will withdraw this request.

gitosaurus:
Closed in favor of #493.

Issue #459, opened 2018-09-24
"Rough Gating" work required


amj:
# Gating work checklist
1. New "schema" for location of data in bucket 
1. Changes to example_buffer / data assembly/preprocessing
1. Changes to c++ "new model" code.
1. Changes to c++ flag reloading w.r.t. resign threshold.
1. Change name generation code to not make the same name twice.

Biggest design question:  Keep with the GCS-flags method, or bite the bullet and do a proper server to serve out selfplay configs.

Any GCS solution will have "spillover", where some set of selfplay workers get flags that are obsolete.  I.e., we want 25k games with Model X, and we have 24.9k finished.  We'll have about N
Overall, the process of having the resign disabled games played upfront (to compute the proper threshold) has a couple of consequences.  These games are expensive and we want to play as few of them as possible, and we don't want to have more than 10% of them.

Any server based solution will also have to deal with the fact that our TPU workers are not currently able to split games between different models.  But the slop will be less.

## New data arrangement:
Old:
-   /data/selfplay/YYYY-MM-DD-HH
-   /sgf/{clean,full}/YYYY-MM-DD-HH
-   /sgf/eval/YYYY-MM-DD-HH

New:
-   /data/selfplay/$MODEL
-   /data/selfplay/$MODEL/resign_disabled
-   /sgf/{clean,full}/$MODEL
-   /sgf/eval/$MODEL  <-- games used to determine promo/no-promo

## Changes to data marshalling

Obviously, now need to look in /data/selfplay/$MODEL vs the other way.  Otherwise, same deque design to keep the most recent 500k, plus our rsync logic gets simpler.
Depending on how we do resign-disable & resign-threshold computation, potentially limit the resign-threshold games to 2500 (because some resign-disable games could come in after the fact)

## Changes to "new model" code
When a new model comes in, we don't want to reload it.  So, how to spin down?  Tom's first idea is just: kill the games in progress when a new model comes up.  On average, you lose `0.5 * parallel_games * num_workers`, which is kind of painful. 

## Changes to flag reloading
Flags can be updated with the new resign-threshold.  Similar to new_model, we can either a) kill the resign_disabled games (easy) or b) retroactively apply the newly computed resign threshold.

## Pseudocode for "commander"/"dispatcher"
```
while {
if in SELFPLAY state
  Check for # of resign disabled games played
    -> If 10%, rsync, parse, and set resign threshold.

  Check for new models
     if so, send for evaluation, enter PENDING EVAL state

[...]

PR #492, opened 2018-10-09
Stop and reloading all c++ games if certain flag updates.

sethtroisi:
This is part of #459 see comment there on how it will be used


cc/main.cc
```diff
@@ -374,6 +376,11 @@ class SelfPlayer {
         }
         player->PlayMove(move);
       }
+      if (!flags_version = required_flags_) {
```
tommadams: shouldn't this be `flags_version != required_flags_`

cc/main.cc
```diff
@@ -456,13 +465,21 @@ class SelfPlayer {
...
   }
 
   absl::Mutex mutex_;
   std::unique_ptr<DualNetFactory> dual_net_factory_ GUARDED_BY(&mutex_);
   Random rnd_ GUARDED_BY(&mutex_);
   std::vector<std::thread> threads_;
+  // Timestamp of newest flag file.
   uint64_t flags_timestamp_ = 0;
+  // Signals all currently playing games to stop and restart with new flags.
+  uint64_t required_flags_ = 0;
```
tommadams: `required_flags_` sounds like it could be a set of flags or something. How about something like `required_flags_timestamp_`?

cc/main.cc
```diff
@@ -456,13 +465,21 @@ class SelfPlayer {
       gflags::SetCommandLineOption(flag_value.first.c_str(),
                                    flag_value.second.c_str());
     }
+
+    if (FLAGS_resign_threshold < old_resign && old_resign == 1.0) {
```
tommadams: I don't understand this logic, could you add an explanatory comment please? Why isn't this just `FLAGS_resign_threshold != old_resign`?

cc/main.cc
```diff
@@ -456,13 +465,21 @@ class SelfPlayer {
...
+      required_flags_ = flags_timestamp_
+    }
   }
 
   absl::Mutex mutex_;
   std::unique_ptr<DualNetFactory> dual_net_factory_ GUARDED_BY(&mutex_);
   Random rnd_ GUARDED_BY(&mutex_);
   std::vector<std::thread> threads_;
+  // Timestamp of newest flag file.
   uint64_t flags_timestamp_ = 0;
```
tommadams: Since `flags_timestamp_` and `required_flags_` are compared from different threads, they either need to be protected by a mutex or change them to `std::atomic<uint64_t>`, which is probably simpler.

cc/main.cc
sethtroisi: I shortened it to avoid a line wrap. renamed per suggestion.

cc/main.cc
sethtroisi: writing is only done in MaybeReloadFlags which is guarded by mutex I believe this is sufficient

sethtroisi:
This is based on 500 and some hacks to disable tensorrt which can be cleaned up before submit.

cc/main.cc
sethtroisi: Added a comment to explain logic

cc/main.cc
tommadams: Sadly not.

Since reading isn't guarded by a mutex, it's possible on 32bit architectures that a read could see a half-written value for `flags_timestamp_`.

cc/main.cc
tommadams: Thanks for the explanation. I think I understand it, but I still don't see why `FLAGS_resign_threshold != old_resign` wouldn't work just as well and is a safer and more easily understandable test: I really do think we want to abandon games whenever the resign threshold changes.

When playing resign disabled games, you'd run with `--resign_disable_pct=1 --resign_threshold=-1` and the normal games are run with  `--resign_disable_pct=0 --resign_threshold=-XXX`

Also, you should probably check absolute values of the resign thresholds since `resign_threshold` is treated the same as `-resign_threshold` by `MctsPlayer`.

tommadams: Seth made the point offline that we don't need to abort the games if the absolute value of the resign threshold is raised. So I think the abort check should be:

```
if (std::abs(FLAGS_resign_threshold) < std::abs(old_resign)) {
  // abort games
}
```

amj:
Did we resolve the questions here from Tom?

tommadams:
> Did we resolve the questions here from Tom?

My recollection of the conversation that resulted from this PR is that we need to take a different approach to avoid wasting compute, but exactly what that approach should be wasn't decided.

I'm happy to go with this for now though, since it's a step in the right direction. I still think that `old_resign == 1.0` is confusing and would prefer something like my suggestion in that comment thread.

amj:
I was thinking of a simpler idea:  Having the resign-disabled be per-machine instead of per-thread.  I.e., we can have our resign-disable games play fewer games (to play them faster), and that also means those machines don't need to worry about this logic at all.

amj:
closed, obsolete.  offline discussion concluded by saying that we should have a second cluster only playing resign-disabled games with fewer parallel games to ensure freshness

amj:
(this isn't actually a solution for gating)

PR #493, opened 2018-10-09
Use GPUs (in TfDualNet) only when --define=gpu=1 is set.

chsigg:
This is a hopefully less intrusive variant of PR #490.


gitosaurus:
Tested this out on my setup and it worked like a champ. Thanks!

PR #494, opened 2018-10-09
Fix incorrect bigtable check.

chsigg:


gitosaurus:
This change looks fine to me, and it's arguably clearer in its intent. Out of curiosity, however, what test case did you find that exposed a problem? The existing code tests for a non-empty string vector, and the new code tests for a non-empty string, but I believe the effect is the same, since `absl::StrSplit` gives an empty vector for an empty string.

chsigg:
I tried to run the following command: bazel run -c opt --define=board_size=9 //cc:main -- --engine=tf --mode=selfplay --model=$HOME/mlperf/models/model --parallel_games=2 --num_readouts=200 --virtual_losses=8

The bigtable_spec.size() is 1, triggering the MG_FATAL. It looks like absl::SplitStr returns an empty string if given an empty string: https://github.com/abseil/abseil-cpp/blob/e821380d69a549dc64900693942789d21aa4df5e/absl/strings/str_split_test.cc#L343

gitosaurus:
Beg your pardon! Blind spot on my part. Thanks for the example.

PR #495, opened 2018-10-09
Upgrade cc/configure_tensorflow.sh

tommadams:
- Upgrade to TF v1.11
- Build the gRPC runtime required by Cloud TPU.


requirements.txt
```diff
@@ -5,6 +5,7 @@ autopep8>=1.3
 fire
 google.cloud.logging
 grpcio-tools
+keras
```
brilee: what's this about?

tommadams: We build the tensorflow install package in order to extract the C++ headers and keras is a required TF dependency now :/

I _really_ want to start using the user's TF install.

cc/configure_tensorflow.sh
```diff
@@ -60,29 +58,54 @@ echo "Copying tensor flow headers to ${dst_dir}"
 cp -r ${tmp_dir}/tensorflow-*.data/purelib/tensorflow/include/* "${dst_dir}"
 
 echo "Building tensorflow libraries"
-bazel build -c opt --config=opt --copt="${cc_opt_flags}" //tensorflow:libtensorflow_cc.so //tensorflow:libtensorflow_framework.so
+
+# Add a custom BUILD target for the gRPC runtime.
```
amj: this is both hilarious and awful.

amj: and i'm also impressed -- i don't think i would've come up with this as a solution XD

PR #498, opened 2018-10-11
Eval: parallel games and even mode.

chsigg:
evaluation where the model and model_two play black in turn.


cc/main.cc
```diff
@@ -65,6 +66,8 @@ DEFINE_uint64(seed, 0,
               "Random seed. Use default value of 0 to use a time-based seed. "
               "This seed is used to control the moves played, not whether a "
               "game has resignation disabled or is a holdout.");
+DEFINE_bool(even, false,
```
amj: this should probably be the default.  i don't think there's any reason not to do this, if parallel_games > 1

cc/main.cc
```diff
@@ -465,46 +468,232 @@ class SelfPlayer {
   uint64_t flags_timestamp_ = 0;
 };
 
+class Evaluator {
+  class Barrier {
+   public:
+    explicit Barrier(size_t count)
```
amj: this barrier stuff seems somewhat complicated; is there a reason we can't just use two queues, one for each model?

cc/main.cc
```diff
@@ -465,46 +468,232 @@ class SelfPlayer {
...
+
+    auto* black = player.get();
+    auto* white = other_player.get();
+
+    auto* factory = model->factory.get();
+    auto* other_factory = other_model->factory.get();
+
+    while (!player->game_over()) {
+      dual_net = factory->New();
+      barrier_->Wait();  // Wait for all threads to create DualNet.
```
amj: sorry, i'm not sure i understand why this is needed?

cc/main.cc
```diff
@@ -465,46 +468,232 @@ class SelfPlayer {
...
+    for (auto& t : threads_) {
+      t.join();
+    }
+
+    std::cerr << "Evaluated " << num_games << " games, total time "
+              << absl::ToDoubleSeconds(absl::Now() - start_time) << " sec."
+              << std::endl;
+
+    auto print_name = [](std::string name) {
+      if (name.size() > 20) {
```
amj: would absl::strformat be useful here?

cc/main.cc
```diff
@@ -465,46 +468,232 @@ class SelfPlayer {
   uint64_t flags_timestamp_ = 0;
 };
 
+class Evaluator {
```
amj: this seems like something that would be an ideal candidate to break out of main.cc

cc/main.cc
chsigg: Done. For parallel_games == 1, it will just work as before.

cc/main.cc
chsigg: I've added additional comments to hopefully explain why this is needed.

cc/main.cc
chsigg: That would result in two models performing inference at the same time (with half the batch size in average). Also see my comments around line 650.

cc/main.cc
chsigg: Yes. Tom requested that I hold off chopping up main.cc for now. Will do in a separate CL.

cc/main.cc
chsigg: I don't think absl::StrFormat could insert the ellipsis. I did managed to make it a little more concise though.

tommadams:
Oh dear... I just realized all the comments I had intended for this PR on #508. 

I'll try and duplicate the comments here too

cc/main.cc
```diff
@@ -16,6 +16,7 @@
 #include <unistd.h>
 #include <cstring>
 #include <functional>
+#include <iomanip>
```
tommadams: please use absl::StreamFormat for consistency with the rest of the codebase

cc/main.cc
```diff
@@ -65,6 +66,8 @@ DEFINE_uint64(seed, 0,
               "Random seed. Use default value of 0 to use a time-based seed. "
               "This seed is used to control the moves played, not whether a "
               "game has resignation disabled or is a holdout.");
+DEFINE_bool(even, true,
```
tommadams: I don't really know what `even` means in this context. Can you choose a more descriptive name or alternatively just remove the flag altogether; I'm not sure there's ever a realistic reason to set it to false is there?

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
   uint64_t flags_timestamp_ = 0;
 };
 
+class Evaluator {
+  // A barrier that blocks threads until the number of waiting threads reaches
+  // the 'count' threshold. Allows decrementing the threshold to handle the tail
+  // of a work queue where some threads exit early.
+  class Barrier {
```
tommadams: To avoid confusion, it's probably worth noting that "This Barrier has substantially different semantics than absl::Barrier".

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
...
+    size_t count_ GUARDED_BY(&mutex_);
+    size_t num_waiting_ GUARDED_BY(&mutex_);
+    size_t generation_ GUARDED_BY(&mutex_);
+  };
+
+  // References a pointer to an actual DualNet. Allows updating the pointer
+  // after the MctsPlayer has been constructed.
+  class WrappedDualNet : public DualNet {
+   public:
+    explicit WrappedDualNet(const std::unique_ptr<DualNet>& dual_net)
```
tommadams: const std::unique_ptr<DualNet>* please. See the internal C++ Tip of the Week 116 for more details.

To be honest, the necessity of this class strongly suggests to me that the current DualNet & factory design is no longer working for your requirements. As discussed offline, let's go ahead with this for now because of the tight schedule and we'll revisit soon when we implement eval on Cloud TPU.

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
...
+        : dual_net_(dual_net) {}
+
+   private:
+    void RunMany(std::vector<const BoardFeatures*> features,
+                 std::vector<Output*> outputs, std::string* model) override {
+      dual_net_->RunMany(std::move(features), std::move(outputs), model);
+    };
+
+    const std::unique_ptr<DualNet>& dual_net_;
```
tommadams: const std::unique_ptr<DualNet>*

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
...
+    std::unique_ptr<DualNetFactory> factory;
+    std::atomic<int> black_wins;
+    std::atomic<int> white_wins;
+  };
+
+ public:
+  void Run() {
+    auto start_time = absl::Now();
+
+    auto create_model = [](const std::string& model_path) {
```
tommadams: Why isn't this a constructor of Model?

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
...
+
+    std::cerr << "Evaluated " << num_games << " games, total time "
+              << absl::ToDoubleSeconds(absl::Now() - start_time) << " sec."
+              << std::endl;
+
+    auto print_name = [](const std::string& name) {
+      if (name.size() > 20) {
+        std::cerr << name.substr(0, 17) << "...";
+      } else {
+        std::cerr << std::setw(20) << name;
```
tommadams: absl::StreamFormat

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
...
+    options_.random_symmetry = true;
+
+    int num_games = FLAGS_parallel_games;
+    barrier_ = absl::make_unique<Barrier>(num_games);
+
+    for (int i = 0; i < num_games;) {
+      threads_.emplace_back(std::bind(&Evaluator::ThreadRun, this, i,
+                                      cur_model.get(), prev_model.get()));
+      ++i;
+      if (FLAGS_even && i < num_games) {
```
tommadams: Instead of putting the "even" logic here in a separate lambda, why make call barrier_->Wait() at the top of ThreadRun for only the threads with odd ids? The two Wait calls would be closer in the code that way too, which aids readability.

cc/main.cc
```diff
@@ -464,46 +467,244 @@ class SelfPlayer {
...
+
+    auto print_name = [](const std::string& name) {
+      if (name.size() > 20) {
+        std::cerr << name.substr(0, 17) << "...";
+      } else {
+        std::cerr << std::setw(20) << name;
+      }
+    };
+    auto print_wins = [&](int wins) {
+      std::cerr << "   " << std::setw(3) << wins << " " << std::setw(6)
```
tommadams: absl::StreamFormat

chsigg:
> Oh dear... I just realized all the comments I had intended for this PR on #508.
> I'll try and duplicate the comments here too

That would have been no problem, but thanks a lot.

cc/main.cc
chsigg: Agreed, removed.

cc/main.cc
chsigg: Done.

cc/main.cc
chsigg: Done.

cc/main.cc
chsigg: We might indeed need to reintroduce the Client ;-)

cc/main.cc
chsigg: Tip of the Week 116 advises to use a reference for exactly this case, right?

tommadams: No, quite the opposite in fact.

The "Risk of dangling reference in classes" example shows how this can lead to a segfault and recommends using a pointer instead.

```c++

class Foo {
 public:
  explicit Foo(const string& content) : content_(content) {}
  const string& content() const { return content_; }

 private:
  const string& content_;
};

void Func() {
  Foo foo("something");
  LOG(INFO) << foo.content();  // BOOM!
}
```

tommadams: Sorry, I was confused with this comment and the constructor parameter.

Regardless... despite what the tip of the week says, member references are rarely used in google3 (at least in all the parts that I've worked in) and not used at all in Minigo. Please use a pointer for consistency.

cc/main.cc
tommadams: prefer `absl::StreamFormat` over `absl::StrFormat` since the latter will create a temporary string.

cc/main.cc
chsigg: Done.

Issue #441, opened 2018-09-17
Document inference_service_pb2


sethtroisi:
inference_worker.py doesn't work without it.

Maybe has to do with gen_inference_service_py_srcs.sh but I can't find any documentation on it.

sethtroisi:
My issue was failing to activate my virtual env which meant I got python2 instead of python3.

Also do we want generated files checked it?

PR #500, opened 2018-10-13
Remove remote inference support

### LINKED ISSUE #499: cc broken at head?
screenshot from 2018-10-11 23-52-00

tommadams:
Fixes #499 

Remote inference was added as a way to get Cloud TPU running. Cloud TPU from C++ is on the way, so remote inference is no longer required.


sethtroisi:
Fixes https://github.com/tensorflow/minigo/issues/441

PR #504, opened 2018-10-15
Create C++ tool for converting ZLIB TFRecord files to Bigtable rows

gitosaurus:
Used to convert old self-play games into the Bigtable format for testing.

### DIFF
--- a/cc/tf_bt_utils.cc
+++ b/cc/tf_bt_utils.cc
@@ -29,6 +29,7 @@
 #include "tensorflow/core/lib/io/record_reader.h"
 #include "tensorflow/core/platform/env.h"
 #include "tensorflow/core/platform/file_system.h"
+#include "tensorflow/core/platform/logging.h"
 
 using google::cloud::bigtable::bigendian64_t;
 using google::cloud::bigtable::CreateDefaultDataClient;
@@ -176,6 +177,10 @@ void PortGamesToBigtable(const std::string& gcp_project_name,
   table.BulkApply(std::move(game_batch));
   auto finish_time = absl::Now();
   double elapsed = absl::ToDoubleSeconds(finish_time - start_time);
+  VLOG(2) << "Total changes: " << changes << " at " << changes / elapsed
+          << " changes/second";
+  VLOG(2) << "Total games: " << paths.size() << " at " << paths.size() / elapsed
+          << " games/second";
 }
 
 }  // namespace tf_utils
--- a/cc/tf_bt_utils.cc
+++ b/cc/tf_bt_utils.cc
@@ -17,17 +17,27 @@
 #include <cstdint>
 
 #include "absl/strings/str_format.h"
+#include "absl/time/clock.h"
+#include "absl/time/time.h"
 #include "cc/constants.h"
 #include "cc/mcts_player.h"
+#include "google/cloud/bigtable/data_client.h"
 #include "google/cloud/bigtable/read_modify_write_rule.h"
 #include "google/cloud/bigtable/table.h"
 #include "tensorflow/core/example/example.pb.h"
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/lib/io/record_reader.h"
+#include "tensorflow/core/platform/env.h"
+#include "tensorflow/core/platform/file_system.h"
 
 using google::cloud::bigtable::bigendian64_t;
 using google::cloud::bigtable::CreateDefaultDataClient;
 using google::cloud::bigtable::ReadModifyWriteRule;
 using google::cloud::bigtable::SetCell;
 using google::cloud::bigtable::Table;
+
+using tensorflow::io::RecordReaderOptions;
+using tensorflow::io::SequentialRecordReader;
 
 namespace minigo {
 namespace tf_utils {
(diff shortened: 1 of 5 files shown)


cc/tfrzz_to_cbt.cc
```diff
@@ -0,0 +1,154 @@
...
+  int conversion_batch = FLAGS_conversion_batch;
+  auto full_start = absl::Now();
+  while (!paths.empty()) {
+    std::vector<std::string> batch;
+    for (int i = 0; !paths.empty() && i < conversion_batch; ++i) {
+      batch.push_back(paths.front());
+      paths.pop_front();
+    }
+    // Run each batch in a separate process in order to manage the
+    // extraordinary growth in a Bigtable client (may be a bug).
```
amj: how big is extraordinary?  why does it make you think it's a bug?

gitosaurus: Will rebase onto the latest. As to the memory growth, it grows so fast as to cause the process to fail after about 490 games. Because of [this link](https://github.com/GoogleCloudPlatform/google-cloud-go/issues/375), I suspect it may be the same issue, one that also has to be fixed for the C++ client. The gRPC channel setting to turn off tracing is not available in this version (yet). Am reaching out to the developers to confirm. Meanwhile, the separate-process approach also allowed me to greatly increase my conversion rate through multiprocessing, so I had more than one motive.

gitosaurus: I found the bug: https://github.com/grpc/grpc/issues/15340

and the fix got merged 10 days ago: https://github.com/grpc/grpc/pull/16767

Once this becomes available, I'll need to add the new parameter to the channel arguments. Till then, the fork-per-batch ends up limiting it as well.

gitosaurus:
Have rebased, and updated my comments on memory use to reference the gRPC bug.

amj:
Nice, thanks!  LGTM, merge at will

sethtroisi:
@amj Do we want a cc/oneoff directory?

cc/tf_bt_utils.cc
```diff
@@ -82,5 +91,92 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
+  }
+
+  google::cloud::bigtable::BulkMutation game_batch;
+  auto start_time = absl::Now();
+  int changes = 0;
+  for (const auto& path : paths) {
+    std::unique_ptr<tensorflow::RandomAccessFile> file;
+    TF_CHECK_OK(tensorflow::Env::Default()->NewRandomAccessFile(path, &file));
+
+    RecordReaderOptions options;
```
sethtroisi: Nit: I think it's slightly preferred to use

`RecordReaderOptions options = RecordReaderOptions::CreateRecordReaderOptions("ZLIB")`

cc/tf_bt_utils.cc
```diff
@@ -82,5 +91,92 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
+    options.compression_type = RecordReaderOptions::ZLIB_COMPRESSION;
+    SequentialRecordReader reader(file.get(), options);
+
+    auto row_prefix = absl::StrFormat("g_%010d", game_counter);
+
+    auto game_id = path;
+    game_id.erase(game_id.rfind(".tfrecord.zz"));
```
sethtroisi: can you add a small comment like

// transforms "/home/user/test/16123.tfrecord.zz" to "16123"

cc/tf_bt_utils.cc
```diff
@@ -82,5 +91,92 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
+      game_batch.emplace_back(std::move(row_mutation));
+      move_number++;
+      changes++;
+    }
+    game_counter++;
+  }
+
+  table.BulkApply(std::move(game_batch));
+  auto finish_time = absl::Now();
+  double elapsed = absl::ToDoubleSeconds(finish_time - start_time);
```
sethtroisi: seems unused

cc/tf_bt_utils.cc
gitosaurus: Thanks! Will fix.

cc/tf_bt_utils.cc
gitosaurus: Definitely. Thanks for the catch.

cc/tf_bt_utils.cc
gitosaurus: Sure enough. I had been having this routine log the individual batch times, but once I was finished, only the top-level stats were interesting. I'll add the report using `VLOG(2)` (from tensorflow), unless there is a reason to use something else.

cc/tf_bt_utils.cc
gitosaurus: Actually, now that I'm there, I realize that you're recommending not only the factory method, but taking the compression as a string rather than as an enumeration. I recognize that such a pattern is valuable when wanting to plumb through command-line arguments, putting the user easily in charge of the configuration, but when the format is known, wouldn't the enumeration be better?

sethtroisi: To use "GZIP" you should really set more than just the compression type hence the more general need for the helper method.

This is non-production code, on a 20% project. Do what makes you happier. Either is fine be me.

PR #508, opened 2018-10-16
Puzzle mode

chsigg:
Add puzzle mode (trying to predict the next move of prerecorded games).

This is on top of eval PR #498. Please only look at the latest commit. I will squash-submit the PRs sequentially.


cc/main.cc
```diff
@@ -519,6 +726,73 @@ void Gtp() {
...
+    if (EndsWith(filename, ".sgf")) {
+      sgf_files.push_back(filename);
+    }
+  }
+  closedir(dirp);
+
+  using Pair = std::pair<std::unique_ptr<MctsPlayer>, Move>;
+  std::vector<Pair> puzzles;
+  for (const auto& sgf_file : sgf_files) {
+    std::ifstream ifs(file::JoinPath(FLAGS_sgf_dir, sgf_file));
```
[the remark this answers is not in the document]
chsigg: Done.

cc/main.cc
```diff
@@ -519,6 +726,73 @@ void Gtp() {
...
+  std::cerr << "DualNet factory created from " << FLAGS_model << " in "
+            << absl::ToDoubleSeconds(absl::Now() - start_time) << " sec."
+            << std::endl;
+
+  MctsPlayer::Options options;
+  ParseMctsPlayerOptionsFromFlags(&options);
+  options.verbose = false;
+
+  std::vector<std::string> sgf_files;
+  DIR* dirp = opendir(FLAGS_sgf_dir.c_str());
```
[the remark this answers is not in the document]
chsigg: Done.

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
...
+    std::unique_ptr<DualNetFactory> factory;
+    std::atomic<int> black_wins;
+    std::atomic<int> white_wins;
+  };
+
+ public:
+  void Run() {
+    auto start_time = absl::Now();
+
+    auto create_model = [](const std::string& model_path) {
```
tommadams: Why isn't this a constructor of Model?

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
...
+
+    std::cerr << "Evaluated " << num_games << " games, total time "
+              << absl::ToDoubleSeconds(absl::Now() - start_time) << " sec."
+              << std::endl;
+
+    auto print_name = [](const std::string& name) {
+      if (name.size() > 20) {
+        std::cerr << name.substr(0, 17) << "...";
+      } else {
+        std::cerr << std::setw(20) << name;
```
tommadams: absl::StreamFormat

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
...
+      }
+    };
+    auto print_wins = [&](int wins) {
+      std::cerr << "   " << std::setw(3) << wins << " " << std::setw(6)
```
tommadams: absl::StreamFormat

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
...
+    options_.random_symmetry = true;
+
+    int num_games = FLAGS_parallel_games;
+    barrier_ = absl::make_unique<Barrier>(num_games);
+
+    for (int i = 0; i < num_games;) {
+      threads_.emplace_back(std::bind(&Evaluator::ThreadRun, this, i,
+                                      cur_model.get(), prev_model.get()));
+      ++i;
+      if (FLAGS_even && i < num_games) {
```
tommadams: Instead of putting the "even" logic here in a separate lambda, why make call barrier_->Wait() at the top of ThreadRun for only the threads with odd ids? The two Wait calls would be closer in the code that way too, which aids readability.

cc/main.cc
chsigg: It's not a constructor because I wanted to keep it a struct, but maybe I'm interpreting the cpp-style too strictly that structs should not contain any functionality.

cc/main.cc
chsigg: That's what I had initially, but I didn't like that the thread id value implies a certain behavior. How about passing a bool whether to Wait() before the first move?

cc/main.cc
```diff
@@ -12,10 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <dirent.h>
```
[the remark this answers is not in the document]
chsigg: Done.

cc/main.cc
```diff
@@ -12,10 +12,13 @@
...
 #include <stdio.h>
 #include <unistd.h>
 #include <cstring>
+#include <fstream>
```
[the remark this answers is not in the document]
chsigg: Done.

cc/main.cc
```diff
@@ -12,10 +12,13 @@
...
 #include <functional>
+#include <iomanip>
```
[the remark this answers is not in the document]
chsigg: Done.

cc/main.cc
```diff
@@ -65,6 +69,8 @@ DEFINE_uint64(seed, 0,
               "Random seed. Use default value of 0 to use a time-based seed. "
               "This seed is used to control the moves played, not whether a "
               "game has resignation disabled or is a holdout.");
+DEFINE_bool(even, true,
```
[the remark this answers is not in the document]
chsigg: Removed.

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
   uint64_t flags_timestamp_ = 0;
 };
 
+class Evaluator {
+  // A barrier that blocks threads until the number of waiting threads reaches
+  // the 'count' threshold. Allows decrementing the threshold to handle the tail
+  // of a work queue where some threads exit early.
+  class Barrier {
```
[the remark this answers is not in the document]
chsigg: Done.

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
...
+    explicit WrappedDualNet(const std::unique_ptr<DualNet>& dual_net)
+        : dual_net_(dual_net) {}
+
+   private:
+    void RunMany(std::vector<const BoardFeatures*> features,
+                 std::vector<Output*> outputs, std::string* model) override {
+      dual_net_->RunMany(std::move(features), std::move(outputs), model);
+    };
+
+    const std::unique_ptr<DualNet>& dual_net_;
```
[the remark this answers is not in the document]
chsigg: Tip of the week 116 suggests to use a reference but pass a pointer to the c'tor.

cc/main.cc
```diff
@@ -464,46 +473,244 @@ class SelfPlayer {
...
+    size_t count_ GUARDED_BY(&mutex_);
+    size_t num_waiting_ GUARDED_BY(&mutex_);
+    size_t generation_ GUARDED_BY(&mutex_);
+  };
+
+  // References a pointer to an actual DualNet. Allows updating the pointer
+  // after the MctsPlayer has been constructed.
+  class WrappedDualNet : public DualNet {
+   public:
+    explicit WrappedDualNet(const std::unique_ptr<DualNet>& dual_net)
```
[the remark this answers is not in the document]
chsigg: Done.

I don't think the current design is very broken, but admittedly it is inconsistent and restrictive / specific to the intended use case.

Factory class (i.e. BatchingDualNet): DualNet instances created through the factory are not independent and can block each other if no inferences are requested. Creating DualNets is cheap.

Factory methods (Tf/Trt/TpuDualNets): DualNets are independent. Creating an instance is expensive.

I don't have a good idea for a clean design. I've tried previously to make the distinction between the two by not making the batching thing a DualNet, but a Client. But that seemed to have caused more problems than it solved.

cc/main.cc
chsigg: Done.

cc/main.cc
chsigg: Done.

cc/main.cc
chsigg: I went this route now.

cc/main.cc
tommadams: Yeah, I can understand that. I have no objection to structs having a constructor.

tommadams:
I think this looks good but because of the overlap between this and #498 I'm going to wait for that PR to be finished and merged before merging this one.

chsigg:
> I think this looks good but because of the overlap between this and #498 I'm going to wait for that PR to be finished and merged before merging this one.

Absolutely, I will rebase this PR and #523 once #498 is in.

PR #512, opened 2018-10-17
Run on Cloud TPU from C++

tommadams:
Additionally:
- Adds `dual_net.freeze_graph_tpu` method that adds the following functionality to the graph
  - TPU init & shutdown ops.
  - Replication on multiple TPU cores.
- Adds a `freeze_graph.py` command line utility.


cc/dual_net/factory.cc
```diff
@@ -50,56 +61,76 @@ DEFINE_string(engine, MG_DEFAULT_ENGINE,
...
+    dual_net = NewLiteDualNet(model_path);
 #else
     MG_FATAL() << "Binary wasn't compiled with lite inference support";
 #endif  // MG_ENABLE_LITE_DUAL_NET
   }
 
+  if (FLAGS_engine == "tpu") {
+#ifdef MG_ENABLE_TPU_DUAL_NET
+    int inferences = FLAGS_virtual_losses * FLAGS_parallel_games;
+    int buffering = FLAGS_parallel_games == 1 ? 1 : 2;
```
amj: i don't think this is needed -- you can double buffer on a single game if you've got vlosses = 2* inference batch size.

tommadams: done

PR #515, opened 2018-10-18
Bigtable input available for training

gitosaurus:
Ability to sample moves from last N games in table.
This is a preview, for discussion and for experimenting;
right now it fixes N=500k, and the probability; these
should be parameters.

Google Cloud Bigtable package now required.

### DIFF
--- a/bigtable_input.py
+++ b/bigtable_input.py
@@ -1,0 +1,137 @@
+"""Read Minigo game examples from a Bigtable.
+"""
+
+import os
+import struct
+from google.cloud import bigtable
+from google.cloud.bigtable import row_filters
+import tensorflow as tf
+import utils
+
+
+# TODO(president.jackson): document these
+_project_name = os.environ['PROJECT']
+_instance_name = os.environ['CBT_INSTANCE']
+_table_name = os.environ['CBT_TABLE']
+
+_bt_table = bigtable.Client(
+    _project_name).instance(
+        _instance_name).table(
+            _table_name)
+
+_tf_table = tf.contrib.cloud.BigtableClient(
+    _project_name,
+    _instance_name).table(
+        _table_name)
+
+
+def get_latest_game_number():
+    """Return the number of the next game to be stored.
+
+    The state of the table is stored in `metadata:game_counter`
+    in the `table_state` row only.
+    """
+    table_state = _bt_table.read_row(
+        'table_state',
+        filter_=row_filters.ColumnRangeFilter(
+            'metadata', b'game_counter', b'game_counter'))
+    value = table_state.cells['metadata'][b'game_counter'][0].value
+    return struct.unpack('>q', value)[0]
+
+
+def get_game_range_row_names(game_begin, game_end):
+    """Get the row range containing the given games.
+
+    Sample row name:
+      g_0000000001_m001
+
+    To capture the range of all moves in the two given games,
+    the end row will need to go up to g_00..(N+1).
+
+    Args:
+      game_begin:  an integer of the beginning game number.
+      game_end:  an integer of the ending game number, inclusive.
+
+    Returns:
+      The two string row numbers to pass to Bigtable as the row range.
+    """
+    row_fmt = 'g_{:0>10}_'
+    return row_fmt.format(game_begin), row_fmt.format(game_end + 1)
+
+
+def get_unparsed_moves_from_last_n_games(n,
+                                         moves=2**21,
+                                         column_family='tfexample',
+                                         column='example'):
+    """Get a dataset of serialized TFExamples from the last N games.
+
+    Args:
+      n:  an integer indicating how many past games should be sourced.
+      moves:  an integer indicating how many moves should be sampled
+        from those N games.
+      column_family:  name of the column family containing move examples.
+      column:  name of the column containing move examples.
+
+    Returns:
+      A dataset containing no more than `moves` examples, sampled
+        randomly from the last `n` games in the table.
+    """
+    latest_game = int(get_latest_game_number())
+    utils.dbg('Latest game: %s' % latest_game)
+    if latest_game == 0:
+        raise ValueError('Cannot find a latest game in the table')
+
+    start = int(max(0, latest_game - n))
+    start_row, end_row = get_game_range_row_names(start, latest_game)
+    # NOTE:  Choose a probability high enough to guarantee at least the
+    # required number of moves, by using a low estimate of the number
+    # of moves per game.  Some experiments show that it's right around
+    # 250 moves/game on average, so 100 is a conservative lower bound.
+    estimate = 100
+    probability = moves / (n * estimate)
+    utils.dbg('Row range: %s - %s; probability %.3f' % (
+        start_row, end_row, probability))
+    ds = _tf_table.parallel_scan_range(start_row, end_row,
+                                       probability=probability,
+                                       columns=[(column_family, column)])
+    # Clip the dataset to the desired number of moves, counting on
+    # the fact that the probability chosen will produce at least the
+    # number of moves required.
+    # TODO: Randomize ordering if it isn't already.
+    ds = ds.take(moves)
+    return ds.map(lambda row_name, s: s)
+
+
+def count_elements_in_dataset(ds, batch_size=1*1024, parallel_batch=8):
+    """Count and return all the elements in the given dataset.
+
+    Debugging function.  The elements in a dataset cannot be counted
+    without enumerating all of them.  By counting in batch and in
+    parallel, this method allows rapid traversal of the dataset.
+
+    Args:
+      ds:  The dataset whose elements should be counted.
+      batch_size:  the number of elements to count a a time.
+      parallel_batch:  how many batches to count in parallel.
+
+    Returns:
+      The number of elements in the dataset.
+    """
+    with tf.Session() as sess:
+        dsc = ds.apply(tf.contrib.data.enumerate_dataset())
+        dsc = dsc.apply(
+            tf.contrib.data.map_and_batch(lambda c, v: c, batch_size,
+                                          num_parallel_batches=parallel_batch))
+        iterator = dsc.make_initializable_iterator()
+        sess.run(iterator.initializer)
+        get_next = iterator.get_next()
+        try:
+            while True:
+                # The numbers in the tensors are 0-based indicies,
+                # so add 1 to get the number counted.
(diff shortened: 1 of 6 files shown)


bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+            _table_name)
+
+_tf_table = tf.contrib.cloud.BigtableClient(
+    _project_name,
+    _instance_name).table(
+        _table_name)
+
+
+_features = {
+    'x': tf.FixedLenFeature((), tf.string, ''),
```
sethtroisi: Can you add "TODO(sethtroisi): Unify with preprocessing"

bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+    'pi': tf.FixedLenFeature((), tf.string, ''),
+    'outcome': tf.VarLenFeature(dtype=tf.float32)
+    }
+
+
+def get_latest_game_number():
+  """Return the number of the next game to be stored.
+  """
```
sethtroisi: nit: can all fit on one line

bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+                                         column_family='tfexample',
+                                         column='example'):
+  latest_game = int(get_latest_game_number())
+  logging.info("Latest game: %s", latest_game)
+  if latest_game == 0:
+    return 0
+
+  start = int(max(0, latest_game - n))
+  start_row, end_row = get_game_range_row_names(start, latest_game)
+  logging.info("Row range: %s - %s", start_row, end_row)
```
sethtroisi: Is it possible to count how many rows match this row range and print that number as a double check we have good n, probability?

bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+  return row_fmt.format(game_begin), row_fmt.format(game_end + 1)
+
+
+def get_unparsed_moves_from_last_n_games(n, probability,
+                                         column_family='tfexample',
+                                         column='example'):
+  latest_game = int(get_latest_game_number())
+  logging.info("Latest game: %s", latest_game)
+  if latest_game == 0:
+    return 0
```
sethtroisi: probably more useful to throw an exception

bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+def get_moves_from_last_n_games(n, probability,
+                                column_family='tfexample',
+                                column='example'):
+  ds = get_unparsed_moves_from_last_n_games(n, probability,
+                                            column_family, column)
+  return ds.map(tf.parse_single_example)
+
+
+def realize_dataset(ds):
+  iterator = ds.make_initializable_iterator()
```
sethtroisi: Is this for debug? If so please give it a docstring explaining

preprocessing.py
```diff
@@ -232,6 +232,27 @@ def get_tpu_input_tensors(batch_size, tf_records, num_repeats=1,
     return dataset
 
 
+def get_tpu_bt_input_tensors(batch_size, num_repeats=1,
+                             shuffle_records=True, shuffle_examples=True,
+                             shuffle_buffer_size=1024,
```
sethtroisi: for now lets drop all the parameters except batch_size and random_rotation

bigtable_input.py
gitosaurus: It is possible, but requires materializing the request. (Even asking for a small field like `move`, which is partly why that is there, still involves making all the row requests.) I can do it under a debug flag so that we will have the check available, but not by default in production.

bigtable_input.py
gitosaurus: Agree. Will raise a `ValueError` instead.

bigtable_input.py
gitosaurus: It definitely is! (See the above problem of counting rows.) I'll document it as such. Also, on reflection, it's better to call it `materialize_dataset`, which I will do.

preprocessing.py
gitosaurus: Can do. Note that I'm also ignoring `random_rotation` in the present implementation because of the problem we saw (which I would still like to get fixed before submitting).

bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+    'pi': tf.FixedLenFeature((), tf.string, ''),
+    'outcome': tf.VarLenFeature(dtype=tf.float32)
+    }
+
+
+def get_latest_game_number():
+  """Return the number of the next game to be stored.
+  """
+  table_state = _bt_table.read_row('table_state')
+  value = table_state.cells['metadata'][b'game_counter'][0].value
```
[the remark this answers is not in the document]
gitosaurus: It is worth it; thanks for the catch. (At present, the `tfexample:example` column is the largest, and the one needed for training, so excluding `metadata:move` doesn't buy a lot, but being specific is the right thing to do.)

bigtable_input.py
gitosaurus: Will do. In fact, now that I consider it, I'm not sure it's necessary to have `_features` in here, since at present, the Bigtable client simply passes back strings, and the parsing happens up in `preprocessing.py` anyway.

bigtable_input.py
```diff
@@ -0,0 +1,91 @@
...
+  return struct.unpack('>q', value)[0]
+
+
+def get_game_range_row_names(game_begin, game_end):
+  """Get the row range containing the given games.
+
+  Sample row name:
+    g_0000000001_m000001
```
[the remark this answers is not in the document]
gitosaurus: Thanks! Will reclaim that bit of key space.

bigtable_input.py
gitosaurus: I did some timing this afternoon to see how fast the rows in a dataset can be counted. The answer is ~50k/second.

amj: for some numbers here:  Training a new model is ~7min-15min, so taking an extra 10 seconds to get our ducks in a row is not a problem.

bigtable_input.py
gitosaurus: Will do. (Or will expand the comment.)

gitosaurus: Will do. (Or will expand the comment.)

bigtable_input.py
gitosaurus: Unfortunately, that 50k/second speed is for row keys (moves), not games. There are on average 250 moves/game, so the total scan time becomes (500e3 g * 250 m/g) / 50e3 m/s = 2500s, or about 40 minutes, if we were to verify the exact move count each time. So I think that using the 250 moves/game estimate, bumping up the probability a little, and then using `take()` to precisely limit it remains the best approach for now.

bigtable_input.py
gitosaurus: Thanks. Will fix (or expand the docstring content).

bigtable_input.py
```diff
@@ -0,0 +1,135 @@
...
+
+import logging
+import os
+import struct
+from google.cloud import bigtable
+from google.cloud.bigtable import row_filters
+import tensorflow as tf
+
+
+_project_name = os.environ['PROJECT']
```
sethtroisi: can you add "TODO(president.jackson): document these"

bigtable_input.py
```diff
@@ -0,0 +1,135 @@
...
+        _table_name)
+
+
+def get_latest_game_number():
+    """Return the number of the next game to be stored.
+
+    The state of the table is stored in `metadata:game_counter`
+    in the `table_state` row only.
+    """
+    counter = b'game_counter'
```
sethtroisi: can this be inlined as a string like 'table_state' and 'metadata'?

bigtable_input.py
```diff
@@ -0,0 +1,135 @@
...
+        from those N games.
+      column_family:  name of the column family containing move examples.
+      column:  name of the column containing move examples.
+
+    Returns:
+      A dataset containing no more than `moves` examples, sampled
+        randomly from the last `n` games in the table.
+    """
+    latest_game = int(get_latest_game_number())
+    logging.info('Latest game: %s', latest_game)
```
sethtroisi: This should never be invoked from GTP so it's fine to maybe fine to log to stdout but should we keep this uniform and log with util.dbg?

bigtable_input.py
```diff
@@ -0,0 +1,135 @@
...
+    if latest_game == 0:
+        raise ValueError('Cannot find a latest game in the table')
+
+    start = int(max(0, latest_game - n))
+    start_row, end_row = get_game_range_row_names(start, latest_game)
+    # Choose a probability high enough to guarantee at least the
```
sethtroisi: nit: "NOTE: Choose a " makes these easier to grep for

bigtable_input.py
```diff
@@ -0,0 +1,135 @@
...
+    probability = moves / (n * estimate)
+    logging.info('Row range: %s - %s; probability %.3f',
+                 start_row, end_row, probability)
+    ds = _tf_table.parallel_scan_range(start_row, end_row,
+                                       probability=probability,
+                                       columns=[(column_family, column)])
+    # Clip the dataset to the desired number of moves, counting on
+    # the fact that the probability chosen will produce at least the
+    # number of moves required.
+    ds = ds.take(moves)
```
sethtroisi: TODO: Randomize ordering.

train.py
```diff
@@ -39,6 +39,8 @@
 flags.DEFINE_string('export_path', None,
                     'Where to export the model after training.')
 
+flags.DEFINE_bool('use_bt', False, 'Whether to use Bigtable as input.')
```
sethtroisi: can you add something like "only supported with TPU currently"

bigtable_input.py
gitosaurus: Done!

bigtable_input.py
gitosaurus: The reason I didn't is because it's referenced 3 times in the code, and does need to be specified as a byte string (unlike the column family and row name, oddly). I don't have a problem with expanding it in this case, since the scope is so small.

bigtable_input.py
gitosaurus: No problem with using `utils.dbg` throughout. Changed.

bigtable_input.py
gitosaurus: Done.

bigtable_input.py
gitosaurus: TODO added.

train.py
gitosaurus: Done!

PR #519, opened 2018-10-19
Remove obsolete dependencies

sethtroisi:
httplib2 + oauth2client were from https://github.com/tensorflow/minigo/commit/eb6ecc07c4f1ab8786f36f86cb52f5a7cc43742d which has been mostly removed

argh was replaced by fire in: https://github.com/tensorflow/minigo/commit/8ed08c8ddd6a03ac602fc011002e1471f34738c4

## Unsure

grpcio-tools was from remote inference? I don't think it's required for TPUs (but I might be wrong)

Derek added argcomplete in https://github.com/tensorflow/minigo/commit/64c992266be546436e6654a32655e9ed22c0a4cc

~keras doesn't match any greps, but tom added [10 days ago](https://github.com/tensorflow/minigo/commit/db886c8ac5ee61169a71653e00653704328dd671) in "a couple of fixes"~ - Needed for tf 1.11


tommadams:
keras is needed by cc/configure_tensorflow.sh to build tf 1.11

amj:
can we check grpcio-tools wasn't needed for tpu selfplay before we check this in?

PR #522, opened 2018-10-19
Add superko support

### LINKED ISSUE #517: Switch to absl hashed containers
absl::flat_hash_map and absl::flat_hash_set were finally released a couple of weeks ago: https://abseil.io/blog/20180927-swisstables

We should use them instead of their std library counterparts.

tommadams:
Fixes #517 while I was touching these files.


cc/mcts_node.cc
```diff
@@ -314,4 +367,18 @@ MctsNode* MctsNode::MaybeAddChild(Coord c) {
     return it->second.get();
   }
 }
+
+bool MctsNode::HasPositionBeenPlayedBefore(zobrist::Hash stone_hash) const {
+  for (const auto* node = this; node != nullptr; node = node->parent) {
+    if (node->superko_cache != nullptr) {
+      return node->superko_cache->count(stone_hash);
```
sethtroisi: s/count/contains/

cc/mcts_node.h
```diff
@@ -143,6 +146,20 @@ class MctsNode {
...
+  // reduces superko detection time complexity from O(N) to O(1).
+  //
+  // If non-null, superko_cache contains the Zobrist hash of all positions
+  // played to this position, including position.stone_hash().
+  // If null, clients should determine whether a position has appeared before
+  // during the game by walking up the tree (via the parent pointer), checking
+  // the position.stone_hash() of each node visited, until a node is found that
+  // contains a non-null superko_cache.
+  using SuperkoCache = absl::flat_hash_set<zobrist::Hash>;
+  std::unique_ptr<SuperkoCache> superko_cache;
```
sethtroisi: This is going to use a fair bit more memory.

At depth = 200
1/8 * 200 * 8bytes = 200 bytes more per node

cc/zobrist.cc
```diff
@@ -0,0 +1,41 @@
...
+std::array<std::array<Hash, 3>, kNumMoves> kMoveHashes;
+std::array<Hash, kN * kN> kKoHashes;
+
+void Init(uint64_t seed) {
+  minigo::Random rnd(seed);
+
+  kBlackToPlayHash = rnd.UniformUint64();
+  for (auto& x : kMoveHashes) {
+    for (size_t i = 0; i < x.size(); ++i) {
+      x[i] = i == 0 ? 0 : rnd.UniformUint64();
```
sethtroisi: bad c++ suggestion.

`(i == 0) * rnd.UniformUint64()`

cc/mcts_node.h
tommadams: The important factor is the depth of the tree search, since the tree above the current position is essentially a linked list.

With a branching factory of 1.5, a search depth of 200 would be ~10^35 reads, at which point we probably have bigger problems than a 10% increase in memory consumption.

cc/zobrist.cc
tommadams: I'll leave it to the compiler to find the optimal way to generate the code :P

cc/mcts_node.cc
tommadams: Wow, C++20 features, today!

Thanks, absl. Thabsl.

cc/gtp_player.cc
```diff
@@ -405,7 +380,18 @@ GtpPlayer::Response GtpPlayer::HandleLoadsgf(absl::string_view cmd,
...
-    PlayMove(move.c);
+    if (!root()->legal_moves[move.c]) {
+      return Response::Error("illegal move");
+    }
+
+    // Perform a single inference for each move with random symmetry disabled so
+    // that the same model will produce the same result every time we load the
+    // same SGF.
+    auto* leaf = root()->MaybeAddChild(move.c);
+    ProcessLeaves({&leaf, 1}, false);
```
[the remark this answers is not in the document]
tommadams: Probably. I've added a TODO.

cc/mcts_node.h
sethtroisi: It's not search depth that matters but number of moves played in the game

tommadams: How so?

Every time we play a move, we delete all the child nodes in the search tree except those in the subtree of the move that was played.

Ignoring the depth of the search tree, a game of 200 moves will consequently only hold 200/8 == 50 caches in the ancestor nodes of the final position's node. That's 50*100*8 == 40kB of data, since the average size of the cache will be 100.

Am I missing something? I'd like to know if my estimations are way off base!

cc/position.cc
```diff
@@ -268,6 +274,7 @@ void Position::RemoveGroup(Coord c) {
 
     MG_CHECK(stones_[c].group_id() == removed_group_id);
     stones_[c] = {};
+    stone_hash_ ^= zobrist::MoveHash(c, removed_color);
```
brilee: There's a cute trick you can do here - have each group remember the zobrist hash of each of its stones. Since zobrist hashing is associative, you can update the hash with one XOR by XORing against the hash of the entire group, instead of XOR'ing through each individual stone.

Although, now that I think about it... that also requires incrementally updating the group's hash, so it might just end up as a wash. *shrug*

cc/mcts_node.h
brilee: Q: If we're in move 200 of the game, and there are 25 caches spaced out at moves 8, 16, 24... then the caches from 8...192 are unused because all tree searches from move 201+ will back up to 200 and then not look any further. So it seems like we shouldn't worry about those caches.

Instead, I'm wondering if we branch by factor 1.5, then at 10,000 reads, that would imply a search depth of ~20, with the number of nodes at search depth 8 being ~25 nodes, and the number of nodes at search depth 16 being ~650. Do each of the search nodes at depth 16 get their own cache?

cc/mcts_node.cc
```diff
@@ -41,11 +83,24 @@ MctsNode::MctsNode(MctsNode* parent, Coord move)
       move(move),
       position(parent->position) {
   position.PlayMove(move);
-  // TODO(tommadams): Only call IsMoveLegal if we want to select a leaf for
-  // expansion.
-  for (int i = 0; i < kNumMoves; ++i) {
-    illegal_moves[i] = !position.IsMoveLegal(i);
+
+  // Insert a cache of ancestor Zobrist hashes at regular depths in the tree.
```
brilee: A cache's cost grows linearly with position.n(); while a cache's benefit is equal to [number of times any child of this node is visited] * [distance between this cache and the previous cache].

So from an optimality perspective, it seems like caches should become less frequent as the game goes on longer. And there's probably some heuristic, like, "if a node has been visited X times and we're on readout 160 of 1600, then, we can probably guess that this node will be visited 10X times by the time tree search completes", and then dynamically insert a cache based on that estimate.

PR #523, opened 2018-10-21
Remove batch_size flag.

chsigg:
Hi Tom, Here is an attempt to remove the batch_size flag. Instead, each DualNet can determine it's batch size based on virtual_losses and num_parallel_games.


chsigg:
Somehow GitHub is not picking up all my commits pushed to my chsigg/minigo-fork batch_size branch. Or it takes a long time. There should be 4 commits in the PR, the last one called 'Missed some changes to TpuDualNet.'.

/retest

cc/dual_net/lite_dual_net.cc
```diff
@@ -73,16 +75,30 @@ minigo::LiteDualNet::LiteDualNet(std::string graph_path)
...
+    return;
+  }
 
   // Initialize input.
   const auto& inputs = interpreter_->inputs();
   MG_CHECK(inputs.size() == 1);
   absl::string_view input_name = interpreter_->GetInputName(0);
   MG_CHECK(input_name == "pos_tensor");
 
+  // Resize input tensor to batch size.
```
tommadams: I think you can move most of the checks into the constructor.
The  only checks you need in `Reserve` are something like:

```c++
  MG_CHECK(interpreter_->ResizeInputTensor(
               inputs[0], {static_cast<int>(capacity), kN, kN,
                           DualNet::kNumStoneFeatures}) == kTfLiteOk);
  MG_CHECK(interpreter_->AllocateTensors() == kTfLiteOk);
  MG_CHECK(input_->dims->data[0] == capacity);
  MG_CHECK(policy_->dims->data[0] == capacity);
  MG_CHECK(value_->dims->data[0] == capacity);
```

We can probably get rid of the `data.raw` checks, since presumably the fact that `ResizeInputTensor` succeeded guarantees that they're non-null.

cc/dual_net/tpu_dual_net.cc
```diff
@@ -158,11 +165,12 @@ TpuDualNet::TpuDualNet(const std::string& graph_path,
 
   // Run warm-up inferences on all sessions.
   // Tensorflow lazily initializes the first time Session::Run is called,
-  // which can take hundreds of milliseconds. This intefers with time control,
+  // which can take hundreds of milliseconds. This interfers with time control,
   // so explicitly run inference once during construction.
+  // TODO(tommadams): Is this still necessary without 'time control'?
```
tommadams: I'm not sure what you mean by this comment. Time control is a feature of the [MctsPlayer](https://github.com/tensorflow/minigo/blob/ae1eec2e40fc4d73616bf000192a1255d8f3c793/cc/mcts_player.cc#L135) and has nothing to do with DualNet.

chsigg: Aha, I put 'time control' in quotes because I didn't really understand what it refers to. Now it makes sense, thanks. Removed.

cc/dual_net/lite_dual_net.cc
tommadams: No. Most of those checks are not necessary after changing the batch size.

The reason the width, height & depth checks were put in the constructor originally is to make sure the model being loaded matches expectations, e.g. you're not trying to load a 9x9 model into an engine compiled for 19x19 games.

Changing the batch size will not change the width, height or depth. Having those checks immediately after calling ResizeInputTensor suggests that's a possibility, which is incorrect.

Checking that the number of outputs a model has remains constant after calling ResizeInputTensor just doesn't make any sense.

chsigg:
Am I crazy or does GitHub drop stuff? I can't find the following comment anymore from Tom:

> No. Most of those checks are not necessary after changing the batch size.
>
> The reason the width, height & depth checks were put in the constructor originally is to make sure the model being loaded matches expectations, e.g. you're not trying to load a 9x9 model into an engine compiled for 19x19 games.
>
> Changing the batch size will not change the width, height or depth. Having those checks immediately after calling ResizeInputTensor suggests that's a possibility, which is incorrect.
>
> Checking that the number of outputs a model has remains constant after calling ResizeInputTensor just doesn't make any sense.

I changed it to get the tensor indices in the c'tor and do a minimal test on the input to verify that the model matches the board size.

I've removed the capacity checks, even though I feel that they did provide some extra safety. TfLite has surprising and not well documented behavior that after AllocateTensors():

- The input tensor has the new size, but data is null. You need to get the new input from the interpreter.
- The output tensors have a completely different size. Might be just a dangling pointer, also need to get them from the interpreter again. The new outputs have the updated size (matching the input size for our graph).

And all that probably only if the tensor size has changed, because the old code called AllocateTensors() after getting inputs and outputs.

amj:
i think the comment is pinned to the older commit, so it's hidden under the 'show outdated' fold.

PR #528, opened 2018-10-23
Run pylint3 even if venv is not activated

sethtroisi:
Help me stop punching myself


sethtroisi:
pylint3 is not defined in the image :( let me try and understand why.

test.sh
```diff
@@ -25,7 +25,7 @@
 cd "$(dirname "$0")"
 
 lint_fail=0
-pylint *.py || {
+python3 -m pylint *.py || {
```
sethtroisi: I'd rather write pylint3 but this appears not to work in the docker image because of weird things.

amj: can you elaborate on the 'weird things' ?  Or leave a comment?

PR #534, opened 2018-10-28
Optionally depend on TensorFlow through bazel.

chsigg:
By default, build TF as bazel dependency of anything under cc/. 

cc/configure_tensorflow.sh asks whether you want to build TF and save the binaries, to get the previous behavior of building TF just once.

Simplifies and checks configuration, makes it safer to share dependencies with TensorFlow.

It requires a bit of Starlark magic to offer both pre-building TensorFlow as well as depending on it through bazel.

This PR affects cc/... only, i.e. the python code still uses the installed TensorFlow.

Sticking with TF 1.11 and bazel 0.17.


chsigg:
> Upgrades to TF 1.12-rc2. Do I need to change other bits for this?
It would also have required bazel 1.18. Too many moving parts. I reworked it to support TF 1.11 and bazel 1.17 as before.

tommadams:
It's great that using TF as an external dependency works now, but can you explain the motivation for this PR please?

We are well aware of the short comings of cc/configure_tensorflow.sh, however I think I've clearly explained our motivation for having it to you before: If I'm a user with a 4 core machine, and I want to change some build flags for whatever reason (debugging, profiling, enabling optimizations), I don't want to have to wait for literally hours for Bazel to recompile TensorFlow.

In an ideal world, optionally using Bazel to build TensorFlow would be great. However, this PR demonstrably adds a significant maintenance burden: it breaks our current test infrastructure and apparently makes it harder to upgrade TF versions. Before we take on the burden for a feature that we haven't historically needed, I'd like to understand whether the benefits outweigh the costs.

chsigg:
Andrew pointed out that it wasn't clear from the PR description that you can prebuild TF just like before ("offer both pre-building TensorFlow as well as depending on it through bazel"). 

You run cc/configure_tensorflow.sh as before to configure TF. Then it asks you if you want to build it now, you hit Enter and everything behaves exactly like before. TensorFlow is built with -c opt and tucked away under cc/tensorflow and bazel will pick it from there and not ever build it again until you run configure_tensorflow.sh again.

tommadams:
The fact that you initially attempted to upgrade to TF 1.12 in this PR and then gave up suggested that it _does_ in fact make upgrading TF versions harder but I don't know what problems you ran into, so I was just inferring from your own PR comments.

I understand that this is a optional alternative to cc/configure_tensorflow. That's the thrust of my objection to the PR: it creates additional maintenance burden. The next time we want to upgrade to a later version of TF, we'll have to test both methods of building TF and fix the problems that inevitably show up twice.

Andrew told me you had some concerns about potential ODR violations, which I agree could be a problem. However, we've never seen any symptoms of an ODR violation and as Titus pointed out in his recent [keynote at Pacific++](https://youtu.be/IY8tHh2LSX4), every C++ codebase that use prebuilt libraries has the same problem.

If we had infinite time & resources, I'd happily accept this PR but we don't. So I shall ask again: before we accept the extra maintenance burden associated with this PR, can you please explain what problems it solves?

chsigg:
It's not that I gave up, I wanted to avoid having to upgrade to TF 1.12. It would have made this change easier because TF 1.12 is easier to configure from an inverse dependency. Same for updating to bazel 0.18. With some extra work this change can stay with TF 1.11 but it does require bazel 0.15, I think. I've only been able to test with 0.17 locally. I haven't figured out what bazel versions are installed in the images. The documentation for C++ Minigo says it requires bazel 0.17, but I think the test image uses an older version.

Maintenance burden: configure_tensorflow.sh builds exactly what would otherwise be the TF dependency. In other words, if you test the prebuilt path and keep it working, the bazel dependency path is covered as well.

ODR violations: google3 and others have this issue, yes. Bazel effectively avoids different versions of common dependencies, which is just one thing that can go wrong. But it clearly doesn't hurt to have that covered.

Benefits: 
- Only build the parts of TensorFlow needed for C++ MiniGo, not the entire wheel. tensorflow_configure.sh takes 11min 42sec instead of 20min on my machine.
- No need to prebuild toco. 
- Fixes incomplete headers when not building TF with CUDA.
- Adding the grpc dependency is cleaner. A similar dependency will be needed to enable XLA:GPU.
- Reuses TF configuration of CUDA and TensorRT.
- Once we have BUILD files for the python code as well, there is no need to pip install TF (in a virtual env or docker), and you can run directly from IDEs with bazel integration like CLion.
- It works with bazel 0.19 which is the current and only version on rodete. Of course, fixing master is a much smaller change.

tommadams:
Thanks for the explanation, that makes a lot of sense. In general, that's the kind of information that would be good to put in the change description: while these issues may be obvious to you, I wasn't aware of them.

Given that motivation, I'm on board with adding this PR. Any idea when TF will add support for CloudTPU to a release? I'd like to remove cc/configure_tensorflow.sh if we're going with this approach.

That said, I cloned your fork and tried building and it failed:

```
$ bazel build cc:main
DEBUG: /usr/local/google/home/tmadams/.cache/bazel/_bazel_tmadams/396ac1642031eaf41fabab3b3dfaf61c/external/bazel_tools/tools/build_defs/repo/http.bzl:46:9: patch file //cc/tensorflow:tensorflow.patch, path /usr/local/google/home/tmadams/chsigg_minigo/cc/tensorflow/tensorflow.patch
ERROR: /usr/local/google/home/tmadams/chsigg_minigo/cc/BUILD:88:1: no such target '@com_google_absl//absl/container:flat_hash_map': target 'flat_hash_map' not declared in package 'absl/container' defined by /usr/local/google/home/tmadams/.cache/bazel/_bazel_tmadams/396ac1642031eaf41fabab3b3dfaf61c/external/com_google_absl/absl/container/BUILD.bazel and referenced by '//cc:mcts'
ERROR: /usr/local/google/home/tmadams/chsigg_minigo/cc/BUILD:88:1: no such target '@com_google_absl//absl/container:flat_hash_set': target 'flat_hash_set' not declared in package 'absl/container' defined by /usr/local/google/home/tmadams/.cache/bazel/_bazel_tmadams/396ac1642031eaf41fabab3b3dfaf61c/external/com_google_absl/absl/container/BUILD.bazel and referenced by '//cc:mcts'
ERROR: Analysis of target '//cc:main' failed; build aborted: Analysis failed
```

Looks like TF v1.11 uses an old version of ABSL that doesn't contain flat_hash_map or flat_hash_set. Do you know if v1.12 is updated?

chsigg:
I don't want to interfere with the discussion whether we actually want this PR, but just lay out the plan to fix the test failure. The PR now builds with bazel 0.11, but the testing procedure still needs some changes. Specifically, it should call cc/configure_tensorflow.sh first, and the cp command to pull in the pre-built TF needs to change slightly. I could work around both by adding a default TF configuration file and using the same directory structure in cc/tensorflow, but this seems the cleaner approach:

1. Copy C++ test script from test-infra to this repo (something like https://github.com/chsigg/minigo-fork/commit/62b9da7174fe3b907316662f24897fa9025c2abb)
1. Change https://github.com/kubernetes/test-infra/blob/master/config/jobs/tensorflow/minigo/minigo.yaml to call the new script.
1. Change the script as part of this PR to run tensorflow_configure.sh and update the cp command.

cc/config/BUILD
```diff
@@ -42,11 +42,16 @@ config_setting(
...
 
 config_setting(
-    name = "enable_tpu",
-    define_values = {"tpu": "1"},
+    name = "bazel_tf",
+    define_values = {"bazel_tf": "1"},
+)
+
+config_setting(
+    name = "never",
```
tommadams: I don't see this referenced anywhere, what's it for?

chsigg: I used this initially for checking a valid configuration, like this:

select({
  if_something(
    if_true = {"//conditions:default": [...]}
    if_false = {"//cc/config:never": []}
  ),
  no_match_error = "if_something is false",
})

No longer needed, removed.

amj:
\retest

PR #535, opened 2018-10-30
Store move count per game, use shuffle buffer

gitosaurus:
In order to count moves in a range of games quickly, store the sum of
moves per game in a new `move_count` column.  They are stored both in
the zero-move row for the game, but also in a new `ct_` key space,
which can be faster to scan.  Using this new column, the exact number
of moves in 500e3 games can be calculated in 2-8 seconds.

The moves are shuffled in the input pipeline using
`tf.data.Dataset.shuffle`, with a buffer, as this preserves
performance.  Earlier work that used `lookup_columns` from Bigtable
was 100X slower.

Also fixed:

  - Off-by-one error in game range
  - Row keys must be `bytes`, not `str`

### DIFF
--- a/bigtable_input.py
+++ b/bigtable_input.py
@@ -1,8 +1,11 @@
 """Read Minigo game examples from a Bigtable.
 """
 
+import collections
+import numpy as np
 import os
 import struct
+import time
 from google.cloud import bigtable
 from google.cloud.bigtable import row_filters
 import tensorflow as tf
@@ -32,10 +35,10 @@ def get_latest_game_number():
     in the `table_state` row only.
     """
     table_state = _bt_table.read_row(
-        'table_state',
+        b'table_state',
         filter_=row_filters.ColumnRangeFilter(
             'metadata', b'game_counter', b'game_counter'))
-    value = table_state.cells['metadata'][b'game_counter'][0].value
+    value = table_state.cell_value('metadata', b'game_counter')
     return struct.unpack('>q', value)[0]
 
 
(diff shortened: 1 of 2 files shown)


bigtable_input.py
```diff
@@ -32,10 +34,10 @@ def get_latest_game_number():
     in the `table_state` row only.
     """
     table_state = _bt_table.read_row(
-        'table_state',
+        b'table_state',
         filter_=row_filters.ColumnRangeFilter(
             'metadata', b'game_counter', b'game_counter'))
-    value = table_state.cells['metadata'][b'game_counter'][0].value
+    value = table_state.cell_value('metadata', b'game_counter')
```
sethtroisi: that's a much nicer call :)

bigtable_input.py
```diff
@@ -45,22 +47,66 @@ def get_game_range_row_names(game_begin, game_end):
...
     row_fmt = 'g_{:0>10}_'
-    return row_fmt.format(game_begin), row_fmt.format(game_end + 1)
+    return row_fmt.format(game_begin), row_fmt.format(game_end)
+
+
+def make_single_array(ds, batch_size=8*1024):
+    full_result = None
+    with tf.Session() as sess:
+        ds = ds.batch(batch_size)
+        iterator = ds.make_initializable_iterator()
```
sethtroisi: why initializable_iterator vs make_one_shot_iterator()

bigtable_input.py
```diff
@@ -45,22 +47,66 @@ def get_game_range_row_names(game_begin, game_end):
...
+        sess.run(iterator.initializer)
+        get_next = iterator.get_next()
+        try:
+            while True:
+                result = sess.run(get_next)
+                if full_result is None:
+                    full_result = result
+                else:
+                    full_result = np.concatenate((full_result, result))
```
sethtroisi: in my experience this can be slow when you are constantly concatenating
maybe preallocate a big number then shrink the np.array afterwards?
I couldn't find a better link than this https://stackoverflow.com/a/13220868

bigtable_input.py
```diff
@@ -82,26 +129,76 @@ def get_unparsed_moves_from_last_n_games(n,
...
+    h = {}
+    try:
+        while True:
+            bresult = sess.run(
+                tf.reduce_join(get_next,
+                               separator='\n'))
+            result = str(bresult, 'utf-8')
+            for line in result.splitlines():
+                g = line.split('_m_')[0]
+                h[g] = h.get(g, 0) + 1
```
sethtroisi: consider
`h = defaultdict(int)`
`h[g] += 1`

bigtable_input.py
gitosaurus: Yes, I was glad to find it!

bigtable_input.py
gitosaurus: Unfortunately, this is a requirement for Bigtable iterators. Attempting to create a one-shot iterator of a Dataset constructed from a Bigtable yields the following error:

Failed to create a one-shot iterator for a dataset. `Dataset.make_one_shot_iterator()` does not support datasets that capture stateful objects, such as a `Variable` or `LookupTable`. In these cases, use `Dataset.make_initializable_iterator()`

bigtable_input.py
brilee: Collecting a list of np.array and then concatenating them is a factor of 2 overhead. Agreed that this method is O(n^2).

gitosaurus: I like the idea in the link you shared, of collecting a list of smaller arrays and then concatenating them in one call at the end. I'll do that.

bigtable_input.py
brilee: modern python is collections.Counter(iterable)

bigtable_input.py
```diff
@@ -45,22 +47,66 @@ def get_game_range_row_names(game_begin, game_end):
...
+    utils.dbg('Row range: %s - %s; total moves: %d; probability %.3f' % (
+        start_row, end_row, total_moves, probability))
+    shards = 8
+    ds = _tf_table.parallel_scan_range(start_row, end_row,
+                                       probability=probability,
+                                       num_parallel_scans=shards,
+                                       columns=[(column_family, column)])
+    if shuffle:
+        rds = tf.data.Dataset.from_tensor_slices(
+            tf.random_shuffle(tf.range(0, shards, dtype=tf.int64)))
```
brilee: I don't think shuffling the shard orders does anything if you're reading them all at the same time anyway?

bigtable_input.py
gitosaurus: Considered! And after a brief moment of profiling, I'm convinced. `defaultdict` is measurably faster than my idiom. Thanks!

bigtable_input.py
gitosaurus: Good point. When I was testing, I alternated between `Dataset.interleave` and `tf.contrib.data.parallel_interleave`, so it was important to shuffle the shards. And even in this case, although I believe you're right that the parallel interleave will likely not be ordered, it doesn't cost anything to intentionally disorder it here. (After all, `parallel_scan_range` itself doesn't promise ordered results, and they're not, but they're not disordered _enough_, hence this update.)

Issue #520, opened 2018-10-19
Fill shuffle buffer very slow


delock:
When training with shuffle buffer size 100,000, it take a long time to fill us shuffle buffer, it can take up to 1 minute for each training epoch.

`
INFO:tensorflow:Saving checkpoints for 0 into /home/gma/results/minigo/final/estimator_working_dir/model.ckpt.                                                              
2018-10-15 00:01:20.219185: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:98] Filling up shuffle buffer (this may take a while): 23398 of 100000                     
2018-10-15 00:01:30.217862: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:98] Filling up shuffle buffer (this may take a while): 47305 of 100000                     
2018-10-15 00:01:40.218016: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:98] Filling up shuffle buffer (this may take a while): 70837 of 100000                     
2018-10-15 00:01:50.218091: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:98] Filling up shuffle buffer (this may take a while): 94768 of 100000                     
2018-10-15 00:01:52.428082: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:136] Shuffle buffer filled.                                                                
`

After some investigation, it turns out in preprocessing.py, a FilterDataset is used to sample input data:

`
dataset = dataset.filter(lambda x: tf.less(
    tf.random_uniform([1]), filter_amount)[0])  
`

When running this FilterDataset, it would run the predicate repeatedly, which is very slow.

I propose to introduce a 'SamplingDataset' to TensorFlow, which accepts a rate as parameter, and do sampling based on an internal random number generator.   It would greatly reduce the time spend on filling shuffle buffer.

brilee:
Adding a new op to TensorFlow itself is outside the scope of the MiniGo project.

That being said - do you have profiler evidence to suggest that the inline filter is what is causing the slowness, rather than having to do the I/O for 20 examples for every 1 that ends up being used?

delock:
I didn't successfully get profiling data with the following profiling code:
```
     with tf.contrib.tfprof.ProfileContext('/tmp/train_dir') as pctx:
         main.train(ESTIMATOR_WORKING_DIR, TRAINING_CHUNK_DIR, save_file,
                generation_num=model_num + 1)
```
[...]

PR #536, opened 2018-10-30
Better handle multiple golden chunks in TPU train

sethtroisi:


preprocessing.py
```diff
@@ -205,17 +205,20 @@ def get_input_tensors(batch_size, tf_records, num_repeats=1,
 
 
 def get_tpu_input_tensors(batch_size, tf_records, num_repeats=1,
-                          shuffle_records=True, shuffle_examples=True,
-                          shuffle_buffer_size=1024,
                           filter_amount=1, random_rotation=True):
+    # TPUs trains on sequential golden chunks to simplify preprocessing and
+    # reproducability.
```
brilee: typo:reproducibility

preprocessing.py
```diff
@@ -119,25 +119,24 @@ def read_tf_records(batch_size, tf_records, num_repeats=1,
...
-                                         tf.data.TFRecordDataset(
-                                             x, compression_type='ZLIB'),
-                                         cycle_length=64, block_length=16)
+        dataset = record_list.flat_map(map_func)
+
     if filter_amount < 1.0:
-        dataset = dataset.filter(lambda x: tf.less(
-            tf.random_uniform([1]), filter_amount)[0])
+        dataset = dataset.filter(
+            lambda _: tf.random_uniform([]) < filter_amount)
```
brilee: does this actually do shape inference to find the right size for tf.random_uniform? That seems pretty cool. 

Curious if this also fixes #520

sethtroisi: I think it does something different than you expect.
[] is a scalar random and filter get's run on every item in the dataset.

---
```
x = np.random.sample((10, 2)) 
dataset = tf.data.Dataset.from_tensor_slices(x)
iter = dataset.make_one_shot_iterator().get_next()
with tf.Session() as sess:
  while True:
    print(sess.run(iter))
iter = dataset.filter(lambda x: tf.random_uniform([]) < .3).make_one_shot_iterator().get_next()
with tf.Session() as sess:
  while True:
    print(sess.run(iter))
```

brilee: oh, okay - never mind then :)

preprocessing.py
```diff
@@ -205,17 +205,20 @@ def get_input_tensors(batch_size, tf_records, num_repeats=1,
 
 
 def get_tpu_input_tensors(batch_size, tf_records, num_repeats=1,
-                          shuffle_records=True, shuffle_examples=True,
-                          shuffle_buffer_size=1024,
                           filter_amount=1, random_rotation=True):
+    # TPUs trains on sequential golden chunks to simplify preprocessing and
```
amj: This isn't really necessarily true after we do Cloud BT integration.  i think you axe this assert and comment.

sethtroisi: BT doesn't use this code path, it calls `get_tpu_bt_input_tensors` which makes a supcall to bigtable_input to do most of it's processing.

PR #537, opened 2018-10-30
Add interleave argument documentation

sethtroisi:
This was missed in https://github.com/tensorflow/minigo/pull/536


sethtroisi:
I named it the same so that hopefully Andrew won't see it #hopechess

sethtroisi:
you play a move and hope your opponent doesn't see the refuting move

PR #539, opened 2018-10-31
Allow export without bootstrap

sethtroisi:
Currently none of the endpoints (freeze_model.py, bootstrap.py, train.py allow you to export a checkpoint).

Bootstrap was the closest (see how small the edit distance is).

Initially I cloned the bootstrap.py file and removed a single line but this seemed silly.


brilee:
i'm super confused how a command named "bootstrap" could not bootstrap.

PR #543, opened 2018-11-02
Add platform abstraction

tommadams:
Required to get `lite_dual_net.cc` running on OS X, since `get_nprocs` doesn't exist.


chsigg:
Tom, I'm using your PR to test my custom image that I just tagged 'latest'.
/retest

PR #552, opened 2018-11-08
During self-play, write bleakest move to Bigtable

gitosaurus:
The Q value will be written to a column named
`metadata:bleakest_q` for the move row at which the
evaluation occurred, for games in which resigning
has been disabled (bleakest-move calculation is
not relevant otherwise, since quitters don't know
how bad it can get).  This makes it straightforward
to collect the games, moves, and Q values for
bleakest-move evaluations over a range of games,
e.g.:

  `cbt read games columns=metadata:bleakest_q`

The only rows selected will be those with a value
in the column, and the row name will contain the
game and move number of the evaluation.

Also reduced some coupling and refactored the
calculation of bleakest move to a function callable
from more than `main.cc`.


cc/tf_bt_utils.cc
```diff
@@ -112,6 +112,18 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
   auto row_prefix = absl::StrFormat(kGameRowFormat, game_counter);
   WriteTfExamples(table, row_prefix, examples);
+  if (!player.options().resign_enabled) {
+    int bleakest_move = 0;
+    float bleakest_q = 0.0;
+    FindBleakestMove(player, &bleakest_move, &bleakest_q);
+    auto bleak_row_name =
+        absl::StrFormat(kPrefixAndMoveFormat, row_prefix, bleakest_move);
+    google::cloud::bigtable::SingleRowMutation row_mutation(bleak_row_name);
+    auto q_data = absl::StrFormat("%f", bleakest_q);
```
tommadams: I think you can just use `absl::StrCat` here instead of `absl::StrFormat`

cc/mcts_player.cc
```diff
@@ -407,4 +407,25 @@ void MctsPlayer::ProcessLeaves(absl::Span<MctsNode*> leaves,
   }
 }
 
+void FindBleakestMove(const MctsPlayer& player, int* move, float* q) {
```
tommadams: This should probably return a bool indicating whether `move` and `q` were updated, depending on whether `player.history()` is empty.

I see it's currently fine because `FindBleakestMove` isn't called if the history is empty, but we might end up reusing this function in the future.

gitosaurus: Makes sense. Will do.

cc/tf_bt_utils.cc
gitosaurus: Yes, that's cleaner, and moreover, I see that `StrCat` will use 6 digits of *precision* (which I prefer in this case) rather than 6 digits to the right of the decimal, which is what the `"%f"` specifier gets me. Thanks!

PR #553, opened 2018-11-09
Move game over from player to game tree

tommadams:
This is required for the upcoming changes that will allow analyzing multiple variations in a full game tree. Since it's possible that a full game tree might contain multiple game over states, we need to store the game over state in the tree itself.


cc/mcts_player.cc
```diff
@@ -275,12 +271,10 @@ bool MctsPlayer::PlayMove(Coord c) {
   }
 
   // Handle consecutive passing.
```
amj: Nit: update comment: "Handle consecutive passing or termination by move limit" or similar?

PR #555, opened 2018-11-10
Script to create distillation training data

sethtroisi:
Read a stream of training examples (e.g. a golden_chunk)
and write a stream of training examples with labels replaced by trainer model's output.

At the end produce a graphic about policy/value error


oneoffs/distillation.py
```diff
@@ -0,0 +1,132 @@
...
+                    print()
+                    print("Breaking OutOfRangeError")
+                    break
+
+    print("Counts", len(policy_err), len(value_err))
+    test()
+
+    plt.subplot(121)
+    n, bins, patches = plt.hist(policy_err, 40)
+    plt.title('Policy Error histogram')
```
brilee: I'd also be interested in seeing histograms for the following:
- entropy of the output distribution (`output_probs = tf.exp(logits); entropy = tf.reduce_sum(tf.log(output_probs) * output_probs)`

brilee: oh actually I think it can also be computed as `entropy = tf.nn.softmax_cross_entropy_with_logits_v2(logits=logits, labels=logits)`

oneoffs/distillation.py
```diff
@@ -0,0 +1,133 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Used to create distillation training examples.
+
+Distillation is training on the value from a teacher network (z=0.7)
+instead of the supervised labels (z=1). In theory this can help because
+the values are expressable by some network.
```
amj: nit: this is distillation, but it doesn't actually explain this file.  propose adding another sentence like 
"this script replaces the game results in a file of TFExamples (-1 or 1) with a network's evaluation of that position" (and probably add a network to your example usage string)

oneoffs/distillation.py
```diff
@@ -0,0 +1,133 @@
...
+def test():
+    records = list(tqdm(tf.python_io.tf_record_iterator(FLAGS.out_path, OPTS)))
+    print()
+    print("Values:")
+    for i, r in enumerate(records[:10]):
+        e = tf.train.Example.FromString(r)
+        print("\t", i, e.features.feature['outcome'].float_list.value)
+
+
+def main(unusedargv):
```
amj: is this better than _ ?

PR #559, opened 2018-11-12
Bump TF version to 1.12 for cc/...

chsigg:
This upgrades the bazel dependency of TF under cc/... to version 1.12.
The latest bazel 0.19.1 doesn't seem to build TF 1.11 with CUDA anymore.

I don't need this change at all, but I already had it locally so I thought I would make a PR if you are interested. It's trivial to redo later if you want to hold off. Feel free to close immediately ;-)


amj:
awesome :)  I don't think we want to close immediately, but i do think this is a good place to discuss what we *should* be doing going forward when we want to bump versions.

At a minimum, i'd like to ensure that it builds with CUDA, and i'd love it if we could bump the versions in the README & other docs with the latest known-working configuration.

At a maximum, i'd like some kind of way to test that it would actually work on a ctpu.  Naturally, our automated tests don't cover ctpus (and in fact now that i think of it, i'm not really sure what the test story for the ctpu product is supposed to be), so it's a good idea to at least ensure that runs with the latest changes.

I'm sure there's other stuff in between those two endpoints.
@tommadams @sethtroisi thoughts?

amj:
closing, obsolete.

PR #565, opened 2018-11-14
Forgot to force cell value to integer

gitosaurus:
Silly but important: `set_cell` errors out if its argument isn't either a string or an int, and in some cases it gets passed a float.


sethtroisi:
LGTM

PR #566, opened 2018-11-15
Create CBT instance only if needed

gitosaurus:
It's much more common to create new tables within an existing
instance, so the CBT instance-creation code should be separate
and only invoked if the instance does not exist.

Introduce new script `create_table.sh` which will create a
CBT table based on the current `CBT_` variable settings.


cluster/utils.sh
```diff
@@ -94,12 +94,28 @@ function create_cbt_table() {
...
+#   CBT_INSTANCE: The Cloud Bigtable instance within PROJECT (create if absent)
+#   CBT_TABLE:  The name of the Cloud Bigtable table to create in CBT_INSTANCE
+function create_cbt_table() {
+  check_cbt_exists
+  if ! ( cbt -project ${PROJECT} listinstances | grep -wq "^${CBT_INSTANCE}" ); then
+    create_cbt_instance
+  fi
+  if ! (cbt -project ${PROJECT} -instance ${CBT_INSTANCE} createtable ${CBT_TABLE} && \
+           cbt -project ${PROJECT} -instance ${CBT_INSTANCE} createfamily ${CBT_TABLE} tfexample && \
+           cbt -project ${PROJECT} -instance ${CBT_INSTANCE} setgcpolicy ${CBT_TABLE} tfexample maxversions=1 &&
```
sethtroisi: it's kind a weird that this doesn't have a \ and all the others do.

gitosaurus: It sure is! Yet it works, as I just reassured myself. Apparently bash will continue the line if it can tell that it hasn't terminated, as must be the case here because of an open parenthesis:

https://stackoverflow.com/questions/3871332/how-to-tell-bash-that-the-line-continues-on-the-next-line/35931689

More like Python than I had realized. It does make me want to *remove* the other backslashes, now! Seriously, I do agree that it should be consistent.

PR #569, opened 2018-11-16
Revert #534 and build cc-base

amj:
i've only tested this with building cc-base only, but this can be a good place to put further discussion about where things need to go :)


amj:
Ok, all images build, i haven't verified that they all run correctly yet, but they do all build :)  If y'all wanna have a look at this, i'll merge it in and then verify that the builds are ok -- and probably will need to fix the test image too.

amj:
ok, our tests will be broken until i get a PR into kubernetes-infra to fix the conf there.

PR #571, opened 2018-11-16
Write eval results to bigtable for CC Eval games.

sethtroisi:


cc/tf_bt_utils.cc
```diff
@@ -127,6 +128,46 @@ void WriteGameExamples(const std::string& gcp_project_name,
             << examples.size() << std::endl;
 }
 
+void WriteEvalRecord(const std::string& gcp_project_name,
+                     const std::string& instance_name,
+                     const std::string& table_name,
+                     const MctsPlayer& player,
+                     const std::string& black_player_name,
+                     const std::string& white_player_name) {
+  // TODO assert("eval" in table_name);
```
sethtroisi: I want to make sure we don't accidentally write to the wrong table. not sure best way to prevent that.

cc/tf_bt_utils.cc
```diff
@@ -127,6 +128,46 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
+  auto row_name = absl::StrFormat(kEvalGameRowFormat, game_counter);
+  google::cloud::bigtable::SingleRowMutation row_mutation(
+      row_name,
+      SetCell("metadata", "black", black_player_name),
+      SetCell("metadata", "white", white_player_name),
+      SetCell("metadata", "black_won", player->result() > 0),
+      SetCell("metadata", "white_won", player->result() < 0)
+      SetCell("metadata", "result", player->result())
+      SetCell("metadata", "length", player->result() < 0)
+  );
```
sethtroisi: I'm not sure how to test this in production. advice would be appreciated.

cluster/utils.sh
```diff
@@ -103,6 +103,7 @@ function create_cbt_table() {
     echo "Could not create table ${CBT_TABLE} on instance ${CBT_INSTANCE} in project ${PROJECT}"
     return 1
   fi
+  # TODO(sethtroisi): Figure out stuff here.
```
sethtroisi: Waiting for #566 then I'll add to the new file.

cc/tf_bt_utils.cc
gitosaurus: When I test, I build for a board size of 9 (for speed) and create a new table in the existing instance: `CBT_TABLE=games-9x9` (or any table that isn't the main one).

You may want to coerce/format your values to string as you pass them to `SetCell`. Even for integers, you may want them stored as strings so that they are more easily inspected. Use cases for storing an integer *as* an integer include `metadata:game_counter`, where you need to do an atomic set-and-get. These end up being big-endian integers, and are hard to read when viewing with `cbt`.

cc/tf_bt_utils.cc
gitosaurus: If I understand the rest of the code, it looks like you're only wanting to add one new evaluation row per game, with the prefix `e_`. If so, it can simply be another keyspace within the same table, like the `ct_` keyspace for move counts.

cluster/utils.sh
gitosaurus: Per our conversation earlier in the week, I believe that we want two tables with the very same schema (where both would have the new `e_` keyspace), but which are then combined by the trainer.

For the game-playing nodes, then, nothing needs to change in this file yet, because each node would be given a different setting for `CBT_TABLE` depending on its role, correct?

But since the trainer needs to identify both tables, maybe we will want some environment variable renaming.

amj: The trainer actually doesn't need to see the evaluation games at all -- these are games played with randomness off (no softpick, no dirichlet noise), which function as the way to evaluate which is a better model.  The resulting games are not actually used by the trainer; they're used by the overall pipeline code to control promotion of models, determine pairings, etc.

sethtroisi:
```
E1116 20:43:40.460187653  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.460334165  207658 ssl_transport_security.cc:785] Could not load root certificates.
E1116 20:43:40.493959211  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.493991119  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.494980694  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.494993323  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.495003015  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.523807843  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.524001804  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.524973260  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.525095074  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.525168671  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.568455620  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.568490597  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.569488509  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.569502892  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.569509124  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.601056704  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.601085800  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.602075593  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.602088948  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.602095668  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.632020766  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.632049615  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.633083183  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.633098219  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.633105894  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.663148048  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.663364450  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.664455056  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.664468420  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.664476312  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.694616134  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.694653909  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.695637052  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.695653798  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.695662782  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
E1116 20:43:40.725445181  207658 ssl_transport_security.cc:618] Could not add root certificate to ssl context.
E1116 20:43:40.725487546  207658 ssl_transport_security.cc:1690] Cannot load server root certificates.
E1116 20:43:40.729632500  207658 security_connector.cc:1062] Handshaker factory creation failed with TSI_INTERNAL_ERROR.
E1116 20:43:40.729662686  207658 secure_channel_create.cc:121] Failed to create secure subchannel for secure name 'bigtable.googleapis.com'
E1116 20:43:40.729672406  207658 secure_channel_create.cc:154] Failed to create subchannel arguments during subchannel creation.
```

cc/tf_bt_utils.cc
sethtroisi: Thanks for that knowledge :)

sethtroisi:
I had to change `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH` (which is set in cluster/common.sh) to a different file `google3/security/cacerts/for_connecting_to_google/roots.pem` to run on my workstation but now everything works

image

cc/tf_bt_utils.cc
```diff
@@ -127,6 +128,44 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
+                     const std::string& black_player_name,
+                     const std::string& white_player_name) {
+  Table table(CreateDefaultDataClient(gcp_project_name, instance_name,
+                                      google::cloud::bigtable::ClientOptions()),
+              table_name);
+
+  // Retrieve the game counter from the Bigtable and increment it atomically.
+  using namespace google::cloud::bigtable;
+  auto rule =
+      ReadModifyWriteRule::IncrementAmount("metadata", "game_counter", 1);
```
sethtroisi: This increments the main game counter. I'm should probably change this eval_counter

cc/main.cc
```diff
@@ -627,6 +627,15 @@ class Evaluator {
     // The player and other_player reference this pointer.
     std::unique_ptr<DualNet> dual_net;
 
+    std::vector<std::string> bigtable_spec =
+        absl::StrSplit(FLAGS_output_bigtable, ',');
+    bool use_bigtable = bigtable_spec.size() == 3;
```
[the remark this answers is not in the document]
sethtroisi: This is copy-pasted from 354 so I assume it's correct

cc/tf_bt_utils.cc
```diff
@@ -127,6 +128,44 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
+  auto row = table.ReadModifyWriteRow("table_state", rule);
+  uint64_t game_counter = 0;
+  // TODO assert size == 1?
```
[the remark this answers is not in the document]
sethtroisi: I was talking about the size of row.cells(),
e.g. can I add `assert(row.cells().size() == 1);`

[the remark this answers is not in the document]
gitosaurus: The `table_state` row may contain more than one cell, especially in the future. However, the documentation says that it returns "A Row containing the new contents of all cells modified by the request," so if you have only updated one cell in the row, it may be safe to insist on its size.

cc/tf_bt_utils.cc
```diff
@@ -118,23 +119,55 @@ void WriteGameExamples(const std::string& gcp_project_name,
...
   auto row = table.ReadModifyWriteRow("table_state", rule);
   for (auto const& cell : row.cells()) {
     if (cell.family_name() == "metadata" &&
-        cell.column_qualifier() == "game_counter") {
+        cell.column_qualifier() == counter_name) {
       return cell.value_as<bigendian64_t>().get();
     }
   }
-  MG_FATAL() << "Failed to increment table_state=metadata:game_counter";
+  MG_FATAL() << "Failed to increment table_state=metadata:" << counter_name;
```
sethtroisi: @gitosaurus Couple of questions here:
1. Why do we iterate row.cells()? Should we ever expect more than one value?
1. Is there a way to set a uint value from the CLI?
    * something like `cbt -project seth-minigo -instance minigo-instance set games table_state metadata:game_counter='\x00\x00\x00\x00\x00\x00\x00\f'`
1. The only downside to using this method is that I have to create the Table (line 158) twice for each table write, can you confirm that table object is "cheap" to create.

gitosaurus: 1. The reason I iterate is because it does return a collection, and I wanted to be robust against any future changes to our code or the client code where we might get back more than one cell from the `ReadModifyWriteRow`. At present, the [documentation](https://googlecloudplatform.github.io/google-cloud-cpp/0.2.0/bigtable/classgoogle_1_1cloud_1_1bigtable_1_1v0_1_1Table.html#af23d1796d60442eac811eac99808ad9f) states simply that "the row" is returned.
2. I don't know how to set a uint64 value from the `cbt` CLI (which is one reason to use them sparingly). From C++ and Python, it is straightforward.
3. Construction of the table object (which is just a handle to the client) is quite cheap. Profiling on Python, table creation is 2 microseconds, compared to 6 milliseconds for a row commit.

cc/tf_bt_utils.cc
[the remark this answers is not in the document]
sethtroisi: Okay, I'll investigate in the future and leave this as a better formatted todo today.

cc/tf_bt_utils.cc
sethtroisi: 1. Thank you fro the explanation.
2. Ack, I ended up using python which was easy.
3. Thank you for measuring / knowing

PR #572, opened 2018-11-16
Refactor into GameQueue class in prep for 2 queues

gitosaurus:
Since we expect the final input to come from a mix of two
game queues, it's necessary to refactor the `bigtable_input`
module to contain a `GameQueue` class so that there can
be more than one instance.

This change adds no functionality and is a refactoring
change only.


bigtable_input.py
```diff
@@ -97,76 +73,259 @@ def make_single_array(ds, batch_size=8*1024):
...
+                start_time = time.time()
+                ds = self.tf_table.keys_by_range_dataset(start_row, end_row)
+                h = _histogram_move_keys_by_game(sess, ds)
+                self._write_move_counts(sess, h)
+                end_time = time.time()
+                elapsed = end_time - start_time
+                print('  games/sec:', len(h)/elapsed)
+
+
+# TODO(president.jackson): document these
```
amj: document what parts?  for who?  PROJECT is the cloud project id...

gitosaurus: I believe this is a comment that Seth added when the first draft of this went in, and the refactor has made it visible again. I can write a quick blurb for each; where do we like this documentation?

amj: ok so my question when i read this request for docs is is "who is looking for this and when do they see it -- i.e., does someone running this without them set get an error?  what is the person reading it trying to accomplish?"  etc.

it looks like this will give a KeyError if any of these aren't defined, so probably a little blurb here about the CBT_ variables is fine.

gitosaurus: Done!

PR #575, opened 2018-11-20
Evaluate win rate for loaded SGFs in the background

tommadams:
This PR adds a map of auxiliary information for positions played during a game (and its variations) to the GtpPlayer. This gives us somewhere to put extra data for moves played without making the base MctsNodes any bigger, since there are normally many mode MctsNodes in a search tree than moves played.

The GtpPlayer maintains a queue of nodes to perform win-rate evaluation on. The queue holds all nodes in the current variation that have had fewer than num_eval_reads win-rate evaluation tree searches performed on them. The number of tree searches performed for win-rate evaluation is tracked separately that the tree search visit count to ensure fairness.


minigui/winrate_graph.ts
```diff
@@ -45,10 +45,11 @@ class WinrateGraph extends View {
   protected mainLine: number[] = [];
   protected variation: number[] = [];
 
-  protected moveNum = 0;
-
   protected xScale = MIN_POINTS;
```
amj: is this xScale unneeded, now?

minigui/winrate_graph.ts
```diff
@@ -191,13 +202,16 @@ class WinrateGraph extends View {
...
     ctx.textAlign = 'left';
     ctx.fillStyle = '#ffe';
-    let y = 0;
-    let values = this.variation.length > 0 ? this.variation : this.mainLine;
+    let q = 0;
+    let values =
+        this.activePosition.isMainLine ? this.mainLine : this.variation;
     if (values.length > 0) {
-      y = values[this.moveNum];
+      console.log(values);
```
amj: obsolete?

cc/gtp_player.h
```diff
@@ -47,18 +48,52 @@ class GtpPlayer : public MctsPlayer {
 
     // If true, we will always pass if the opponent passes.
     bool courtesy_pass = false;
+
+    // Number of times to perform tree search for each position in an SGF in
+    // order to evalute a win rate estimation.
+    int num_eval_reads = 4;
```
amj: can there be a GTP command (or a flag) for this?

cc/gtp_player.cc
```diff
@@ -115,13 +116,55 @@ void GtpPlayer::Run() {
...
+}
+
 Coord GtpPlayer::SuggestMove() {
   if (courtesy_pass_ && root()->move == Coord::kPass) {
     return Coord::kPass;
   }
   return MctsPlayer::SuggestMove();
 }
 
+bool GtpPlayer::PlayMove(Coord c) {
```
amj: can you comment a bit about how this is different than genmove / make explicit the linkages here w.r.t. the front end?

tommadams: I've added a TODO to gtp_player.h to document all the Handle* methods of GtpPlayer. I'll do a pass over the comments & documentation once the important features are mostly finished

cc/gtp_player.h
tommadams: added a "ponder winrate N" command

minigui/winrate_graph.ts
tommadams: removed.

minigui/winrate_graph.ts
tommadams: added a comment describing why xScale is needed

PR #578, opened 2018-11-22
Read games from two CBT tables

gitosaurus:
If regular games are recorded in a table called `games`, there should be a table in the same instance called
`games-nr`, containing games which are "no-resign", played to their end.  The final selection includes
90% of the former and 10% of the latter.


bigtable_input.py
```diff
@@ -349,15 +376,24 @@ def get_unparsed_moves_from_last_n_games(n,
...
-    _games.wait_for_fresh_games()
-    latest_game = int(_games.latest_game_number())
-    utils.dbg('Latest game: %s' % latest_game)
-    if latest_game == 0:
-        raise ValueError('Cannot find a latest game in the table')
-
-    start = int(max(0, latest_game - n))
-    ds = _games.moves_from_games(start, latest_game, moves, shuffle,
-                                 column_family, column)
+    c_r, c_nr = 9, 1
```
amj: i assume these are 'count', 'take', and 'percent', but a one line comment would aid legibility

gitosaurus: I agree. Will do.

PR #580, opened 2018-11-27
First step to getting Minigo running on Windows

tommadams:
Gets Minigo building and its tests passing on Windows. This PR doesn't make any attempt to get any inference engine working (other than FakeDualNet), so the windows code currently only builds with `--define=tf=0`.

Most of the work here was file system yak shaving.


cc/file/path.cc
```diff
@@ -14,10 +14,20 @@
 
 #include "cc/file/path.h"
 
+#include "absl/strings/match.h"
 #include "absl/strings/str_cat.h"
 
 namespace minigo {
 namespace file {
+
+#ifdef _WIN32
```
amj: no absl lib for this i guess

cc/coord.cc
```diff
@@ -24,7 +24,7 @@ namespace minigo {
 
 constexpr uint16_t Coord::kPass;
 constexpr uint16_t Coord::kInvalid;
-constexpr char Coord::kKgsColumns[];
+const char Coord::kKgsColumns[20] = "ABCDEFGHJKLMNOPQRST";
```
amj: i'm scared to ask why you needed to do this.

cc/file/path.cc
tommadams: Not yet. It's wrapped up with the rest of the absl::File API, which is on the road map to release but not for quite some time.

PR #582, opened 2018-11-28
Updated using google::cloud::bigtable in tf_bt_utils

sethtroisi:
Makes things prettier IMHO


cc/tf_utils.h
```diff
@@ -16,6 +16,7 @@
 #define CC_TF_UTILS_H_
 
 #include <string>
+#include <vector>
```
amj: was this needed or is this git artifact?

sethtroisi: It should be here and I included it because I was fixing this files.

PR #584, opened 2018-11-29
Add validate_misc.py script

sethtroisi:
I want a place to add these, alternatively we can have a bunch of gists linked in some issue.


oneoffs/validate_misc.py
```diff
@@ -0,0 +1,41 @@
...
+        pi = np.frombuffer(example.features.feature['pi'].bytes_list.value[0], np.float32)
+        value = example.features.feature['outcome'].float_list.value[0]
+        assert abs(pi.sum() - 1) < 1e-4, pi.sum()
+        assert value in (-1, 1), value
+
+    opts = tf.python_io.TFRecordOptions(tf.python_io.TFRecordCompressionType.ZLIB)
+    for record in tqdm(tf.python_io.tf_record_iterator(example_file, opts)):
+        test_example(record)
+
+#validate_examples("../data/300.tfrecord.zz")
```
amj: you should make this an ifmain block

PR #585, opened 2018-11-29
Adds ReloadingDualNet

tommadams:
`ReloadingDualNetFactory` wraps a "real" `DualNetFactory`, monitoring a directory for new models that match a file pattern specified with a "%d" in the basename. The `ReloadingDualNetFactory` uses the wrapped factory to create `DualNet` instances wrapped by `ReloadingDualNet`. When a new model is found, the corresponding `ReloadingDualNet` instances are all automatically updated with new instances of the wrapped `DualNet`.

...I've made this sound more confusing than it actually is, I think...

This architecture allows us to automatically reload models of any `DualNet` implementation.

I've tested this code manually using `selfplay` and `eval` modes of `TfDualNet` and `LiteDualNet` with and without the `BatchingDualNet` wrapper.

I haven't yet tested Cloud TPU and it probably doesn't work because I haven't yet separated out the platform initialization and shutdown code. I'll tackle that in a followup PR.


cc/dual_net/reloading_dual_net.cc
```diff
@@ -0,0 +1,194 @@
...
+
+// ReloadingDualNetUpdater methods follow.
+
+ReloadingDualNetUpdater::ReloadingDualNetUpdater(const std::string& pattern,
+                                                 DualNetFactory* factory_impl)
+    : factory_impl_(factory_impl) {
+  auto pair = file::SplitPath(pattern);
+
+  directory_ = std::string(pair.first);
+  MG_CHECK(directory_.find('%') == std::string::npos &&
```
amj: worth it to extract the validator to an independent method?

cc/dual_net/reloading_dual_net.h
```diff
@@ -0,0 +1,128 @@
...
+                          DualNetFactory* factory_impl);
+
+  // Scans directory_ for a new model that matches basename_pattern_.
+  // If a new model is found, all registered ReloadingDualNets are updated
+  // using DualNet instances created from the updater's factory_.
+  // Returns true if a new model was found.
+  bool Poll();
+
+  // Unregisters a model with the updater.
+  // There isn't a matching RegisterModel method because updater registers
```
amj: ha!  i was wondering this.  thanks4doc

cc/dual_net/reloading_dual_net.cc
amj: might allow for testability for any globby/stringy code

tommadams: good point. added some tests

PR #588, opened 2018-11-30
Allow [tt] for pass

### LINKED ISSUE #587: sgf parser explosion
i don't have a callstack handy as i was using Seth's box.
But i was using [this sgf](https://pandanet-igs.com/system/sgfs/8031/original/AQEYAWDNSF.sgf?1543413330)

from [this page](https://pandanet-igs.com/communities/agacityleague/rounds/518)

sethtroisi:
Fixes #587


cc/coord.cc
```diff
@@ -30,7 +30,7 @@ namespace {
 
 Coord TryParseKgs(absl::string_view str) {
   auto upper = absl::AsciiStrToUpper(str);
-  if (upper == "PASS") {
+  if (upper == "PASS" || upper == "TT") {
```
tommadams: I don't think this is correct: `TT` isn't a valid KGS coordinate. The GTP and KGS docs I've found all say that pass should be represented as the string `pass`. I think you can revert this particular line.

cc/coord.cc
```diff
@@ -52,7 +52,7 @@ Coord TryParseKgs(absl::string_view str) {
 }
 
 Coord TryParseSgf(absl::string_view str) {
-  if (str.empty()) {
+  if (str.empty() || str == "tt") {
```
amj: Cast to lowercase first?

tommadams: Uppercase letters are reserved for representing points 27-52 on larger boards, see: https://www.red-bean.com/sgf/go.html#types

tommadams: To be clear, assuming lowercase is the correct thing to do here.

cc/coord.cc
amj: i think "KGS" is the wrong name for these coordinates.  these are GTP coordinates and we're using this to translate our GTP commands.

So i think this should be "TryParseGtp"

cc/coord.cc
amj: i guess the right thing to do is to compare `(str == "tt" && boardsize ==19)`

PR #592, opened 2018-12-03
put the barrier back, it's still needed on some platforms it seems

tommadams:


cc/main.cc
```diff
@@ -661,6 +714,9 @@ class Evaluator {
       }
       std::swap(model_path, other_model_path);
       std::swap(player, other_player);
+
+      // Notify the barrier that this thread is no longer participating.
+      barrier_->DecrementCount();
```
chsigg: This goes after the while block, right?

PR #593, opened 2018-12-04
CHECK that tpu_name looks somewhat like a TPU name

tommadams:
TensorFlow gives some very confusing error messages if `tpu_name` is wrong or missing.


tommadams:
Actually tpu_name can't be a name of the VM because there isn't a C++ equivalent of the [TPUClusterResolver](https://github.com/tensorflow/tensorflow/blob/a26f3b05986eaec30fa3d9547decbaa9607291dd/tensorflow/python/distribute/cluster_resolver/tpu_cluster_resolver.py#L45).

sethtroisi:
I see this is C++ :(

PR #594, opened 2018-12-04
make gRPC use proto v3.6.0 instead of v3.5.0

tommadams:


amj:
(LGTM and merge and maybe we keep track of this in an issue somewhere)

PR #596, opened 2018-12-05
Added oneoff_utils.final_position_sgf

sethtroisi:
replay_sgf doesn't return the final position. you have to play_move(p.next_move)

I updated oneoff_utils to help avoid this problem in the future.


brilee:
LGMT.

PR #598, opened 2018-12-06
Calibration updater

amj:
Script for updating the resign-threshold value in a flagfile.


rl_loop/update_resign_threshold.py
```diff
@@ -0,0 +1,83 @@
...
+import sys
+import re
+import os
+import time
+from absl import app
+from absl import flags
+import numpy as np
+import tensorflow as tf
+
+sys.path.insert(0, '.')
```
gitosaurus: I assume this is needed because `bigtable_input` isn't located sibling to this module, but that the script gets run from the top-level minigo directory. Is this a pattern used elsewhere here?

rl_loop/update_resign_threshold.py
```diff
@@ -0,0 +1,83 @@
...
+        lines = f.read()
+    if new_threshold > 0:
+        new_threshold *= -1
+    if not RESIGN_FLAG_REGEX.search(lines):
+        print("Resign threshold flag not found in flagfile {}!  Aborting.".format(flags_path))
+        sys.exit(1)
+    old_threshold = RESIGN_FLAG_REGEX.search(lines).groups(1)
+    lines = re.sub(RESIGN_FLAG_REGEX, "--resign_threshold={:.3f}".format(new_threshold), lines)
+
+    print("Updated percentile from {} to {:.3f}".format(old_threshold, new_threshold))
```
gitosaurus: Same format string for same quantities?

rl_loop/update_resign_threshold.py
```diff
@@ -0,0 +1,83 @@
...
+
+FLAGS = flags.FLAGS
+RESIGN_FLAG_REGEX = re.compile(r'--resign_threshold=([-\d.]+)')
+
+def get_95_percentile_bleak(n_back=1000):
+    """Gets the 95th percentile of bleakest_eval from bigtable"""
+    end_game = int(bigtable_input._games_nr.latest_game_number())
+    start_game = end_game - n_back if end_game >= n_back else 0
+    moves = bigtable_input._games_nr.bleakest_moves(start_game, end_game)
+    evals = np.array([m[2] for m in moves])
```
gitosaurus: Now that I think of it, I should take the trouble to return a named tuple from `bleakest_moves`. I'll do it in another PR.

cc/mcts_player.cc
```diff
@@ -421,6 +421,7 @@ void MctsPlayer::ProcessLeaves(absl::Span<TreePath> paths,
   }
 }
 
+// &q is the bleakest move from the perspect of the winner.
```
sethtroisi: +(e.g. probably negative)

rl_loop/update_resign_threshold.py
amj: that answer isn't actually right?

rl_loop/update_resign_threshold.py
amj: SGTM

rl_loop/update_resign_threshold.py
```diff
@@ -0,0 +1,83 @@
...
+
+def update_flagfile(flags_path, new_threshold):
+    """Updates the flagfile at `flags_path`, changing the value for
+    `resign_threshold` to `new_threshold` 
+    """
+    if abs(new_threshold) > 1:
+        raise ValueError("Invalid new percentile for resign threshold")
+    with tf.gfile.GFile(flags_path) as f:
+        lines = f.read()
+    if new_threshold > 0:
```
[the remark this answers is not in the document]
amj: it's possible (though unlikely) that for a biased model, enough of the thresholds are positive that the set of bleakest evals are all > 0 , which is why i didn't make this a failure case.

cc/mcts_player.cc
amj: agreed, i'll fix the comment.

cc/mcts_player.cc
```diff
@@ -442,7 +443,7 @@ bool FindBleakestMove(const MctsPlayer& player, int* move, float* q) {
     }
   }
   *move = int(bleakest_move);
-  *q = history[bleakest_move].node->Q();
+  *q = history[bleakest_move].node->Q() * result;
```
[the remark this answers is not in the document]
amj: yes it is.  i thought i liked the explicit-ness but the other way is better.  done!

rl_loop/update_resign_threshold.py
```diff
@@ -0,0 +1,83 @@
...
+FLAGS = flags.FLAGS
+RESIGN_FLAG_REGEX = re.compile(r'--resign_threshold=([-\d.]+)')
+
+def get_95_percentile_bleak(n_back=1000):
+    """Gets the 95th percentile of bleakest_eval from bigtable"""
+    end_game = int(bigtable_input._games_nr.latest_game_number())
+    start_game = end_game - n_back if end_game >= n_back else 0
+    moves = bigtable_input._games_nr.bleakest_moves(start_game, end_game)
+    evals = np.array([m[2] for m in moves])
+    return np.percentile(evals, 5)
```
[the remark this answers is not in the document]
amj: see below.

rl_loop/update_resign_threshold.py
amj: Right.

PR #604, opened 2018-12-08
s/EVAL_VERSION_TAG/VERSION_TAG

sethtroisi:
EVAL_VERSION_TAG isn't set anywhere in the code so I'm changing this to VERSION_TAG


amj:
EVAL_VERSION_TAG shouldn't be set anywhere -- the eval containers can have totally different image names, whereas version_tag is usually set for a single run.

sethtroisi:
it should still probably get a value from common.sh or be mentioned in some documentation.

PR #605, opened 2018-12-10
Avoid buffer edge conditions in sharding

gitosaurus:
Do `ds.take()` up front, before sharding, and use a buffer,
rather than a block, for interleaving.


bigtable_input.py
```diff
@@ -294,23 +294,24 @@ def moves_from_games(self, start_game, end_game, moves, shuffle,
...
             ds = rds.apply(
                 tf.contrib.data.parallel_interleave(
                     lambda x: ds.shard(shards, x),
-                    cycle_length=shards, block_length=1024,
+                    cycle_length=shards,
+                    buffer_output_elements=buffer_size,
                     sloppy=True))
-            ds = ds.shuffle(shards * 1024 * 2)
-        ds = ds.take(moves)
+            ds = ds.shuffle(shards * buffer_size * 2)
```
amj: this is a smaller shuffle, right? (shards * 128 * 2 vs shards * 1024 * 2).  Can we keep the larger one or is this important to the fix?

gitosaurus: Yes, I think this final shuffer buffer can remain at its original size, or possibly a little larger, since it's the final shuffle. I'll test the performance effect of some larger sizes.

PR #607, opened 2018-12-11
Rewrite the batching code

tommadams:
This PR adds `StartGame` and `EndGame` methods to `BatchingDualNetFactory` so that it can keep track of which models are playing each other and perform better batching. The idea is that if model A is playing model B, the batching code knows that model B isn't going to make any inferences while model A is performing tree search.

This rewrite means the batching code will do the correct thing without callers having to worry about synchronization, meaning with can eliminate the `Barrier` and `WrappedDualNet` code from the `Evaluator`. As an added bonus, we also don't need to know the number of parallel games ahead of time, meaning we can scale the parallel game dynamically (something that would be useful for playing resign disabled games as quickly as possible).

In my single GPU tests, the batching performed is close to optimal for selfplay and eval. I've also done some light testing with 3 or more models playing games amongst each other and the batching in these cases also seems sensible. 

Here's the output of running 32 parallel eval games with 32 reads per move at HEAD and with this PR:

```
# Current code at HEAD
Evaluated 32 games, total time 10.098505238s
Wins          Total         Black         White
eval_a    13  40.62%     7  21.88%     6  18.75%
eval_b    19  59.38%    10  31.25%     9  28.12%
                        17  53.12%    15  46.88%
Ran 372 batches with an average size of 68.0914.
Ran 371 batches with an average size of 68.5067.
```

```
# This PR
Evaluated 32 games, total time 8.313876682s
Wins          Total         Black         White
eval_a    19  59.38%    10  31.25%     9  28.12%
eval_b    13  40.62%     7  21.88%     6  18.75%
                        17  53.12%    15  46.88%
Ran 292 batches with an average size of 86.7466.
Ran 292 batches with an average size of 87.0411.
```

For the time being, I've shoved the `StartGame` and `EndGame` methods into the base `DualNetFactory` to keep the changes to a minimum. I plan on refactoring the code in a followup CL to remove these methods from the base class.

I also had to rewrite the puzzle mode because when I tried to test it locally, it immediately OOMed because it was submitting too large batches.


tommadams:
@chsigg if you have the chance, could you try this PR out on your multi-GPU machine please?

chsigg:
Happy to do so, thanks for this change. It might take a little bit, we are in the middle of TF 1.13 release. I have to push CUDA 10 through for that and I just got sick. Bad timing.

tommadams:
> Happy to do so, thanks for this change. It might take a little bit, we are in the middle of TF 1.13 release. I have to push CUDA 10 through for that and I just got sick. Bad timing.

Thanks Christian! Absolutely take care of yourself first :)

cc/main.cc
```diff
@@ -757,40 +683,63 @@ void Puzzle() {
...
+      auto trees = GetTrees(ast);
+      MG_CHECK(!trees.empty());
+      auto moves = trees[0]->ExtractMainLine();
+
+      total_moves += moves.size();
+
+      // Create player.
+      auto player = absl::make_unique<MctsPlayer>(
+          model_factory->NewDualNet(FLAGS_model), options);
+      model_factory->StartGame(player->network(), player->network());
```
amj: maybe puzzlers can have a single-network at some point in the future?  I dunno.  Most of the ad-hoc stuff with nets should be done in python anyway.

PR #609, opened 2018-12-12
Change the 'wait for games' to trigger at end of training.

amj:
add window_size to train.py flags.  Change train.py to set freshness marker at the end of training.  Change preprocessing to call get_unparsed_moves_from_last_n_games and refactor bigtable_input.get_fresh_moves to new set_fresh_watermark

### DIFF
--- a/bigtable_input.py
+++ b/bigtable_input.py
@@ -383,27 +383,28 @@ _games_nr = GameQueue(_project_name,
                       _instance_name,
                       _table_name + '-nr')
 
-def get_fresh_moves(n, fresh_fraction=0.025, minimum_fresh=18000):
-    """Get a dataset of serialized TFExamples from the last N games.
 
-    This top level wrapper returns the dataset of shuffled moves, using the
-    'require_fresh_games' function to block until enough new games have been
-    played.  The number of fresh games required is the larger of:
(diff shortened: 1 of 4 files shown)


amj:
For this to pass i think bigtable_input.py is gonna need to be refactored to use flags instead of environment variables to specify the bigtable

bigtable_input.py
```diff
@@ -383,27 +383,28 @@ def update_move_counts(self, start_game, end_game, interval=1000):
...
-      n:  an integer indicating how many past games should be sourced.
+      window_size:  an integer indicating how many past games are considered
+      fresh_fraction: a float in (0,1] indicating the fraction of games to wait for
       minimum_fresh:  an integer indicating the lower bound on the number of new
       games.
     """
     latest_game = int(_games.latest_game_number())
     if n > latest_game: # How to handle the case when the window is not yet 'full'
-        _games.require_fresh_games(minimum_fresh)
+        _games.require_fresh_games(int(minimum_fresh * .9))
```
gitosaurus: What is the reason for the 0.9 multiplier in both branches?

amj: since you're only mixing in 90% of the regular games, it doesn't matter whether you have a full window already or not.

PR #610, opened 2018-12-12
add a thread safe logging class

tommadams:
std::cerr is not thread safe: if multiple threads write to it concurrently, their output will be mixed together.

This PR introduces a `MG_LOG` macro, similar in nature to glog's `LOG` macro (which we can't use because of name collisions inside TensorFlow). It also removes `MG_FATAL`, replacing its uses with `MG_LOG(FATAL)` and moves the definition of `MG_CHECK` from `cc/check.h` to `cc/logging.h` to avoid some circular build dependencies.

`MG_LOG` differs from `LOG` in that `INFO` messages don't have a file:line prefix (`WARNING`, `ERROR` and `FATAL` still do). This is done to avoid breaking things that rely on exact strings being logged.

Since `MG_LOG` automatically inserts a newline, this PR also changes `Position::To{Simple,Pretty}String` to not write a newline after the last board row. It also removes the `Position::ToGroupString` which is some really old debugging code for the `Group` class.


tommadams:
The only difference in behaviour that we'd ever likely notice is if some code tried to write hundreds of MBs amounts of data to a `LOG(INFO)` or something, since that would have to be stored on the heap first.

This implementation is lower overhead than glog but more expensive than logging to `std::cerr` directly. The specific costs incurred are:

- A small amount of memory allocated on the stack: `sizeof(std::stringstream) + 16` bytes.
- Heap allocations when building the temporary string. The max is probably ~1kB in our use case.
- One mutex lock/unlock for each `LOG`.

The mutex locking is potentially less work than logging to `std::cerr` directly because `std::cerr` is thread safe, which requires one lock for each call to `operator<<`; logging to the internal `std::stringstream` doesn't require. This is offset by having to perform heap allocations though of course.

PR #612, opened 2018-12-13
get multiple models running on the same TPU

tommadams:
The primary thing this PR is work around a caching bug in the current TF release that prevents multiple models being run on the same TPU. It does this by making the model graph depend on a constant with a random value.

Now that multiple sessions work correctly this PR moves the TPU init & shutdown ops into a separate graph, hard coded into tpu_dual_net.cc.


tommadams:
Oops, I pushed to tommadams:tpu from my desktop instead of the GCE VM I wrote the actual change on.

Should be fixed now, ptal.

PR #613, opened 2018-12-13
Move gtp & puzzle modes out into separate binaries

tommadams:
Let's start with the simplest two modes.


cc/gtp.cc
```diff
@@ -0,0 +1,93 @@
...
+
+namespace minigo {
+namespace {
+
+void Gtp() {
+  GtpPlayer::Options options;
+  options.name = absl::StrCat("minigo-", file::Basename(FLAGS_model));
+  options.ponder_limit = FLAGS_ponder_limit;
+  options.courtesy_pass = FLAGS_courtesy_pass;
+  options.inject_noise = false;
```
sethtroisi: So nice to not have to set these anymore :)

PR #614, opened 2018-12-15
Adds a Game abstraction

tommadams:
Note: this PR is built on top of #613, you can ignore the `cc/gtp.cc` and `cc/puzzle.cc` code in this PR.

The main goal of this PR is to add a class other than `MctsPlayer` that holds enough data so that we can write TF examples & SGFs at the end of a game without having to pass `MctsPlayer` instances to `WriteSgf`, `WriteGameExamples`, etc.

Depending on the configuration, we can have a single `MctsPlayer`, or two `MctsPlayer`s either playing with the same or different models. Having a single `Game` instance for each game played is a little easier to reason about during post-game handling: we don't have to do slightly weird things like passing the same `MctsPlayer` instance to `WriteSgf` twice, or check that two different `MctsPlayer` instances played the same game.

I wanted to do this before splitting selfplay and eval modes into two separate binaries because before I can do the split, I'll need to create some kind of third library that holds `WriteSgf`, `GetOutputName`, `GetOutputDir`, etc that currently live in `cc:main`. I'd like that library to have clean(ish) dependencies and interfaces.


cc/mcts_player.cc
```diff
@@ -279,62 +279,60 @@ bool MctsPlayer::PlayMove(Coord c) {
...
+
+  // Build a comment for the move.
+  auto comment = root_->Describe();
+  if (!models.empty()) {
+    comment =
+        absl::StrCat("models:", absl::StrJoin(models, ","), "\n", comment);
   }
 
   // Convert child visit counts to a probability distribution, pi.
+  std::array<float, kNumMoves> search_pi;
```
amj: is there an issue with this alloc happening every move?

tommadams: A `std::array` is allocated on the stack. There's an extra copy happening here, which is unfortunate but since it's once per move rather than once per read, I'm not that bothered.

cc/tf_bt_utils.cc
```diff
@@ -135,12 +132,12 @@ void WriteEvalRecord(const std::string& gcp_project_name,
...
   auto row_name = absl::StrFormat(kEvalGameRowFormat, game_counter);
   SingleRowMutation row_mutation(
-      row_name, SetCell("metadata", "black", black_player_name),
-      SetCell("metadata", "white", white_player_name),
-      SetCell("metadata", "black_won", absl::StrCat(player.result() > 0)),
-      SetCell("metadata", "white_won", absl::StrCat(player.result() < 0)),
-      SetCell("metadata", "result", player.result_string()),
-      SetCell("metadata", "length", absl::StrCat(player.history().size())),
+      row_name, SetCell("metadata", "black", game.black_name()),
+      SetCell("metadata", "white", game.black_name()),
```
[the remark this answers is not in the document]
tommadams: Oof, thanks for catching that.

tommadams:
> LGTM. at least one place where it looks like there's a white/black switched around tho, should probably break a test at least.

It probably should, but I don't think any of that code has tests :/

PR #615, opened 2018-12-16
Add methods to help trim and repair GameQueues

gitosaurus:
- Add ability to set the game counter directly
 - Fetch list of games with timestamp
 - Trim games from the end by delta time
 - Trim by relative or absolute time


bigtable_input.py
```diff
@@ -169,24 +173,98 @@ def create(self):
...
+    @latest_game_number.setter
+    def latest_game_number(self, latest):
+        table_state = self.bt_table.row(TABLE_STATE)
+        table_state.set_cell(METADATA, GAME_COUNTER, int(latest))
+        table_state.commit()
+
+    def games_by_time(self, start_game, end_game):
+        """Given a range of games, return the games sorted by timestamp.
+
+        Returns [(timestamp, game), ...]
```
amj: where `game` is the row id?  what's the type of game?

bigtable_input.py
```diff
@@ -169,24 +173,98 @@ def create(self):
...
+            nonlocal batch
+            utils.dbg('...deleting batch %s..%s' % (
+                batch[0].row_key, batch[-1].row_key))
+            self.bt_table.mutate_rows(batch)
+            batch = []
+        while rows:
+            r = rows.pop()
+            r.delete()
+            batch.append(r)
+            if len(batch) > 50000:
```
amj: what's the 50k?

bigtable_input.py
gitosaurus: The game number, which is the basis of the row ID. Will clarify. The type of game (ordinary or calibration) depends on the `GameQueue` being used; the method is an instance method.

bigtable_input.py
gitosaurus: Oops! That's half the maximum Bigtable mutation batch size. There is a module constant for this, and I'll use it. I was using half the batch size for testing.

PR #621, opened 2018-12-18
Use `multiprocessing.Pool` to speed up game trim

gitosaurus:
Permits multiple maximum-batch requests to be dispatched at once from newly constructed clients in subprocesses.


bigtable_input.py
```diff
@@ -239,28 +269,38 @@ def parse(r):
...
-            r.delete()
-            batch.append(r)
-            if len(batch) >= bigtable.row.MAX_MUTATIONS:
-                mutate_batch()
-        mutate_batch()
+        assert sum([len(b) for t, b in batches]) == total_keys
+        utils.dbg('Deleting %d batches, total of %d keys' % (
+            len(batches), total_keys))
+        # The below is only possible with threaded pool.
+        time_taken = _pool.map(delete_rows, batches)
```
sethtroisi: I don't think time_taken is measuring anything.

andrew has something like
`with dbg_util_something.logged_timer("blah"):`
that helps with measuring time

bigtable_input.py
```diff
@@ -239,28 +269,38 @@ def parse(r):
         return sorted([parse(r) for r in rows], key=operator.itemgetter(0))
 
     def delete_row_range(self, format_str, start_game, end_game):
+        global _pool
+        if _pool is None:
+            _pool = Pool(16)
```
sethtroisi: I'd avoid the global pool and just recreate every time (as this presumably is taking a long time)

gitosaurus: It's true that each call takes awhile (about a minute) so that recreating the pool is not too impactful. I did like the idea of reusing the same pool for an entire trim operation (since delete_row_range is called multiple times through that) but I can refactor it so that the pool is passed in and limited in scope to the trim.

bigtable_input.py
gitosaurus: Removed the per-process timings

bigtable_input.py
gitosaurus: Thanks for the prompt on this! I ended up using a properly scoped `Pool` via the `with` statement, and along the way, set the pool size to be dependent on the number of vCPUs.

bigtable_input.py
```diff
@@ -75,6 +76,11 @@
 _game_row_key = re.compile(r'g_(\d+)_m_(\d+)')
 _game_from_counter = re.compile(r'ct_(\d+)_')
 
+# To parse out project, instance, and table from a Bigtable path like:
+#   'projects/tensor-go/instances/minigo-instance/tables/v14-games'
+_unpack_table_path = re.compile(
+    r'^projects/([^/]+)/instances/([^/]+)/tables/([^/]+)$')
```
brilee: I see bigtable is a very stringly typed API

bigtable_input.py
```diff
@@ -108,7 +114,7 @@ def make_single_array(ds, batch_size=8*1024):
                 batches.append(sess.run(get_next))
         except tf.errors.OutOfRangeError:
             pass
-    return np.concatenate(batches)
+    return np.concatenate(batches) if batches else []
```
brilee: For consistency, return an empty np array of the correct dimension?

bigtable_input.py
```diff
@@ -75,6 +76,11 @@
 _game_row_key = re.compile(r'g_(\d+)_m_(\d+)')
 _game_from_counter = re.compile(r'ct_(\d+)_')
 
+# To parse out project, instance, and table from a Bigtable path like:
+#   'projects/tensor-go/instances/minigo-instance/tables/v14-games'
+_unpack_table_path = re.compile(
```
brilee: Make this a function that does re.match().groups() and returns the value? Then the usage site looks less cluttered.

bigtable_input.py
```diff
@@ -156,6 +162,30 @@ def _game_keys_as_array(ds):
...
+
+    The args are (table_path, row_keys), but are passed as a single
+    argument in order to work with multiprocessing.Pool.map .
+    """
+    table_path, row_keys = args
+    # Unpack project, instance, and name from path.
+    (project_name,
+     instance_name,
+     table_name) = _unpack_table_path.match(table_path).groups()
+    bt_table = bigtable.Client(
```
brilee: does bigtable.Client really take its args like this?... *facepalm*

bigtable_input.py
```diff
@@ -239,28 +269,44 @@ def parse(r):
...
+        batches = []
         batch = []
-        def mutate_batch():
-            nonlocal batch
-            utils.dbg('...deleting batch %s..%s' % (
-                batch[0].row_key, batch[-1].row_key))
-            self.bt_table.mutate_rows(batch)
+        total_keys = len(row_keys)
+        table_name = self.bt_table.name
+        while row_keys:
```
brilee: This block of code is identical to utils.iter_chunks

https://github.com/tensorflow/minigo/blob/master/utils.py#L59

replace with `batches = list(utils.iter_chunks(bigtable.row.MAX_MUTATIONS, row_keys))

brilee: The comments mention that the reverse() is important-ish, but I'm not sure how that works - sholudn't the multiprocessing make order nondeterministic?

bigtable_input.py
```diff
@@ -239,28 +269,44 @@ def parse(r):
...
-            r.delete()
-            batch.append(r)
-            if len(batch) >= bigtable.row.MAX_MUTATIONS:
-                mutate_batch()
-        mutate_batch()
+        assert sum([len(b) for t, b in batches]) == total_keys
+        utils.dbg('Deleting %d batches, total of %d keys' % (
+            len(batches), total_keys))
+        concurrency = multiprocessing.cpu_count() * 2
+        with multiprocessing.Pool(processes=concurrency) as pool:
```
brilee: perhaps cap the concurrency at some sane number that BT can handle? There are some very multi-core machines out there :D

bigtable_input.py
gitosaurus: Yes. The `BigtableClient` doesn't pickle, so it can't be passed to via `multiprocessing.map`. I needed to pass the three parameters required to create the client (project, instance, table), but the client doesn't have inspectors for those. It did have a `name` attribute, in the above format.

The alternative would be for me to have `GameQueue` save these parameters and pass them directly. This might be safer, now that I think about it, because the API doesn't describe the format of the `name` attribute in its docs. It would still be a three-string triple, but I can make it a named tuple, which would help.

bigtable_input.py
gitosaurus: This turned out to be an interesting homework problem! Fixed by limiting the function to require a dataset that produces a one-dimensional array. (It wasn't as general a debugging function as it looked like, because of the behavior of `np.concatenate`).

bigtable_input.py
gitosaurus: Yeah, the `Client` is constructed with a project, from which the caller looks up an instance, from which the caller looks up a table. The unpacking of the args from the path is the part that I can change, per the earlier comment, by simply keeping those three strings in the instance.

bigtable_input.py
gitosaurus: Thanks for the tip! I'll swap that in.

The deleting in reverse is to allow for a more sensible end state if you change your mind (say, due to a mistake in the timestamp) and abort the process: you'll have a little trimmed from the end, rather than a little trimmed out of the middle. I'll add a comment to that effect.

I can reverse the list of keys before calling `utils.iter_chunks` and preserve that behavior.

bigtable_input.py
gitosaurus: Obviated by the change above (use of a named tuple for the Bigtable spec).

bigtable_input.py
gitosaurus: Used a named tuple to store the Bigtable spec.

PR #624, opened 2018-12-21
Turn rl_loop train & validate into a k8s deployment

amj:


mask_flags.py
```diff
@@ -100,7 +100,7 @@ def run(cmd):
     cmd = prepare_subprocess_cmd(cmd)
     print("Running the following cmd:")
     for token in cmd:
```
brilee: you can do `print(' '.join(cmd))` if you want to print without the python list. That being said, I think the list is a more correct representation, because it shows you how the subprocess module passes in the args.

rl_loop/distributed_flags
```diff
@@ -20,6 +20,10 @@
 --sgd_momentum=0.95
 --steps_to_train=1024
 --window_size=500000
+--cbt_project=tensor-go
+--cbt_instance=minigo-instance
+--cbt_table=v14-games
```
brilee: I'm curious whether you could run 2 k8s clusters in the same GCP project - would we run into hopeless collisions everywhere?

rl_loop/train_and_validate.py
```diff
@@ -118,6 +121,8 @@ def validate_holdout_selfplay():
 
 def validate_pro():
     """Validate on professional data."""
+    if not FLAGS.pro_dataset:
+        print("*** --pro_dataset not set, skipping pro validation ***")
```
brilee: add a return here?

brilee: also, maybe put this print in loop() and gate it in loop() so that you don't endlessly print out this message.

PR #625, opened 2018-12-22
Calibrator k8s job

amj:
work up update_resign_threshold to run with a flagfile, enclose it in a dockerfile, and turn it into a deployment.

Everything works except all of it.

### DIFF
--- a/rl_loop/update_resign_threshold.py
+++ b/rl_loop/update_resign_threshold.py
@@ -16,7 +16,9 @@
 
 Reads the bigtables defined by env vars in PROJECT,CBT_INSTANCE,CBT_TABLE to
 compute the 95 percentile of the bleakest-evaluations found in calibration games,
-then updates the flagfile on the default bucket path, resetting that value
+then updates the flagfile on the default bucket path, resetting that value.
+
+Recommended usage is with the common flagfile (e.g. rl_loop/distributed_flags)
 """
 
 import sys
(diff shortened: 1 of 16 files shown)


amj:
Actually it all worked great except for the output wasn't being flushed and so was not showing up.  Hi-larious.

rl_loop/update_resign_threshold.py
```diff
@@ -61,22 +69,32 @@ def update_flagfile(flags_path, new_threshold):
...
 
     while True:
-        new_pct = get_95_percentile_bleak()
+        new_pct = get_95_percentile_bleak(games_nr)
         update_flagfile(fsdb.flags_path(), new_pct)
         time.sleep(60 * 3)
 
 if __name__ == '__main__':
-    app.run(main)
+    valid_flags = list(map(lambda f: '--' + f, FLAGS.flag_values_dict().keys()))
```
brilee: so this jumble here happened because this code directly imports bigtable_input, which isn't part of the rl_loop set of modules. So it's crossing a boundary that's usually bridged via masked_flags.run and subprocesses. If you move the bigtable code here into its own top level module get_resign_rate.py, then use masked_flags.run(), then that'll clean stuff up.

amj: we should probably put a README in rl_loop documenting the boundary's expectations

amj:
retest

PR #626, opened 2018-12-22
Init to loss

amj:
This doesn't need to be merged, but this is the code i'm going to build into the v15 run.


sethtroisi:
superceded by #629

PR #628, opened 2018-12-31
Support init-to-parent, init-to-draw, init-to-loss

sethtroisi:
A more extensible version of #626


cc/mcts_node.h
```diff
@@ -32,6 +32,13 @@
 
 namespace minigo {
 
+// See IncorporateResults for details.
+enum class InitStrategy {
```
sethtroisi: maybe this should be part of MctsNode class?

cc/main.cc
```diff
@@ -87,6 +87,8 @@ DEFINE_bool(soft_pick, true,
 DEFINE_bool(random_symmetry, true,
             "If true, randomly flip & rotate the board features before running "
             "the model and apply the inverse transform to the results.");
+DEFINE_string(init_strategy, "parent",
```
amj: Do we need init to draw?  Would it be better to make this a float from [0,-2] (with clamping, naturally)?

cc/mcts_node.cc
```diff
@@ -304,17 +305,37 @@ void MctsNode::IncorporateResults(absl::Span<const float> move_probabilities,
...
+    //
+    // The initial value of a child's Q is not perfectly understood.
     //
-    // Conversely, if W is winning, then B will explore all 362 moves before
-    // continuing to explore the most favorable move. This is a waste of
-    // search.
+    // There are a couple of general ideas:
+    //   * Init to Parent: Init a new child to it's parent value
+    //      We think of this as saying "The game is probably the same after *any* move
+    //   * Init to Draw (AKA init to zero): Init a new child to 0 AKA "position looks even":
```
amj: AKA "the game looks even on any unexplored subtree"

cc/mcts_node.cc
```diff
@@ -304,17 +305,37 @@ void MctsNode::IncorporateResults(absl::Span<const float> move_probabilities,
...
+    //      We tested this in v11, Because this is how we interpretted the original AGZ paper.
+    //      This doesn't make a lot of sense: The losing player tends to explore every move before
+    //      reading a seond one twice.  The winning player tends to read only the top policy move
```
amj: nit

cc/mcts_node.h
amj: what do you think of a float flag?

cc/mcts_node.cc
```diff
@@ -304,17 +305,37 @@ void MctsNode::IncorporateResults(absl::Span<const float> move_probabilities,
...
+    //      because it has much higher value than any other move
+    //   * Init to Parent minus constant AKA FPU (Leela's approach):
+    //      This outperforms init to parent in eval matches. Leela has used for a long time.
```
amj: maybe also mention LC0's large (1.2) value.  Also missing a 'this'

cc/mcts_node.cc
```diff
@@ -304,17 +305,37 @@ void MctsNode::IncorporateResults(absl::Span<const float> move_probabilities,
...
+    //   * Init to Loss:
```
amj: this is missing a prose explanation, e.g. "This is like saying 'any unexplored subtree is as bad as a loss"

cc/mcts_node.cc
sethtroisi: not true because of policy. Let's get something in and we (READ: you) can clean up wording later if you still care

cc/mcts_node.cc
sethtroisi: done in other PR.

cc/mcts_node.cc
sethtroisi: Done in other PR.

sethtroisi:
Superceded by #629

PR #632, opened 2019-01-03
Turn on init-to-loss in distributed_flags

sethtroisi:


rl_loop/distributed_flags
```diff
@@ -34,3 +34,6 @@
 --num_readouts=800
 --disable_resign_pct=0.00
 --resign_threshold=-0.999
+
+# New in v15
```
amj: nit: i don't think the 'new in' is needed.

PR #635, opened 2019-01-08
Zoo-loop fixes

amj:
-shuffle the list
-don't double count games in ratings.py


ratings/ratings.py
```diff
@@ -295,7 +295,7 @@ def wins_subset(bucket):
             "select model_winner, model_loser from wins "
             "join models where "
             "    models.bucket = ? AND "
-            "    (model_winner = models.id or model_loser = models.id)",
+            "    model_winner = models.id",
```
brilee: use triple quoted string here to avoid bugs with trailing commas? SQL isn't space-sensitive.

amj: not sure what you mean -- there are no trailing commas needed?

brilee: The last comma threw me off somewhat. Had to stare at the diff for 15 seconds to see what was going on.

PR #639, opened 2019-01-09
replace incorrectly named KGS coord with GTP

tommadams:


sethtroisi:
can you make the same change in the python side too?

PR #641, opened 2019-01-10
move Minigui GTP extensions to a new subclass

tommadams:
Adds a `--minigui` flag to `cc:gtp` that controls whether Minigui GTP extensions are enabled or not.

This required some refactoring because some of the Minigui functionality was mixed in with normal GTP command handling:

- Added a `done` field to GtpPlayer::Response that indicates when the `Run` loop should exit.
- `HandleCmd` returns a `Response` instead of writing the GTP command output to `stdout`. This allows the `MiniguiPlayer` subclass to write `__GTP_CMD_DONE_` to `stderr` before the result of the GTP command is written to `stdout`. Previously this was all mixed up in `HandleCmd`.
- Background win-rate evaluation works a bit differently now. Previously, while there were still positions that required further win-rate evaluation, 50% of each ponder batch was devoted to win-rate evaluation. Now, the `MiniguiPlayer` alternates between running full batches of pondering and win-rate evaluation reads, which simplifies the code.
- SGF parsing has been refactored so that the base `GtpPlayer` only parses the main line and the `MiniguiPlayer` processes all variations and prints a bunch of information to `stderr` for each position.


cc/minigui_player.h
```diff
@@ -0,0 +1,152 @@
...
+  // Map from unique ID associated with every position played in a game or
+  // variation to the position's auxiliary info and MctsNode.
+  absl::flat_hash_map<std::string, AuxInfo*> id_to_info_;
+
+  // Queue of positions that require their win rate to be evaluated.
+  std::deque<AuxInfo*> to_eval_;
+
+  // Number of times to perform tree search for each position when evaluating
+  // its win rate.
+  int num_eval_reads_ = 6;
```
amj: 6?

tommadams: With v14, the predicted win rate was still changing noticeably after 4 reads per move, so I increased it a bit. 8 reads per position took a little too long on my GPU, so I split the difference and went with 6.

PR #643, opened 2019-01-11
Move eval mode into a separate binary

tommadams:
I'll rename `cc:main` to `cc:selfplay` in a followup PR.


cc/eval.cc
```diff
@@ -0,0 +1,261 @@
...
+             "Number of virtual losses when running tree search.");
+DEFINE_double(value_init_penalty, 2.0,
+              "New children value initialization penalty.\n"
+              "Child value = parent's value - penalty * color, clamped to"
+              " [-1, 1].  Penalty should be in [0.0, 2.0].\n"
+              "0 is init-to-parent [default], 2.0 is init-to-loss.\n"
+              "This behaves similiarly to Leela's FPU \"First Play Urgency\".");
+
+// Inference flags.
+DEFINE_string(model, "",
```
amj: if parallel_games is 1, which model is black and which is white?

cc/main.cc
```diff
@@ -116,10 +108,7 @@ DEFINE_string(model, "",
               "inferece engine. For engine=tf, the model should be a GraphDef "
               "proto. For engine=lite, the model should be .tflite "
               "flatbuffer.");
-DEFINE_string(model_two, "",
-              "When running 'eval' mode, provide a path to a second minigo "
-              "model, also serialized as a GraphDef proto.");
-DEFINE_int32(parallel_games, 32, "Number of games to play in parallel.");
+DEFINE_int32(parallel_games, 1, "Number of games to play in parallel.");
```
amj: selfplay now defaults to 1?  any particular reason?

amj: sounds like a good plan.

PR #645, opened 2019-01-11
rename cc:main to cc:selfplay

tommadams:


amj:
can you change rl_loop/selfplay.py as well?

tommadams:
done

amj:
LGTM

PR #647, opened 2019-01-11
Add a very simple selfplay example binary

tommadams:


cc/simple_example.cc
```diff
@@ -0,0 +1,95 @@
...
+              "inference engine. For engine=tf, the model should be a GraphDef "
+              "proto. For engine=lite, the model should be .tflite "
+              "flatbuffer.");
+DEFINE_int32(num_readouts, 100,
+             "Number of readouts to make during tree search for each move.");
+
+namespace minigo {
+namespace {
+
+void SimpleExample() {
```
amj: If this is the 'demonstration example' it might be worth giving a preamble or introduction indicating what the overall goal is and what ways the example might be extended (threading, auto-updating things, writing sgfs or examples, whatever).  I'm not sure how much the second part is necessary -- 'negative' comments are weird -- but it might be worth calling out how its different from its larger counterpart.

tommadams: done

PR #656, opened 2019-01-15
Update readme with v14/v15 notes

brilee:
@gitosaurus @sethtroisi @amj  double check for correctness?

### DIFF
--- a/RESULTS.md
+++ b/RESULTS.md
@@ -63,10 +63,10 @@ Q-value](http://computer-go.org/pipermail/computer-go/2017-December/010555.html)
 led to weird ping-ponging behavior inside search, and changed to init to the
 average of the parent.
 
-Around v148, we realized we had the learning rate decay time off by an order of
+Around generation 148, we realized we had the learning rate decay time off by an order of
 magnitude...oops.  We also started experimenting with different values for `c_puct`, but didn't really see a difference for our range of values between 0.8 and 2.0.
 
-Around v165, we changed how many games we tried to play per generation, from
+Around generation 165, we changed how many games we tried to play per generation, from
 12,500 games to 7,000 games, and also increased how many positions each
 generation was trained on, from 1M to 2M.  This was largely due to how we had
 originally read the AGZ paper, where they say they had played 25,000 games with
@@ -74,17 +74,18 @@ the most recent model.  However, with 5M games and 700 checkpoints, that works
 out to about 7,000 games/checkpoint.  As we were doing more of an 'AZ' approach,
 with no evaluation of the new models, we figured this was more accurate.
 
-Around v230, we disabled resignation in 5% of games and checked our
+Around generation 230, we disabled resignation in 5% of games and checked our
 resign-threshold false positive rate.  We'd been pretty accurate, having tuned
 it down to about .9 (i.e., odds of winning at 5%) based on our analysis of true
 negatives.  (i.e., what was the distribution of the closest a player got to
 resigning who ended up winning)
 
-Around v250, we started experimenting with tree reuse to gather data about its
+Around generation 250, we started experimenting with tree reuse to gather data about its
 effect on the Dirchlet noise, and if it was affecting our position diversity.
-We also changed from 'stretching' the probabilities pi by exponentiating to a
-factor of 8 (i.e., temperature 0.125), to making it onehot (infinitesimal
-temperature), to leaving it proportional to visits (unity temperature).
+
+We had also been stretching the probabilities pi by raising to 8th power
+before training on them, as a compromise between training on the one-hot
+vector versus the original MCTS visit distribution.
 
 There's still ambiguity about whether or not a onehot training target for the
 policy net is good or bad, but we're leaning towards leaving pi as the softmax
@@ -96,11 +97,14 @@ the move for play (after the fuseki).
 We put the models on KGS and CGOS as 'somebot' starting around version 160.  The
 harness running these models is very primitive: they are single-threaded python,
 running nodes through the accelerator with a batch-size of ONE... very bad.  The
-CGOS harness only manages to run around 300-500 searches per move.
+CGOS harness only manages to run around 300-500 searches per move (5s/move).
 
 However, with that said, we were able to watch it improve and reach a level
 where it would consistently lose to the top bots (Zen, AQ, Leela 0.11), and
 consistently beat the basic bots, GnuGo, michi, the Aya placeholder bots, etc.
+We found CGOS not particularly useful because of this large gap. Additionally,
+the large population of LeelaZero variants all had similar playstyles, further
+distorting the rating system.
 
 Sometime between v180 and v230, our CGOS results seemed to level off.  The
 ladder weakness meant our performance on CGOS was split: either our opponents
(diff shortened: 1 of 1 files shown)


RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+
+In terms of bottlenecks, our sampler took a big chunk of time to run, but it
+was still shorter than the training time, so it was not technically a bottleneck
+yet. But we were still training on a single TPU (8 cores) rather than a pod
+(128 cores), so if we were to scale to more TPUs for selfplay, and upgrade our
+trainer to use a pod to compensate, then this shuffler would quickly become the
+bottleneck.
+
+For v14, our selfplay workers would instead directly upload each move to BigTable
+as a (position, next_move, winner) tuple. A set of parallel readers would then
```
gitosaurus: The TFExamples stored in the Bigtable row were identical to the TFExamples that were stored in the previous approach. Each TFExample was stored in the Bigtable with a row key based on its game index and move index. Each TFExample contains, I believe, 8 moves of history, just as before.

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+order to ensure a 10% ratio of resign-disabled games to resign-enabled games, we
+used the BigTable input readers to enforce a 90:10 ratio.
+
+Another benefit of the BigTable integration was that we could directly run a
+query on BigTable to compute the appropriate resign threshold. Previously, this
+had been done by manually inspecting a Stackdriver graph and editing a flagfile
+on GCS. This meant that the latency of resign threshold updates dropped from
+hours to minutes.
+
+Overall, time between checkpoints dropped from 30 minutes to 20 minutes. (With
```
gitosaurus: This was fast enough that the trainer had to be modified to wait on each training iteration until a requisite number of "fresh" games from the inference cluster had been added to move the window forward. This did mean that we could be specific about the ratio of fresh to existing games.

(Note that the probabilistic sampler does sample different TFExamples even when sampling games it has sampled before, as expected.)

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+v14 was our first model to be able to consistently read ladders. As a result,
+it was actually not clear whether v14 was actually stronger than v12, or whether
+v14 was just bullying v12 around by reading ladders that v12 couldn't read, and
+winning easy rating points. But since we had achieved success against external
+reference points, we concluded that v14/v15 were in fact really stronger than v12.
+
+During v15's run, we realized that our evaluation cluster infrastructure was in
+need of some upgrades, primarily for two reasons:
+
+1. We wanted to include LZ in the mix, so we could cross-evaluate LZ models with
```
amj: i'd say we want to evaluate against models with a different "genealogy"; it's not LZ specific.

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+our own.
+1. We wanted to be able to run each model with its preferred configuration (init
+to parent for v10/v12, and init to loss for v14/v15.)
```
amj: (..and we can't do this because it's an in-process eval with both sides using the same flags)

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+
+The run was uneventful and was a bit better than previous runs; it escaped random
+play much more quickly, probably because the improved resign threshold latency
+meant that our training data was appropriately pruned of 'easy' positions.
+So all in all, the rewrite didn't create any new pipeline issues.
+
+Around generation 280, we discovered on LCZero forums that AlphaGoZero had used
+init-to-loss for Q values. We switched to init-to-loss from init-to-parent and
+saw our network gain dramatically in strength. Soon, our top model thoroughly
+exceeded our previous best model, and beat LZ's v157, the last model from before
```
amj: maybe mention the sharpness observation here.

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+### v15, clean run with init to loss
+
+This run beat v14 pretty early on, producing a model that was stronger than v14's
+best at generation 330. v15 ended up slightly stronger than v14 but was otherwise
+similar.
+
+Qualitatively, we noticed that V14/V15's models were much sharper,
+meaning that they had more concentrated policy network output and the value
+network output would have stronger opinions on whether it was winning or losing.
+Simultaneously, the calibrated resign threshold for init-to-loss ended up at 0.5
```
amj: 0.5-0.7 compared to 0.8-0.95

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+compared to 0.9 in previous runs. So v14/v15's MCTS was operating in the sweet
```
amj: how do you figure this is the sweet spot?

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+spot more frequently. (The MCTS algorithm tends to go blind once Q gets close to
```
amj: re: blindness.  With a properly high c_puct (2.5ish), there's a lot more room, even with our sharper value.

RESULTS.md
```diff
@@ -63,28 +63,29 @@ Q-value](http://computer-go.org/pipermail/computer-go/2017-December/010555.html)
 led to weird ping-ponging behavior inside search, and changed to init to the
 average of the parent.
 
-Around v148, we realized we had the learning rate decay time off by an order of
+Around generation 148, we realized we had the learning rate decay time off by an order of
```
sethtroisi: consider s/generation/model/

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
 examples without dramatically distorting the ratio of opening/middle/endgame
 examples seen by the network.
 
+This run ended up not really having fixed anything. On our cross-evaluation
```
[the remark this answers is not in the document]
brilee: Yeah, but it wasn't that much stronger, so it was very unexpected that such significant gains remained in MCTS by tweaking init-loss. What you're saying is only evident in hindsight.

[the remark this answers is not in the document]
brilee: The fact that we still couldn't read ladders with bs=2 meant that the single-threaded MCTS was still unable to read ladders out of the box.

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+Previously, we'd write out tiny TFRecords containing ~200
+TFExamples each, and another Python process would rsync the records to one
+machine, read them into memory, and maintain a large moving window of positions
+in memory that would be periodically be spit out as training chunks. This shuffler
+had all sorts of hacks in it - we'd entirely ignore any games with < 30 moves;
+we'd ignore any moves with move number > 400; we'd draw a fixed number (4) of
+moves from each game (which oversampled short games). This hacky fixed-number
+sampling was because exactly 10% of our games were resign-disabled, but those
+resign disabled games were much longer than the other games, meaning that an
+even sampling by move would oversample our resign-disabled games. A fixed-number
```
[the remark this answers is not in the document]
brilee: Hm? I thought we had a 10% enforced split, as measured by moves. I think turning it down from 20% was definitely successful, so sweet spot is somewhere around 10%.

RESULTS.md
```diff
@@ -570,5 +590,94 @@ examples from the first 400 moves should mostly rid us of those pointless
...
+
+Overall, time between checkpoints dropped from 30 minutes to 20 minutes. (With
+one run being 800 checkpoints, this means that we can complete a run in 11 days.)
+We had much more consistent sampling for resign disabled/enabled games and for
+games of varying lengths. We reduced the latency for resign threshold measurements.
+
+The run was uneventful and was a bit better than previous runs; it escaped random
+play much more quickly, probably because the improved resign threshold latency
+meant that our training data was appropriately pruned of 'easy' positions.
+So all in all, the rewrite didn't create any new pipeline issues.
```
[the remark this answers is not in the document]
brilee: what do you mean by aggressive? More timely? Came up with different numbers than our scripts did?

RESULTS.md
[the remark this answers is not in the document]
brilee: Ah, I see what you mean.

RESULTS.md
brilee: we'd previously noticed that MCTS just plays the policy move without reading at all, when it gets close to 0.95

PR #664, opened 2019-01-18
fix value_init_penalty comment

tommadams:


sethtroisi:
https://github.com/tensorflow/minigo/blob/2fc2b77b6fc055dae92c0fbb9d2d995b89a1e8b0/cc/eval.cc
was
also changed to 2.0 without the "[default]" being moved.

PR #665, opened 2019-01-19
Minigui changes to support LeelaZ & gomill-style control files

tommadams:
I've also started adding support for two players: the GTP connection passes a `player` field in the JSON blobs containing the name of the relevant engine.


minigui/control/leelaz.ctl
```diff
@@ -0,0 +1,8 @@
+players = {
+  "leelaz" : Player("../leela-zero/build/leelaz"
```
amj: relative path?

amj: maybe just assume they have it on $PATH

tommadams: These are just example .ctl files. I've added an comment to that effect.

PR #668, opened 2019-01-28
Avoid OOM on large trim operations, show progress

gitosaurus:
The number of batches to give to the multiprocessing pool
were being materialized, which caused it to OOM.  Changed
to create batches at the concurrency level only, and to
use `tqdm` to communicate progress.


gitosaurus:
> LGTM. nice progress bar!

It certainly is a nice progress bar. tqdm lgtm. Thanks for the tip!

PR #672, opened 2019-02-01
New requirements file for upcoming project

sethtroisi:


sethtroisi:
colab has a long list of preinstalled packages :)

PR #675, opened 2019-02-02
Minigui vs mode

tommadams:
Summary of the changes:

 - Add support for GTP command ID
 - Add multiple player support to the Minigui Python server.
 - Vs mode frontend: `minigui/vs.ts` & `minigui/static/vs.html`. There's some code Duplication between this and `minigui/app.ts` that I hope to get rid of later.
 - Remove `variations` GTP command which used to select which variation to send from the backend. The backend now sends the top N variations. The backend no longer explicitly sends a `pv` variation, the frontend just takes the best variation sent.
 - C++ backend sends tree stats at the end of the `genmove` command
 - Update Python engine to current version of the basic Minigui protocol
 - Changed the `Position` TypeScript class so its constructor takes the JSON blob directly, which cleaned a bunch of stuff up.
 - Backend now tells the frontend the board size, rather than the frontend querying for it.
 - Stop passing commandline flags to the control files.
 - Add a `Graph` base class for the TypeScript graph rendering.
 - Make the player's name (from the control file) part of the GTP WebSocket connection ID so that we can route messages correctly between multiple players.
 - Rewrote the `Search` TypeScript board layer to used the new list of top variations that gets sent from the backend.


cc/mcts_node.cc
```diff
@@ -143,14 +143,30 @@ Coord MctsNode::GetMostVisitedMove() const {
...
+            [](const ChildInfo& a, const ChildInfo& b) {
+              if (a.N != b.N) {
+                return a.N > b.N;
+              }
+              if (a.action_score != b.action_score) {
+                return a.action_score > b.action_score;
+              }
+              return a.c > b.c;
+            });
+  return child_info;
```
sethtroisi: nit: indent

cc/mcts_node.h
```diff
@@ -34,13 +34,29 @@ namespace minigo {
 
 class MctsNode {
  public:
+  struct TreeStats {
```
sethtroisi: Much nicer :)

gtp_cmd_handlers.py
```diff
@@ -300,6 +282,36 @@ def _tree_search_wrapper(self, parallel_readouts=None):
                 self._last_report_time = now
         return leaves
 
+    def _minigui_report_position(self):
```
sethtroisi: Is this behind a flag so that it doesn't happen in normal GTP mode?

minigui/control/vs.ctl
```diff
@@ -0,0 +1,25 @@
+players = {
+  "leelaz" : Player("../leela-zero/build/leelaz"
+                    " --weights best-network"
+                    " --timemanage off -r 3"
+		    " --noponder"
```
sethtroisi: nit: indents are messedup

minigui/gtp_socket.ts
```diff
@@ -36,29 +36,28 @@ function trimText(str: string, len: number) {
...
+    this.playerName = playerName;
     this.debug = debug;
     this.sock = io.connect(uri)
 
     this.sock.on('json', (msg: string) => {
       let obj = JSON.parse(msg);
-      if (obj.token != this.gameToken) {
-        console.log('ignoring', obj, `${obj.token} != ${this.gameToken}`);
+      if (obj.token != this.token) {
+        // console.log('ignoring', obj, `${obj.token} != ${this.token}`);
```
sethtroisi: probably remove

amj:
too cool.
headsup-gui

cc/mcts_node.cc
tommadams: That's what Clang Format generates

gtp_cmd_handlers.py
tommadams: yes

minigui/control/vs.ctl
tommadams: done

minigui/gtp_socket.ts
tommadams: done

alreadydone:
Actually people have always wanted a UI that displays two networks/engines' evaluations to complement each other for better judgment, and minigui seems to be the first one that support this!

PR #690, opened 2019-02-04
make Graph beginPath, moveTo, lineTo, stroke methods private

tommadams:
All subclasses now use the `drawPlot` method


tommadams:
Yeah, I know what you mean...

I wasn't happy with the amount of code in the WinrateGraph & ReadsGraph subclasses and was trying to move more functionality into the base class. I'm not sure this actually helped much :)

PR #701, opened 2019-02-08
Add eval_games bigtable

sethtroisi:
We can follow this be setting the bigtable spec for evaluator.


cc/tf_bt_utils.cc
```diff
@@ -141,6 +141,8 @@ void WriteEvalRecord(const std::string& gcp_project_name,
       SetCell("metadata", "length", absl::StrCat(game.moves().size())),
       SetCell("metadata", "sgf", sgf_name), SetCell("metadata", "tag", tag));
 
+  // NOTE: In addition to metadata also eval_games column family.
```
gitosaurus: Incomplete sentence?

cluster/common.sh
```diff
@@ -50,6 +50,7 @@ export BUCKET_LOCATION=${BUCKET_LOCATION:-"asia-east1"}
 export CBT_INSTANCE=${CBT_INSTANCE:-"minigo-instance"}
 export CBT_ZONE=${CBT_ZONE:-"us-central1-b"}
 export CBT_TABLE=${CBT_TABLE:-"games"}
+export CBT_EVAL_TABLE=${CBT_EVAL_TABLE:-"eval_games"}
```
gitosaurus: The two tables right now have the form `v15-games` and `v15-games-nr`. Should this be formed like `v15-games-eval`?

Actually, I believe that we wanted to move to underscores for the table names for some compatibility reasons, and call the non-resign ones "calibration", so maybe the end result should be

  - `v15_games`
  - `v15_games_cal`
  - `v15_games_eval`

cluster/unset-common.sh
```diff
@@ -22,15 +22,21 @@ unset CLUSTER_NAME
...
 
 unset VERSION_TAG
 unset EVAL_VERSION_TAG
+
 unset MINIGUI_PY_CPU_CONTAINER
 
 unset BUCKET_NAME
 unset BUCKET_LOCATION
+
+unset CBT_INSTANCE
```
gitosaurus: Ah, thanks for getting those `unset` commands.

cluster/common.sh
sethtroisi: I think eval is less tightly coupled.

We have maybe ~50-200k rows of data per run + ~10k cross eval + ~5k random things mixed in.

I'm planning to place them all in the same table but with difference key prefix (this remains to be coded).

Ultimately I'd rather write filters on keys than figure out how to join tables.

cc/tf_bt_utils.cc
sethtroisi: removed. I think I'll remember and the whole design is flexible.

sethtroisi:
I ended up some extra changes bundle in this. I think they are low risk but there's a small change that Andrew will see them at some point.

PR #705, opened 2019-02-11
V17 changes

amj:
- fixes a bug where 'freeze' wouldn't read the model flags


rl_loop/distributed_flags
```diff
@@ -8,6 +8,8 @@
 --conv_width=256
 --fc_width=256
 --trunk_layers=19
+--use_SE
```
sethtroisi: Should we set these default true so that the python code runs while v17 is active

rl_loop/distributed_flags_nr
```diff
@@ -5,16 +5,19 @@
 --conv_width=256
 --fc_width=256
 --trunk_layers=19
+--use_SE
```
sethtroisi: Why are these being specified? I don't think this code calls python?

rl_loop/distributed_flags
amj: probably not; most folks will use our v15 models until this has come along farther.

rl_loop/distributed_flags_nr
amj: consistency i guess 🙃

PR #709, opened 2019-02-12
Write eval games to bigtable

sethtroisi:
Builds on #701 and #571


cluster/evaluator/evaluator_cc_wrapper.sh
```diff
@@ -38,7 +38,9 @@ DATE=`date +%Y-%m-%d`
 bazel-bin/cc/eval \
   --model=$BASENAME_BLACK \
   --model_two=$BASENAME_WHITE \
-  --sgf_dir "gs://$SGF_BUCKET_NAME/sgf/eval/$DATE" \
+  --sgf_dir="gs://$SGF_BUCKET_NAME/sgf/eval/$DATE" \
+  --output_bigtable="tensor-go,minigo-instance,eval_games" \
+  --bigtable_tag="$JOBNAME" \
```
amj: i really don't like using the arbitrary k8s-jobname as something we'll rely on later.  As we know, it's not indicative of the actual contents of the job, and so we know it's not a good source-of-truth. Further, there's nothing that checks them -- all k8s cares is that they're unique.

 It might be better (and DRY-er) to construct the tag out of the paths to the two models in the script, and i think we would have better assurances that our metadata would be well-formed.

What do you think?

amj: I guess this is actually just a generic 'tag' field.  Ok sounds fine to me :)

sethtroisi: discussed offline, it's just a tag and it may have use later (or for aggregation in sql queries)

PR #712, opened 2019-02-14
Add an LRU cache for inferences (but don't hook it up yet)

tommadams:
I've plumbed the cache into the `MctsPlayer` and its subclasses but I haven't actually added the code to actually populate or use the cache, since that will probably require some refactoring.


cc/dual_net/inference_cache.cc
```diff
@@ -0,0 +1,91 @@
...
+
+InferenceCache::CompressedFeatures InferenceCache::CompressFeatures(
+    const DualNet::BoardFeatures& features) {
+  CompressedFeatures result;
+  auto num_full_u64s = (features.size() + 63) / 64 - 1;
+  const auto* src = features.data();
+  for (size_t i = 0; i < num_full_u64s; ++i) {
+    uint64_t bits = 0;
+    for (uint64_t j = 0; j < 64; ++j) {
+      bits |= static_cast<uint64_t>(*src++ != 0) << j;
```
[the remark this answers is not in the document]
brilee: does this actually temporarily write the bit into a 64-bit space, shift it, and then OR it in? Or does this compile down to something slightly more reasonable?

cc/dual_net/inference_cache.cc
```diff
@@ -0,0 +1,91 @@
...
+    }
+    result[i] = bits;
+  }
+  uint64_t bits = 0;
+  for (uint64_t j = 0; src != features.end(); ++j) {
```
brilee: is it possible to put `j < 64 && src != features.end()` in the above loop to cut down on the code repitition?

cc/dual_net/inference_cache.h
```diff
@@ -0,0 +1,108 @@
...
+  // The element must have been newly constructed, or previously unlinked from
+  // the list.
+  void PushFront(Element* elem) {
+    elem->prev = &list_;
+    elem->next = list_.next;
+    list_.next->prev = elem;
+    list_.next = elem;
+  }
+
+  // Circular intrusive list sentinel node.
```
brilee: waiiiit a moment... does this mean that all linked lists are basically more cleverly implemented as circular linked lists?

cc/dual_net/inference_cache.cc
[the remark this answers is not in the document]
tommadams: Gcc does more or less what you describe.

Clang fully unrolls the inner loop and generates ~4 instructions per bit. It's all vector intrinsics, which I haven't spent the time to understand.

cc/dual_net/inference_cache.h
tommadams: Dunno :)

There are very few cases where linked lists are an appropriate data structure to use (this is pretty much the only case I'm familiar with). This approach is similar to how std::list was implemented last time I checked (about 15 years ago).

cc/dual_net/inference_cache.cc
```diff
@@ -0,0 +1,91 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cc/dual_net/inference_cache.h"
+
+namespace minigo {
+
+InferenceCache::CompressedFeatures InferenceCache::CompressFeatures(
+    const DualNet::BoardFeatures& features) {
```
sethtroisi: I don't remember if LZ ever implemented this but features (19x19x17) >> zobrist hash (64)

cc/dual_net/inference_cache.cc
```diff
@@ -0,0 +1,91 @@
...
+  // This is a pessimistic estimate of the cache's load factor but since the
+  // size of each node pointer is much smaller than that of the node itself, it
+  // shouldn't make that much difference either way.
+  float load_factor = 0.4375;
+
+  // absl::node_hash_map allocates each (key, value) pair on the heap and stores
+  // pointers to those pairs in the table itself, along with one byte of hash
+  // for each element.
+  float element_size =
+      sizeof(Map::value_type) + (sizeof(Map::value_type*) + 1) / load_factor;
```
sethtroisi: shouldn't this also include the key_size?

tommadams: This is a weirdness in C++ container interfaces: `map<Key, Value>::value_type` is a defined as a `std::pair<Key, Value>` and so does actually include the key size.

I think the rationale behind this decision this is C++ iterators should point to "values", which makes the the `value_type` of an associative container a (key, value) pair, or some kind of proxy for one.

cc/dual_net/inference_cache.cc
tommadams: I originally thought about using the Zobrist hash for the cache key but wasn't happy about effectively ignoring 14 of the feature planes on a cache hit.

However, your comment did give the idea that we can instead use concatenate the most recent 8 Zobrist hashes and 1 bytes for the player. This makes the cache key only 65 bytes instead of 768. I'll add that in a followup PR.

Thanks! 10x impact :)

PR #713, opened 2019-02-15
"sliding window" bigtable training mode

amj:
instead of repeatedly calling train, create repeated datasets by incrementing along the dataset window.

@gitosaurus this is kind of a first cut -- it has to do all the full key retrieval and shuffle before it can start, and it'd be great if i could make that lazy somehow.

### DIFF
--- a/preprocessing.py
+++ b/preprocessing.py
@@ -261,7 +261,7 @@ def get_tpu_bt_input_tensors(games, games_nr, batch_size, num_repeats=1,
     return dataset
 
 
-def get_many_tpu_bt_many_input_tensors(games, games_nr, batch_size,
+def get_many_tpu_bt_input_tensors(games, games_nr, batch_size,
                                        start_at, num_datasets,
                                        moves=2**21,
                                        window_size=500e3,
--- a/preprocessing.py
+++ b/preprocessing.py
@@ -278,12 +278,9 @@ def get_many_tpu_bt_input_tensors(games, games_nr, batch_size,
                                        column_family=bigtable_input.TFEXAMPLE,
                                        column='example')
         ds = ds.repeat(1)
-        if dataset:
-            dataset = dataset.concatenate(ds)
-        else:
-            dataset = ds
+        ds = ds.batch(batch_size)
+        dataset = dataset.concatenate(ds) if dataset else ds
 
-    dataset = dataset.batch(batch_size)
     dataset = dataset.filter(lambda t: tf.equal(tf.shape(t)[0], batch_size))
     dataset = dataset.map(
         functools.partial(batch_parse_tf_example, batch_size))
--- a/preprocessing.py
+++ b/preprocessing.py
@@ -277,8 +277,8 @@ def get_many_tpu_bt_input_tensors(games, games_nr, batch_size,
                                        shuffle=True,
                                        column_family=bigtable_input.TFEXAMPLE,
                                        column='example')
-        ds = dataset.repeat(1)
-        if datasets:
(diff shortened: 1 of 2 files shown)


preprocessing.py
```diff
@@ -261,6 +261,40 @@ def get_tpu_bt_input_tensors(games, games_nr, batch_size, num_repeats=1,
...
+        # timestamps?)
+        ds = games.moves_from_games(start_at + (i * window_increment),
+                                       start_at + (i * window_increment) + window_size,
+                                       moves=moves,
+                                       shuffle=True,
+                                       column_family=bigtable_input.TFEXAMPLE,
+                                       column='example')
+        ds = ds.repeat(1)
+        ds = ds.map(lambda row_name, s: s)
+        dataset = dataset.concatenate(ds) if dataset else ds
```
gitosaurus: Regarding the general approach: if the training loop does multiple scans, I would expect to create a new dataset for each pass, rather than try to create a single enormous dataset, which I imagine would be harder to debug, inspect, etc.

amj: yes, but multiple calls to tpuestimator.train will create new graphs :(  I am not sure what a good solution for lazy evaluating of these Datasets would be.  As it is, it takes a real long time to build the datasets before training even starts -- i suspect the concatenate is doing something bad as things get slower and slower.

preprocessing.py
```diff
@@ -261,6 +261,40 @@ def get_tpu_bt_input_tensors(games, games_nr, batch_size, num_repeats=1,
...
+        # TODO(amj) mixin calibration games with some math. (from start_at that
+        # is proportionally along compared to last_game_number?  comparing
+        # timestamps?)
+        ds = games.moves_from_games(start_at + (i * window_increment),
+                                       start_at + (i * window_increment) + window_size,
+                                       moves=moves,
+                                       shuffle=True,
+                                       column_family=bigtable_input.TFEXAMPLE,
+                                       column='example')
+        ds = ds.repeat(1)
```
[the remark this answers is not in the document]
amj: Done.

preprocessing.py
```diff
@@ -261,6 +261,40 @@ def get_tpu_bt_input_tensors(games, games_nr, batch_size, num_repeats=1,
     return dataset
 
 
+def get_many_tpu_bt_input_tensors(games, games_nr, batch_size,
+                                       start_at, num_datasets,
```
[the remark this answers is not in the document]
amj: Done.

preprocessing.py
```diff
@@ -261,6 +261,40 @@ def get_tpu_bt_input_tensors(games, games_nr, batch_size, num_repeats=1,
...
+                                       moves=2**21,
+                                       window_size=500e3,
+                                       window_increment=25000):
+    dataset = None
+    for i in range(num_datasets):
+        # TODO(amj) mixin calibration games with some math. (from start_at that
+        # is proportionally along compared to last_game_number?  comparing
+        # timestamps?)
+        ds = games.moves_from_games(start_at + (i * window_increment),
+                                       start_at + (i * window_increment) + window_size,
```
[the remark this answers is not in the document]
amj: Done.

train.py
```diff
@@ -132,6 +132,36 @@ def after_run(self, run_context, run_values):
             self.before_weights = None
 
 
+def train_many(start_at=1000000, num_datasets=3):
```
sethtroisi: can you expose moves here also.

amj: what do you mean?  number of steps?

PR #715, opened 2019-02-15
a long overdue update for cc/README.md

tommadams:


cc/README.md
```diff
@@ -9,115 +10,188 @@ build system. We have experienced build issues with the latest version of Bazel,
...
-multiple times:
+The TensorFlow models can found on our public Google Cloud Storage bucket. A
+good model to start with is 000990-cormorant from the v15 run. The C++ engine
+requires the frozen model in a `GraphDef` proto format, which is the `.pb` file.
+You can copy it locally as follows:
 
 ```shell
-bazel test --define=board_size=9 cc/...  &&  bazel test cc/...
+mkdir -p saved_models
+gsutil cp gs://tensor-go-minigo-v15-19/models/000990-cormorant.pb saved_models/
```
brilee: Needs a public bucket link

cc/README.md
```diff
@@ -9,115 +10,188 @@ build system. We have experienced build issues with the latest version of Bazel,
...
-appropriate, however a lot of the implementation details are different for
-performance reasons. In particular, the logic that handles the board state
-strives for a small memory footprint and eschews data structures more
-sophisticated than a simple array in an effort to minimize memory allocations
-and maximize cache locality of data. In doing so, the implementation is kept
-fairly simple (there is no need for a LibertyTracker) and at the time of writing
-performance of the Position code was more than 450x that of its Python
-counterpart.
+Loads all SGF files found in the given `sgf_dir` and tries to predict the move
+made at each position in each game.
```
brilee: what's the output format? Does it edit the sgfs or output a new directory of sgfs with the model's guesses?

cc/README.md
```diff
@@ -9,115 +10,188 @@ build system. We have experienced build issues with the latest version of Bazel,
...
+argument:
+
+ - **tf**: peforms inference using the TensorFlow libraries built by
+   `cc/configure_tensorflow.sh`. Compiled & used as the inference by default,
+   disable the engine with `--define=tf=0`.
+ - **lite**: performs inference using TensorFlow Lite, which runs in software on
+   the CPU.
+   Compile by passing `--define=lite=1` to `bazel build`.
+   Use by passing `--engine=lite` when you run your compiled binary.
+ - **tpu**: perform inference on a Cloud TPU. Your code must be compiled and run
```
brilee: the compilation has to happen on a ctpu too?

cc/README.md
tommadams: I shouldn't really just copy stuff out of my bash history :)

cc/README.md
tommadams: I didn't write this code. It looks like it just prints the results to stderr.

cc/README.md
tommadams: I've removed that mention, though I don't think we've ever tried compiling on a different machine. The model definitely has to be frozen on a Cloud TPU machine.

cc/README.md
```diff
@@ -9,115 +10,189 @@ build system. We have experienced build issues with the latest version of Bazel,
...
- - lite: performs inference using TensorFlow Lite, which runs in software on
-   the CPU.
+Minigo's C++ unit tests operate on both 9x9 and 19x19, and some tests are only
+enabled for a particular board size. Consequently, you must run the tests
+multiple times:
 
-The Compilation and linking of these engines into the `//cc:main` binary is
-controlled by the Bazel defines `--define=tf=<0,1>` and `--define=lite=<0,1>`.
+```shell
+bazel test --define=board_size=9 cc/...  &&  bazel test cc/...
```
amj: maybe either 
1. call out that --define=board_size=19 is the default in the text above
2. explicitly write it here.

tommadams: 1 it is.

The defines don't work how a sane person would expect, so --define=board_size=19 would actually just be completely ignored :P

amj: .... that's amazing.

PR #718, opened 2019-02-15
update Minigui docs

tommadams:


minigui/README.md
```diff
@@ -1,8 +1,24 @@
...
 
-### Simple Instructions
+Minigui runs as a Python web server that you can connect to from your browser
+of choice. We recommend Chrome and occasionally test with Firefox.
+
+Like other UIs, Minigui doesn't contain a Go playing engine and in fact doesn't
+even understand the rules of Go (it relies on the engine to tell it what stones
+are on the board after each move is played).
+
+Minigui communicates to the engine using Go Text protocol
```
amj: nit, capital P

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
-1. Here's an example, note that we specify the board size when compiling the
-   C++ engine binary, not when running the Minigui server:
+### Vs mode
 
-    ```shell
-    bazel build -c opt --define=tf=1 cc:main
-    python minigui/serve.py --port=8888 --model=$MODEL_DIR/$MODEL.converted.pb --engine=tf
-    ```
+A mode that plays two models or engines against each other and displays the
+variations that each player is considering in real time.
```
amj: nit: s/player/engine

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+The JSON object is made up of the following fields (a `?` denotes an optional
+field, other fields are required):
+
+ - `id: string`: an ID that uniquely identifies the current board
+   position.
+ - `parentId?: string`: an ID that uniquely identifies the previous board
+   position (the parent node in the game tree). This field must be set for
+   all board positions except the starting empty board.
+ - `moveNum: number`: the current move number. Used for consistency checking
+   (both Minigui and the engine must agree on the move number).
```
amj: these 'must agree' lines make me wonder what happens if they don't.

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+`mg-position:` prefix.
+
+```
+mg-position: {"gameOver":false,"id":"0x559687366c10","move":"G10","moveNum":36,"parentId":"0x559687364820","stones":"........................................O.O...............XX....X.......X....OXO................XOO................XXO.................XO.................XXOO................OXXO.......................................................................X...............................O......O........X.X.O.......O..X....XOO.........................................","toPlay":"B"}
+```
+
+#### Update messages
+
+While an engine is thinking (in response to a `genmove` command or pondering
+while it's the other player's turn), an engine should also send incremental
```
amj: can or should?

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+rate, number of visits, etc.
+
+These updates should be printed as a JSON object (again on a single line) with
+the prefix `mg-update:`.
+
+ - `id: string`: the position's id.
+ - `n?: number`: the number of reads for the position.
+ - `q?: number`: estimated winrate for the position from black's perspective in
+   the range [-1, 1].
+ - `pv?: string`: the GTP coordinate of the variation in `variations` that
```
amj: Nit: consider adding something like "E.g., the coordinate "K10" for the variation beginning at K10..."

amj: esp as the PV is missing from below.

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+### Lightweight Demo mode
+
+A lightweight demo mode built as experiment for running Minigo on a Raspberry
+Pi.
+
+### Kiosk mode
+
+A self-play only version of the demo mode.
+
+
```
sethtroisi: nit remove one line here

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+The Minigui frontend doesn't understand the rules of Go. Consequently, every
+time a move is played or a new game is started, the engine must communicate the
+game state to the frontend. This is done by printing a JSON object on a single
+line to `stderr` with the prefix `mg-position:`.
+
+The JSON object is made up of the following fields (a `?` denotes an optional
+field, other fields are required):
+
+ - `id: string`: an ID that uniquely identifies the current board
+   position.
```
sethtroisi: "(we use can be syzgygygygy hash)" (if that's true)

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+ - `gameOver?: boolean`: true or false. If `gameOver` is not set, Minigui
+   assumes that the game isn't over.
+ - `move?: string`: the move that led to this position, formatted as a GTP
+   coordinate. This field must be set for all board positions except the
+   starting empty board.
+ - `comment?: string`: any comments for this position (perhaps from a loaded
+   SGF).
+ - `caps?: number[]`: the number of captured stones. Black captures should be
+   stored in `caps[0]` and white captures should be stored in `caps[1]`. If no
+   stones have been captured, the number of captures for both players is assumed
```
sethtroisi: If caps is not provided,

minigui/README.md
sethtroisi: seperator?

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+
+### Synchronizing stdout and stderr
+
+All `mg-position` and `mg-update` generated in response to a GTP command must be
+printed and flushed to `stderr` because GTP mandates what can and cannot be
+written to `stdout`.
+
+When processing a GTP command, the engine must ensure that both `stdout` and
+`stderr` stay in sync by printing and flushing the string `__GTP_CMD_DONE__` to
+`stderr` before it writes the GTP response to `stdout`.
```
sethtroisi: can you give an example of this.

e.g.

genmove
stderr: aoeu
stderr: aoeu
stderr: __GTP_CMD_DONE__
stdout: D4

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
+
+### GTP extensions
+
+ - **echo**
```
sethtroisi: @amj 
per GTP I think you're supposed to preference these with "mg-"

minigui/README.md
tommadams: done. actually the updates don't even have a `pv` field any more but I forgot to update the interface definition the typescript code

minigui/README.md
tommadams: can

minigui/README.md
tommadams: Removed. Currently, the frontend will throw an exception that gets written to the console and ignore the message. It's probably not worth getting into that discussion here.

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
 
 1. Run `./fetch-and-run.sh`
 
-1. open localhost:5001 (or whatever value you used for $MINIGUI_PORT).
+1. open localhost:5001 (or whatever value you used for $MINIGUI\_PORT).
```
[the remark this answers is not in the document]
tommadams: 5001. I removed the reference to 8888

minigui/README.md
```diff
@@ -53,36 +69,247 @@ A UI for Minigo
...
-1. Before running, you'll need to freeze a model so the C++ job can consume it.
-   This assumes a converted model from above.
+Study mode allows you to load previously recorded SGF games, explore variations
+and explore what the engine thinks of each position.
 
-    ```shell
-    python main.py freeze-graph $MODEL_DIR/$MODEL.converted
-    ```
+**Minigui's study mode currently requires the C++ Minigo engine for full
+functionality.**
```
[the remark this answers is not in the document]
tommadams: It does basically run using the python engine and I have a branch which is adding more functionality, but I'll avoid mentioning that for now

minigui/README.md
tommadams: actually, I realized a couple of days ago that the Zobrist hash of the stones won't work for study mode because multiple variations could lead to the same board state, so it's no longer a unique ID

minigui/README.md
tommadams: From the spec:

> Engines are allowed to use private extensions without a dash in the name, but then they do it at their own risk and must be prepared to change if the name later becomes used for a standard command.

I'll file an issue to update the commands.

FWIW, pachi also has an `echo` command

PR #726, opened 2019-02-22
C++ clean up

tommadams:
- Fix up build dependencies so everything compiles with strict dependency checking. These are all trivial changes to BUILD files.
 - Stop adding an `MctsNode` for resign moves to the game tree (`cc/mcts_player.cc`). This was actually resulting in undefined behaviour because the `MctsNode` stats member is initialized to `&edges[move]` of it's parent. When `move` is `Coord::kResign`, this address is off the end of the `edges` array :( We never dereference the invalid pointer to my knowledge so I don't think this will have caused any "real" bugs.
 - As a result of the game tree no longer having a node for resign, I've had to rewrite all the game over handling code in Minigui.
 - I've also updated `tools/bazel.rc` to run all tests using debug binaries so that our `DCHECKS` actually get checked and not compiled away.

I've manually tested `cc:selfplay` and all the Minigui modes.


minigui/control/minigo_tf.ctl
```diff
@@ -9,7 +9,7 @@ players = {
                        " --value_init_penalty=0"
                        " --courtesy_pass=true"
                        " --virtual_losses=8"
-                       " --resign_threshold=-0.8",
+                       " --resign_threshold=-0.3",
```
amj: revert?

PR #730, opened 2019-02-23
MLPerf reference implementation

tommadams:
This is based on Christian's work but brought up to date with our latest codebase: https://github.com/chsigg/training/blob/19x19/reinforcement/reinforcement.py

Christian's original fork ran eval against LeelaZero but I've dropped that in favor of training against a target model inside Minigo for simplicity. I picked a random 9x9 model I had on my machine to use as a placeholder for the target model (000496-polite-ray) and put it on GCS: gs://tensor-go-ml-perf/models/9x9/target

I've verified that the RL loop runs for a couple of iterations but my GPU is too wimpy to train a full model in a reasonable amount of time.

Sending this out early so we can have a discussion about what we want the real RL loop to look like.


ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+Note that MiniGo is an independent effort from AlphaGo, and that this fork is minigo is independent
+from minigo itself.
+
+### Reinforcement Setup
+This benchmark includes both the environment and training for 9x9 Go. There are four primary phases
+in this benchmark, these phases are repeated in order:
+
+ - Selfplay: the *current best* model plays games against itself to produce board positions for
+   training.
+ - Training: train the neural networks using selfplay data from recent models. We are not using the
```
sethtroisi: The neural network weights are updateds from recent selfplay games

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+   selfplay data from the current iteration so that selfplay and the rest of the iteration may run
```
sethtroisi: consider dropping the "we are not using..."

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+   in parallel.
+ - Model Evaluation: the *current best* and the most recently trained model play a series of games.
+   In order to become the new *current best*, the most recently trained model must win 55% of the
+   games.
+ - Target Evaluation: if the newly trained model has been promoted to the current best, play a series
```
sethtroisi: s/has been/is

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+   of games against a previously trained model. The termination criteria for the benchmark is to win
```
sethtroisi: Specify what "previously trained model" is. Do you mean from earlier this run, from us, from gnugo...

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+### Structure
+This task has a non-trivial network structure, including a search tree. A good overview of the
+structure can be found here: https://medium.com/applied-data-science/alphago-zero-explained-in-one-diagram-365f5abf67e0.
+
+### Weight and bias initialization and Loss Function
+Network weights are initialized randomly. Initialization and loss are described here;
+["Mastering the Game of Go with Deep Neural Networks and Tree Search"](https://www.nature.com/articles/nature16961)
+
+### Optimizer
+We use a MomentumOptimizer to train the primary network.
```
sethtroisi: "primary"?

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+# 4. Quality
+Due to the difficulty of training a highly proficient Go model, our quality metric and termination
+criteria is based on winning against another Go progream at intermediate amateur strength.
```
sethtroisi: this doesn't match above

sethtroisi: s/progream/program

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+from rl_loop import example_buffer, fsdb, shipname
+
+flags.DEFINE_string('engine', 'tf', 'Engine to use for inference.')
+
+FLAGS = flags.FLAGS
+
+
+class State:
+
+  _NAMES = ['bootstrap'] + random.Random(0).sample(shipname.NAMES,
```
sethtroisi: why don't we just use them alphabetically?

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+      '--engine={}'.format(FLAGS.engine),
+      '--virtual_losses=8',
+      '--seed={}'.format(state.seed),
+  ]
+
+
+def py_flags(state):
+  return [
+      '--work_dir={}'.format(fsdb.working_dir()),
+      '--trunk_layers=10',
```
sethtroisi: consider moving to constants at the top.

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+    return item
+
+
+make_slice = MakeSlice()
+
+
+def cc_flags(state):
+  return [
+      '--engine={}'.format(FLAGS.engine),
+      '--virtual_losses=8',
```
sethtroisi: I think this should be a constant with explanation at the top as it's one factor that will vary with GPU/CPU/TPU

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # These could run in parallel.
+    validate(state, holdout_glob)
+    model_win_rate = evaluate_model(state)
+    target_win_rate = evaluate_target(state)
+
+    # This could run in parallel to the rest.
+    selfplay(state)
+
+    if model_win_rate >= 0.55:
```
sethtroisi: if you're only playing 100 games this should be higher but IDK if it matters

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+      logging.FileHandler(os.path.join(FLAGS.base_dir, 'reinforcement.log')))
+  formatter = logging.Formatter('[%(asctime)s] %(message)s',
+                                '%Y-%m-%d %H:%M:%S')
+  for handler in logging.getLogger().handlers:
+    handler.setFormatter(formatter)
+
+  with utils.logged_timer('Total time'):
+    for target_win_rate in rl_loop():
+      if target_win_rate > 0.5:
+        return logging.info('Passed exit criteria.')
```
sethtroisi: log what model it was

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+# Self-play a number of games.
+def selfplay(state):
+  play_output_name = state.play_output_name
+  play_output_dir = os.path.join(fsdb.selfplay_dir(), play_output_name)
+  play_holdout_dir = os.path.join(fsdb.holdout_dir(), play_output_name)
+
+  result = checked_run([
+      'bazel-bin/cc/selfplay', '--parallel_games=2048',
```
brilee: that's a lot of parallel games :D
I guess low readouts means you can support them all. Seems like too few readouts to make this work properly, but I guess that's up to MLPerf guys to optimize.

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+  ] + cc_flags(state), 'selfplay')
+  logging.info(get_lines(result, make_slice[-2:]))
+
+  # Write examples to a single record.
+  pattern = os.path.join(play_output_dir, '*', '*.zz')
+  logging.info('Extracting examples from "{}"'.format(pattern))
+  random.seed(state.seed)
+  tensorflow.set_random_seed(state.seed)
+  numpy.random.seed(state.seed)
+  buffer = example_buffer.ExampleBuffer(sampling_frac=1.0)
```
brilee: wow, this code lives on. I'm slightly horrified. It's also on the main path, so its speed actually matters too.

brilee: Another thing I disliked about the examplebuffer was that it was set up for in-process instead of subprocessed. if this whole thing is going to be run continuously in one python process, the memory usage of the EB could be less than ideal.

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+  buffer.parallel_fill(tensorflow.gfile.Glob(pattern))
```
brilee: You've gone through the trouble to set up the seeds, but parallel fill's implementation uses subprocesses, which is definitely going to be nondeterministic. FWIW, if you set up a TF dataset that reads in parallel over these files, it provides deterministic read behavior over the many files.

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+  play_output_dir = os.path.join(fsdb.selfplay_dir(), play_output_name)
+  play_holdout_dir = os.path.join(fsdb.holdout_dir(), play_output_name)
+
+  result = checked_run([
+      'bazel-bin/cc/selfplay', '--parallel_games=2048',
+      '--num_readouts=100', '--model={}.pb'.format(
+          state.play_model_path), '--output_dir={}'.format(play_output_dir),
+      '--holdout_dir={}'.format(play_holdout_dir)
+  ] + cc_flags(state), 'selfplay')
+  logging.info(get_lines(result, make_slice[-2:]))
```
brilee: you can construct this as slice(-2, None, None)
```
>>> x =  slice(-2, None, None)
>>> y = 'blah'
>>> y[x]
'ah'
```

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+
+# Train a new model.
+def train(state, tf_records):
+  result = checked_run([
+      'python3',
+      'train.py',
+      *tf_records,
+      '--export_path={}'.format(state.train_model_path),
+  ] + py_flags(state), 'training')
+  logging.info(get_lines(result, make_slice[-8:-8]))
```
brilee: is -8:-8 intentional? that's a slice of size 0.

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+    # This could run in parallel to the rest.
+    selfplay(state)
+
+    if model_win_rate >= 0.55:
+      # Promote the trained model to the play model.
+      state.play_model_num = state.train_model_num
+      state.play_model_name = state.train_model_name
+      state.train_model_num += 1
+    elif model_win_rate < 0.4:
+      # Bury the selfplay games which produced a significantly worse model.
```
brilee: not sure this is necessary given what we've seen with AZ-style runs, but /shrug

ml_perf/reference_implementation.py
```diff
@@ -0,0 +1,279 @@
...
+
+  with utils.logged_timer('Total time'):
+    for target_win_rate in rl_loop():
+      if target_win_rate > 0.5:
+        return logging.info('Passed exit criteria.')
+    logging.info('Failed to converge.')
+
+
+if __name__ == '__main__':
+  sys.path.insert(0, '.')
```
brilee: this is already at the top?

ml_perf/reference_implementation.py
brilee: should be 60/100 for 95% confidence that it's better.

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+    # Install python dependencies
+    pip3 install -r requirements.txt
+
+    # Install bazel
+    wget https://github.com/bazelbuild/bazel/releases/download/0.17.1/bazel-0.17.1-installer-linux-x86_64.sh
+    chmod +x bazel-0.17.1-installer-linux-x86_64.sh
+    ./bazel-0.17.1-installer-linux-x86_64.sh
+
+    # Compile TensorFlow
+    ./cc/configure_tensorflow.sh
```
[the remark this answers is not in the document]
tommadams: Yeah, I understand how much of a pain point this can be.

I'm going to attempt to rewrite `configure_tensorflow.sh` this week to pull in TensorFlow via Bazel's WORKSPACE mechanism (but still have it as a separate pre-build step rather than a direct dependency). So I'm going to hold off adding more documentation until that rewrite is done (or I fail and give up).

ml_perf/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Steps to download and verify data
+Unlike other benchmarks, there is no data to download. All training data comes from games played
+during benchmarking.
+
+# 3. Model
+### Publication/Attribution
+
+This benchmark is based on the minigo project (https://github.com/tensorflow/minigo); which is
+inspired by the work done by Deepmind with ["Mastering the Game of Go with Deep Neural Networks and
```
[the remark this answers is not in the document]
tommadams: done

ml_perf/README.md
tommadams: I'm not sue we've actually decided what the target model should be. I've put "a target model that was previously trained via this reference benchmark" for now, which what I believe is the simplest approach: selfplay against a previously trained Minigo model can be easily done in parallel, while playing against a different engine like Leela or GnuGo over GTP requires adding extra support for parallel playing of games over GTP (see the [GtpEvaluator](http://go/gh/chsigg/training/blob/19x19/reinforcement/minigo.patch) that Christian had written in his original version of the benchmark).

ml_perf/reference_implementation.py
tommadams: added a TODO

ml_perf/reference_implementation.py
tommadams: It's a shrug from me too, I'm leaving 0.55 for now. 
Once I've got this running on a VM with 8 v100s, we can run some experiments to determine better values

ml_perf/reference_implementation.py
tommadams: added a TODO

ml_perf/reference_implementation.py
tommadams: yeah, Christian was running 9x9 on 8 v100s, which required a large number of parallel games to get decent utilization, as you might imagine.

We will want to support playing N games with M<N in parallel, so that games will finish quicker for 19x19 and maybe also help with the ragged edge problem.

ml_perf/reference_implementation.py
tommadams: added a TODO to pipe a seed to the subprocesses

ml_perf/reference_implementation.py
tommadams: lol, probably not. I've removed the first -8 for now. I'll figure the correct numbers once I have the the RL loop running for real.

ml_perf/reference_implementation.py
tommadams: added a TODO to figure this out

tommadams:
Thanks for all the comments, I've addressed most of them (even if it's merely to add a TODO).

tommadams:
Also FYI, I didn't write the README; that was all Christian's work, which is why I didn't have a good answer to some of the questions :)

PR #737, opened 2019-02-28
correctly handle soft-pick when no points have any reads

tommadams:


cc/mcts_player.cc
```diff
@@ -207,6 +207,15 @@ Coord MctsPlayer::PickMove() {
   for (size_t i = 1; i < cdf.size(); ++i) {
     cdf[i] += cdf[i - 1];
   }
+
+  if (cdf.back() == 0) {
+    // It's actually possible for an early model to put all its reads into pass,
+    // in which case the SearchSorted call below will always return 0. In this
+    // case, we'll just let the model have its way and allow a pass. Yes, this
```
amj: you can probably strip the 'yes this actually happened' comment ;)

PR #742, opened 2019-03-04
MLPerf reference implementation updates

tommadams:
- Run bootstrap games using a random fake model: each selfplay worker uses a different random seed, so this will produce a more diverse set of initial random games than all workers sharing a single randomly initialized model.
- Update `train.py` to optionally freeze the model at the end of training. This removes the explicit freeze step because that can take up to 30 seconds on a GCE VM with 8 GPUs (it looks like TensorFlow does several seconds of work for each attached GPU during initialization but I have yet to run a profile).
- Stop burying bad models, since the rationale for doing this is questionable: if we're having trouble training a model that can beat the previous best, we actually _do_ want to update the training chunks, since otherwise we're not getting any new data that could possibly get us unstuck.
- Remove shipnames: models are named by the RL loop's iteration number & the model generation number (how many models passed gating), e.g. `000004-000003`.
- Suppress some TF deprecation logging that was polluting the training logs.

I'm currently in the middle of a 9x9 training run. The model looks to be overfitting with the current hyperparameters but after 10 generations it had definitely learned _something_.


evaluate.py
```diff
@@ -86,6 +86,7 @@ def play_match(black_model, white_model, games, sgf_dir):
             if active.is_done():
                 fname = "{:d}-{:s}-vs-{:s}-{:d}.sgf".format(int(time.time()),
                                                             white_name, black_name, i)
+                active.set_result(active.root.position.result(), was_resign=False)
```
amj: what was this for?

ml_perf/reference_implementation.py
```diff
@@ -18,59 +18,53 @@
...
 
-  @property
-  def play_model_path(self):
-    return os.path.join(fsdb.models_dir(), self.play_model_name)
+    self.best_model_name = 'random'
 
   @property
-  def train_model_name(self):
-    return '%06d-%s' % (self.iter_num, self._NAMES[self.train_model_num])
+  def output_model_name(self):
```
amj: around here might be a good place for a comment about the naming scheme.

train.py
```diff
@@ -47,6 +47,9 @@
                   'Whether to use Bigtable as input.  '
                   '(Only supported with --use_tpu, currently.)')
 
+flags.DEFINE_bool('freeze', False,
```
amj: this is funny; should we plumb this into the rl_loop/ usage?

evaluate.py
tommadams: the result string doesn't get set otherwise

train.py
tommadams: Probably.

Assuming we pay a similar ~30 second TF init cost on TPU to what I see on my 8 GPU GCE VM, then freezing in the same binary as training would save ~8 hours in a 1000 model run.

ml_perf/reference_implementation.py
tommadams: added it to the top of the State class

PR #743, opened 2019-03-06
MLPerf: add flagfiles, remove eval games against target 

tommadams:
Eval games against target will be played out-of-band at the end of the training run.


ml_perf/flags/9/train.flags
```diff
@@ -0,0 +1,6 @@
+# Flags for training & validation.
+
+--conv_width=64
+--fc_width=64
+--trunk_layers=10
```
amj: trunk_layers=9

ml_perf/reference_implementation.py
```diff
@@ -225,28 +244,24 @@ def rl_loop():
     #     generations.
     #   - uniformly resample the window each iteration (see TODO in selfplay
     #     for more info).
-    tf_records = get_golden_chunk_records(min(5, state.iter_num));
+    tf_records = get_golden_chunk_records(min(5, state.iter_num))
```
amj: can we extract this to a flag?  `window_size` or similar.

tommadams:
Addressed comments. I also changed `gating_win_rate` to `0.5` as we discussed yesterday.

instinct79:
Are the files for flags/19/ put as a placeholder for now ? flags/19/train.flags shows conv_width=64 while dual_net.py has that set to 256 for 19x19.

-Nitin.

amj:
yes, the flags/19 are a placeholder, sorry!

PR #744, opened 2019-03-07
fix crash after loading an SGF that ends in consecutive passes

tommadams:
oof, this is a bit of a mess.

When loading an SGF, we read all variations in the game, then reset the root node back to the start of the tree via `MctsPlayer::ResetRoot` and then play out the main line. The current state of the game is tracked by a `Game` object that previously `MctsPlayer` would update in some of its methods (`PlayMove`, `UndoMove`, `UpdateGame`).

The crash was caused by the fact that `ResetRoot` wasn't resetting the `Game` state back to an empty position.

Instead of plumbing a `Game` argument through more methods, I've changed the `MctsPlayer` to hold a pointer to the `Game` being played and a `Game` instance is now required to create an `MctsPlayer`. 

This adds a wrinkle in that in eval games, two `MctsPlayers` both have pointers to the same `Game` and both call `PlayMove`. To avoid all moves being duplicated, `Game` now looks for duplicate moves when its `ignore_repeated_moves` option is `true` and discards them. I've yet to come up with a better solution :/

Once nice result of this change is that `MctsPlayer::Options` now no longer has to keep a duplicate copy of the `Game::Options`, which suggests this is a somewhat better design.


amj:
Terrible shared ownership idea:  Can `Game` have pointers to its players?  if there's two, yay, if there's one, done, if there's none, game is over.

not clear if that's any cleaner than switching on a flag.

tommadams:
I think that will work. We can add a `RegisterPlayer` method to `Game` that the player calls in its constructor and pass the player in as an additional argument to `Game::AddMove`. I'm going to do more mlperf work today, so I'll push this PR now because it fixes Minigui and I think the changes required to `Game` will be localized.

PR #745, opened 2019-03-07
Fix tensorrt engine build

godmoves:
I met some problems when I was building the minigo trt engine, then I did this fix and finally got the engine to work.

BTW: I also met the error of missing `cuda_runtime_api.h` and I fixed it by adding `"@local_config_cuda//:cuda_headers"` and `"@local_config_cuda//:cudart"` into the tensorrt part of the `BUILD` file, but I am not familiar with Bazel, so I don't know if that is the right way to fix, so I don't include that in the PR

### DIFF
--- a/cc/BUILD
+++ b/cc/BUILD
@@ -87,6 +87,8 @@ cc_library(
     name = "tensorrt",
     tags = ["manual"],
     deps = [
+        "@local_config_cuda//:cuda_headers",
+        "@local_config_cuda//:cudart",
         "@local_config_tensorrt//:nv_infer",
         "@local_config_tensorrt//:nvparsers",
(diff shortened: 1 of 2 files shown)


godmoves:
I signed it!

tommadams:
Thanks for the fix, it's greatly appreciated!

None of the core Minigo team members have TensorRT setup on our machines so we haven't been able to maintain the code as we would have liked. I'm planning on getting TensorRT set up soon, so hopefully this won't be a problem for much longer :)

Please also add your `BUILD` file changes to your pull request and then I'll trigger the integration tests.

godmoves:
@tommadams I've updated the BUILD file :)

tommadams:
Thanks again for the fix!

PR #747, opened 2019-03-08
Fix flags for ml perf

amj:
- don't train on every example
- increase the window side
- fix up reads for bootstrap
- other stuff.

### DIFF
--- a/ml_perf/flags/19/train.flags
+++ b/ml_perf/flags/19/train.flags
@@ -3,4 +3,4 @@
 --conv_width=64
 --fc_width=64
 --trunk_layers=5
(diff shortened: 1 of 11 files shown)


ml_perf/flags/19/train.flags
```diff
@@ -3,4 +3,4 @@
 --conv_width=64
 --fc_width=64
 --trunk_layers=5
---value_cost_weight=0.25
+--value_cost_weight=0.125
```
tommadams: --filter_amount=0.1

amj: Done.

PR #754, opened 2019-03-08
[ml_perf]Better flags organization

delock:
Move common parts of flags into separate files, and avoid same flags appear twice by inclusion.
* play.flags -- flags control the policy of play, used by both selfplay and evaluation
* topology.flags -- describe model topology
* Separate validation.flags for validation

### DIFF
--- a/ml_perf/flags/9/eval.flags
+++ b/ml_perf/flags/9/eval.flags
@@ -1,7 +1,7 @@
 # Flags for playing eval games.
 
 # Import the base selfplay flags.
---flagfile=ml_perf/flags/9/selfplay.flags
+--flagfile=ml_perf/flags/9/play.flags
 
 # Play fewer games for eval than selfplay.
 --parallel_games=100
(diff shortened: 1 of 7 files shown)


amj:
We want to keep the flags to one-per-phase for readability, and there are already checks in place to remove duplicate flags from multiple sources.  I don't think this is the right approach.

tommadams:
To expand a little on what Andrew said, separating the validate flags is a good idea but there isn't really a need for a separate play.flags: it's fine for the bootstrap and eval flags to "inherit" from selfplay.

delock:
> To expand a little on what Andrew said, separating the validate flags is a good idea but there isn't really a need for a separate play.flags: it's fine for the bootstrap and eval flags to "inherit" from selfplay.

Thanks, I saw your fix on the flags.  I think I'm okay with that.  I'll do some test to see whether it will pass on CPU environment.   I'll close this PR.

PR #758, opened 2019-03-09
add support for out-of-band model evaluation

tommadams:
- Append a (timestamp, model_name) pair to a train_times.txt file in the models directory.
- Add an `eval_models.py` script that runs eval games for each trained model against the target model (also in the models directory).


delock:
Should add a scripts to do training and evaluating as a single step.
The training can pass midnight so the following two commands may not always work.
```
    # Run training loop
    BOARD_SIZE=9  python  ml_perf/reference_implementation.py \
      --base_dir=$(pwd)/results/$(date +%Y-%m-%d) \
      --flagfile=ml_perf/flags/9/rl_loop.flags

    # Once the training loop has finished, run model evaluation to find the
    # first trained model that's better than the target
    BOARD_SIZE=9  python  ml_perf/eval_models.py \
      --base_dir=$(pwd)/results/$(date +%Y-%m-%d) \
      --flags_dir=ml_perf/flags/9
```
Better change to something like
```
    TRAIN_FOLDER_NAME=$(date+$Y-%m-%d)
    # Run training loop
    BOARD_SIZE=9  python  ml_perf/reference_implementation.py \
      --base_dir=$(pwd)/results/$(TRAIN_FOLDER_NAME) \
      --flagfile=ml_perf/flags/9/rl_loop.flags

    # Once the training loop has finished, run model evaluation to find the
    # first trained model that's better than the target
    BOARD_SIZE=9  python  ml_perf/eval_models.py \
      --base_dir=$(pwd)/results/$(TRAIN_FOLDER_NAME) \
      --flags_dir=ml_perf/flags/9
```

ml_perf/reference_implementation.py
```diff
@@ -180,28 +189,31 @@ def validate(state, holdout_glob):
...
+      '--model_two={}.pb'.format(target_model_path),
       '--sgf_dir={}'.format(sgf_dir),
-      '--seed={}'.format(state.seed))
+      '--seed={}'.format(seed))
   result = get_lines(result, make_slice[-7:])
   logging.info(result)
   pattern = '{}\s+\d+\s+(\d+\.\d+)%'.format(eval_model)
   win_rate = float(re.search(pattern, result).group(1)) * 0.01
-  logging.info('Win rate %s vs %s: %.3f', eval_model, best_model, win_rate)
+  logging.info('Win rate %s vs %s: %.3f', eval_model, target_model, win_rate)
```
amj: might be a nice place to put a big PASS or FAIL in stdout.

tommadams: Pass or fail depends on the threshold, which isn't know at this point (and is different for gating and the exit criteria)

PR #759, opened 2019-03-09
[ml_perf]Bag of tricks to improve training convergence

delock:
This change added two tricks to improve convergence, especially when batch size is large (4096).
1. When a model is about to be promoted, check it's selfplay bias, if the model is highly biased, it will be hard to improve the new model use these biased data, so we will give up promote this model and keep on play with the old model.
2. When a model win over old model overwhelmingly, the data generated by old model is considered of little help to new model, or even misleading, so new rounds of training will be from data generated by new model only.


delock:
The effect of tricks with large training batch size (4096).  There was a trick for model winrate <0.4 but I saw that part is gone...
bag-of-tricks

amj:
the implementation here (opening all .sgfs and searching for 'w+' and 'b+') is obsolete via the changes in #766 .  Closing.

PR #761, opened 2019-03-11
MLPerf flag tweaks

amj:
Here's what i'm running now


ml_perf/flags/9/selfplay.flags
```diff
@@ -5,10 +5,12 @@
 
 --engine=tf
 --num_games=2048
```
tommadams: Should `num_games` be 4096?

ml_perf/flags/9/bootstrap.flags
```diff
@@ -7,6 +7,4 @@
 
 # Don't perform holdout for the first bootstrap round.
 --holdout_pct=0
-
-# Don't look for signal in what is known to be noise.
---num_readouts=50
+--num_readouts=250
```
tommadams: is there a reason `num_readouts` is larger than the 160 used for selfplay?

ml_perf/flags/9/selfplay.flags
```diff
@@ -5,10 +5,12 @@
...
---virtual_losses=8
---value_init_penalty=2
+--value_init_penalty=0.2
 --holdout_pct=0.03
 --disable_resign_pct=0.1
 --resign_threshold=-0.99
+
+# Device-specific selfplay flags.
+--parallel_games=2048
+--virtual_losses=8
```
tommadams: Is `virtual_losses` going to be fixed for MLPerf or are participants free to tune it? The comment suggests the latter, but I would have expected the former to be the case, since it effects the read depth.

amj: it effects read depth less than we might think with tree reuse.  I think it should be tunable; the actual effect on the pipeline seems very small -- v12 vs v10 we only changed the vlosses and it had a very mild effect on overall strength.  I think we may want to specify a maximum;

ml_perf/flags/9/train.flags
```diff
@@ -1,7 +1,12 @@
-# Flags for training.
+# train.flags: Flags for training.
 
 --flagfile=ml_perf/flags/9/architecture.flags
 
 --shuffle_buffer_size=10000
---filter_amount=0.1
---steps_to_train=256
+--filter_amount=0.1
```
delock: Does filter_amount affect training result?   I saw it was 1.0 in Christian's repo, and it was 0.05 in v0.5 repo.   Not sure what it is doing.

amj: The AlphaGo Zero paper says it trains on moves sampled from the most recent 500,000 games; each training checkpoint is produced every 1,000 steps at batch size 2,048 = 2M examples.  500,000 games at average game length 200 = 100M examples.  So the original alphago zero paper suggests only using about 2% of the total examples!  But they have a lot more of them...

It's unclear what the correct value is for the much smaller mlperf scenario, so i don't know what the right thing to do is.  It seems like seeing every example about once is reasonable, which would imply setting filter_amount and window_size such that `filter_amount * window_size == 1`

delock: I see.  Does it mean if my learning rate works for filter_amount=1.0, and it is now changed to 0.1, I need to push learning rate to 10x to compenstate that I didn't see that sample 10 times?

amj: Probably not; most likely the value-head of the network is strongly overfit anyway due to the small number of games played.  To check, if the `value_cost` component of the training error is < 0.15, the value head is probably overfit.  The delta between the holdout and the train error is a good proxy for whether or not it's overfit.

I don't really know, though!  It's certainly worth trying.

ml_perf/flags/9/architecture.flags
```diff
@@ -1,6 +1,7 @@
 # Flags that control the model architecture.
 
---conv_width=64
+--conv_width=32
 --fc_width=64
 --trunk_layers=9
 --value_cost_weight=0.25
```
delock: Why value_cost_weight is set to 0.25?  I saw it is 1.0 by default, as in Christian's repo

amj: This also prevents the value head from overfitting.
Basically, for 9x9 the policy head has 50x-100x more data points than the value head.  For 19x19, the policy head has ~200-250x more data.

When training a supervised network on relatively small numbers of games, DeepMind was "weighting the mean-squared error component by a factor of 0.01", i.e., `value_cost_weight=0.01`

My hunch is that as filter_amount approaches 1, value_cost_weight should approach `1 / (average_game_length)` with some adjustment for window size.

ml_perf/flags/9/bootstrap.flags
amj: no, this could be turned down again, esp if we're going to use normal distribution.

amj:
Closed as obsolete from #781 .

PR #762, opened 2019-03-11
upload 000057-fury.pb as evaluation target

delock:
Propose to use 000057-fury.pb as evaluation target.   Its win rate over leela in Christian's repo 51%.
Black win rate: 40%
White win rate: 62%


tommadams:
Thanks! I've uploaded the `000057-fury.pb` to GCS as `gs://tensor-go-ml-perf/models/9x9/target.pb`

PR #767, opened 2019-03-13
index and meta files for 000057-fury

delock:
These files are trained using tensorflow commit bba3f6a4e733dfd5865dfb14943624c13a7ba9fd
6b06740240d1d424c192fc00eb1fedb1  000057-fury.data-00000-of-00001         
e669002db0978d75eb7a29009dc44322  000057-fury.index                       
9f7db618ad3bc1a3058cbc0a52144db2  000057-fury.meta                        
0d60af29fed1e92a130cc6d66d1f9edd  000057-fury.pb


tommadams:
Thanks! I've copied the `.data-00000-of-00001`, `.index` and `.meta` files to `gs://tensor-go-ml-perf/models/9x9/target.*` and replaced the `.pb` file with one compiled from TF 1.11, since this is the version specified in the [Minigo README](https://github.com/tensorflow/minigo/#install-tensorflow).

We will upgrade Minigo to TF 1.13 some time after the MLPerf 0.6 code freeze.

PR #778, opened 2019-03-18
fix eval_models to use new "models" format, use new "wait" coroutine

amj:


amj:
@tommadams fyi

tommadams:
Thanks for fixing my mess!

PR #780, opened 2019-03-18
Update pip requirements.

w-hat:
Running `pip3 install -r requirements.txt` gave the following errors:

pyasn1-modules 0.2.4 has requirement pyasn1<0.5.0,>=0.4.1, but you'll have pyasn1 0.1.9 which is incompatible.
google-api-core 1.8.1 has requirement setuptools>=34.0.0, but you'll have setuptools 20.7.0 which is incompatible.


amj:
Thanks for the patch!

PR #781, opened 2019-03-18
update all flags with final values

amj:
depends on the normal-distribution noise PR.


amj:
These should be the final flags needed for MLPerf.

Two changes should be tested to optionally speed up the benchmark by > 50%.  Both will be at a cost of convergence stability:

1. Cut the games played per iteration back to 2048
2. Decrease the inferences per move ("readouts") back to 240 or 200.

Both of these decrease the total compute required by a large amount, at the cost of decreasing the total amount of data (option 1) or decreasing the data's overall quality (option 2).

PR #788, opened 2019-03-26
move legal_moves from MctsNode to Position

tommadams:
This is a small step towards removing all go-specific game logic out of `MctsNode`.

`legal_moves` is now updated in `Position::PlayMove` rather than the `MctsNode` ctor.

I've also removed the `n` move number argument from the `Position` ctor. It was only ever used for a single test case.

All unit tests pass and this change plays the same game as before. I've also successfully played 1000 random games without a crash.


cc/mcts_node.cc
```diff
@@ -28,54 +28,37 @@ namespace minigo {
...
-      }
-
-      case Position::MoveType::kNoCapture: {
-        // The move will not capture any stones: we can calculate the new
-        // position's stone hash directly.
-        auto new_hash = position_hash ^ zobrist::MoveHash(c, to_play);
-        node->legal_moves[c] = !node->HasPositionBeenPlayedBefore(new_hash);
-        break;
-      }
+// Superko implementation that uses MctsNode::superko_cache.
```
amj: this is a weird noun for a classname; it's not clear to me what a superko as a noun is but w/e.  I don't have any better ideas, really; "ZobristHistory" ? "SuperkoDetector?"  Nah.

tommadams: `ZobristHistory` sgtm

PR #789, opened 2019-03-27
Update flags for better convergence

delock:
This set of hyper parameter make convergence better than the original by enlarge learning rate.
Changes:
1. Initial learning rate is 16x larger and l2_strength is set to 1e-4
2. Virtual loss is set to 8, just in case for reproducible, 16 should still work.
3. Target exit criteria is set to >=0.50, to make target easier to reach.

### DIFF
--- a/ml_perf/flags/9/train.flags
+++ b/ml_perf/flags/9/train.flags
@@ -7,8 +7,9 @@
 
 # Device specific hyperparameters re: batch size and LR schedules.
 --train_batch_size=4096
---lr_rates=0.01
---lr_rates=0.002
+--lr_rates=0.16
+--lr_rates=0.016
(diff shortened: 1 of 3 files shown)


ml_perf/flags/9/train.flags
```diff
@@ -7,8 +7,9 @@
 
 # Device specific hyperparameters re: batch size and LR schedules.
 --train_batch_size=4096
---lr_rates=0.01
---lr_rates=0.002
+--lr_rates=0.16
+--lr_rates=0.016
+--lr_rates=0.0016
```
amj: can you post your tensorboard logs to examine?  I'd love to see the effect of these rate cuts on the metrics.

PR #790, opened 2019-03-27
add checkpoint and golden chunks that a model start to learn to win

delock:
This change allows training start from a checkpoint, than a random initialized model.  The start point is chosen that the model is close to learn to have non-zero win rate against target.  This ensure steady raise of target win rate and remove part of the variance of iteration to train a model to beat target model.

Note start.pb is generated by Tensorflow 1.13.   Conversion from meta file will be needed if TF1.10 needs to be used.


tommadams:
Thanks Guokai. I'm going to try this PR now.

Can you move the checkpoint files into a directory named "checkpoint" please?

ml_perf/reference_implementation.py
```diff
@@ -414,7 +423,12 @@ def rl_loop():
                                 '*')
 
     # Calculate the window size from which we'll select training chunks.
-    window = 1 + state.iter_num
+    if not FLAGS.check_point:
+      window = 1 + state.iter_num
+    else:
+      window = (FLAGS.slow_window_size
```
tommadams: I don't really understand this logic. To compute the window size correctly, you have to take into account the number of training chunks that were in the checkpoint.

It might be easiest to move the logic that computes the window size directly into `get_golden_chunk_records` and calculate it from the number of files matched by the `*.zz` glob instead of `state.iter_num`.

delock: This logic is to make sure window size is max_windows_size when training started, and yes, this logic needs improve.
I'll modify the code to be more flexible according to suggestion, and think over this part.

amj: the "slow window" flag was only needed when we were starting from random noise; its purpose is to get the initial random games out of the window sooner, because we had thought that the initial way out of the randomness was right about when the window would fill up.  That's not actually the case, apparently, so we can actually get rid of the whole slow-window flag in the reference implementation.

ml_perf/reference_implementation.py
```diff
@@ -469,6 +489,23 @@ def main(unused_argv):
   # Copy the target model to the models directory so we can find it easily.
   shutil.copy('ml_perf/target.pb', fsdb.models_dir())
 
+  if FLAGS.check_point:
+    # Copy the start model to the models directory.
+    shutil.copy('ml_perf/start.pb', fsdb.models_dir())
```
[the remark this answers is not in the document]
delock: > Thanks Guokai. I'm going to try this PR now.
> 
> Can you move the checkpoint files into a directory named "checkpoint" please?

All the checkpoint file been moved to checkpoint directory.

ml_perf/reference_implementation.py
delock: The logic is improved.  The new logic also support chunk number between 1 and max_window_size, so a premature checkpoint can be replicated and window size would grow to max_window_size.

I think the original train from scratch mode should be kept as optional in reference implementation, so people can actually train from scratch and know the implementation is following the spirit of reinforcement learning.

tommadams:
Thanks a lot for the changes! I tested this PR some more today and discovered a few problems:

- the code crashes with `--parallel_post_train=true`.
- the first model trained after loading the checkpoint performs badly: it looses almost all its eval games against `start.pb`. We believe the initial problem was that `train.py` wasn't actually loading the checkpoint model to resume training from. I've fixed that but the training still isn't initializing properly. We're still looking into the cause.
- the checkpoint `.tfrecord.zz` files are much bigger than I originally thought, so we don't want to add them to our github repo. I've copied them up to GCS and have script to copy the files down and put them in the correct location.
- we're going to remove the slow window code completely.

Andrew and I are still working on these changes and I hope to have a PR ready tomorrow. There's no action needed on your part.

tommadams:
See #793 for the PR that I've been working on.

delock:
The initial model is the last model with zero win rate in a successful training.  The low win rate against start.pb is kind of strange.  I see the same in three runs I have locally.

There might be some reasons:
1. the checkpoint model is picked from run with low learning rate (0.01).  I don't see there is a problem, however.
2. checkpoint mismatch between model and golden chunks?  But this had been checked, the original filename is below:

**start.pb:**
6aa55c166d59fba91444bfa47a6ee355  000040-000030.pb  
**golden_chunks:**
10238a5c2d837da05646fabfac3b59ac  data/golden_chunks/000031-000024.tfrecord.zz     
b0a567b0bff9f6d6438448e56375eda2  data/golden_chunks/000032-000025.tfrecord.zz     
3a7aafbb044168d3a4863079290adb94  data/golden_chunks/000033-000025.tfrecord.zz     
4e0ff8e36c5445741fa71fb4eff1b68b  data/golden_chunks/000034-000026.tfrecord.zz     
70633be203a9a202d185dbbebf486d40  data/golden_chunks/000035-000026.tfrecord.zz     
9384f92c61ed2ea1501d1d575fc2bb7a  data/golden_chunks/000036-000026.tfrecord.zz     
6b1877b827ba10aca4314bf857cbd67e  data/golden_chunks/000037-000027.tfrecord.zz     
e589665de7cbd747fff6fc74cdffd708  data/golden_chunks/000038-000028.tfrecord.zz     
cd1b62d22cb353e03398ecf63be698a0  data/golden_chunks/000039-000028.tfrecord.zz     
e7f9056dbe159be7cb9a933ff0ad99c3  data/golden_chunks/000040-000029.tfrecord.zz     



> Thanks a lot for the changes! I tested this PR some more today and discovered a few problems:
> 
> * the code crashes with `--parallel_post_train=true`.
> * the first model trained after loading the checkpoint performs badly: it looses almost all its eval games against `start.pb`. We believe the initial problem was that `train.py` wasn't actually loading the checkpoint model to resume training from. I've fixed that but the training still isn't initializing properly. We're still looking into the cause.
> * the checkpoint `.tfrecord.zz` files are much bigger than I originally thought, so we don't want to add them to our github repo. I've copied them up to GCS and have script to copy the files down and put them in the correct location.
> * we're going to remove the slow window code completely.
> 
> Andrew and I are still working on these changes and I hope to have a PR ready tomorrow. There's no action needed on your part.

tommadams:
Closed in favour of #793

PR #791, opened 2019-03-27
add script to repeatedly run the ml_perf reference implementation and sync results

amj:


ml_perf/repeat_run.sh
```diff
@@ -0,0 +1,38 @@
+# Copyright 2018 Google LLC
```
tommadams: 2019

ml_perf/repeat_run.sh
```diff
@@ -0,0 +1,38 @@
...
+  BOARD_SIZE=9 python ml_perf/reference_implementation.py \
+      --base_dir=$BASE_DIR \
+      --flagfile=ml_perf/flags/9/rl_loop.flags \
+      --parallel_post_train
+
+  BOARD_SIZE=9 python ml_perf/eval_models.py \
+      --base_dir=$BASE_DIR \
+      --flags_dir=ml_perf/flags/9/
+
+  gsutil -m rsync -r -x "data" $BASE_DIR gs://$1/$(basename $BASE_DIR)
```
tommadams: s/$1/$BUCKET_NAME/

ml_perf/repeat_run.sh
```diff
@@ -0,0 +1,38 @@
...
+
+set -e
+
+: ${1?"Need to call with 'repeat_run.sh BUCKET_NAME' (no gs:// prefix needed)"}
+
+BUCKET_NAME=$1
+
+while true
+do
+  BASE_DIR=$(pwd)/results/$(hostname)-$(date +%Y-%m-%d-%H)
```
tommadams: It's not actually guaranteed that BASE_DIR will be unique: if someone sets the number of iterations to 1 for some reason, we'll finish multiple loops withing a single hour.

I'm not sure how much we actually care about that, but consider including a unique id to the BASE_DIR.

amj: added minutes.

ml_perf/repeat_run.sh
amj: Done.

ml_perf/repeat_run.sh
amj: Done.

PR #792, opened 2019-03-28
change inference cache to use a zobrist hash for a key

tommadams:
Use a Zobrist hash consisting of the following as the inference cache key:

- the stones on the board.
- who is to play.
- which moves are legal (due to ko, superko).
- whether the previous move was a pass.

This should cover the majority of situations where the model would might return different results for the same basic position.

Note that the inference cache still isn't actually hooked up yet :P


cc/zobrist.h
```diff
@@ -28,19 +28,31 @@ namespace zobrist {
...
+inline Hash OpponentPassedHash() { return internal::kOpponentPassedHash; }
+
+// Hashes for moves by black and white.
 inline Hash MoveHash(Coord c, Color color) {
-  return kMoveHashes[c][static_cast<int>(color)];
+  return internal::kMoveHashes[c][static_cast<int>(color)];
 }
 
-inline Hash KoHash(Coord c) { return c == Coord::kInvalid ? 0 : kKoHashes[c]; }
+// Hashes used for empty points that can't be played because of (e.g.) ko or
```
sethtroisi: nit: (e.g.) is weird in this sentence.

also self-capture...

tommadams: yeah, not sure what that (e.g.) was all about

PR #794, opened 2019-03-29
introduce multi-instance support to minigo cc

delock:
This code change add an instance_id parameter to selfplay.cc and eval.cc, to make sure multiple instance running on same host would generate unique play data filename.

### DIFF
--- a/cc/eval.cc
+++ b/cc/eval.cc
@@ -68,6 +68,7 @@ DEFINE_string(model, "",
               "engine=lite, the model should be .tflite flatbuffer.");
 DEFINE_string(model_two, "", "Descriptor for the second model");
 DEFINE_int32(parallel_games, 32, "Number of games to play in parallel.");
+DEFINE_int32(instance_id, 0, "Unique id with multi-instance.");
 
 // Output flags.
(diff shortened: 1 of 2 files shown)


tommadams:
/ok-to-test

Thanks for sending the C++ changes by themselves.

It just occurred to me that we wouldn't actually need to add these changes if the name computed by [GetOutputName](https://github.com/tensorflow/minigo/blob/5d8ab6bee3c45a66b217676486ee59c9d7d4cdc1/cc/game_utils.cc#L65) contained the process ID.

Would that work for your use case?

delock:
Yes, process ID in the output name would be sufficient to make unique name.

> /ok-to-test
> 
> Thanks for sending the C++ changes by themselves.
> 
> It just occurred to me that we wouldn't actually need to add these changes if the name computed by [GetOutputName](https://github.com/tensorflow/minigo/blob/5d8ab6bee3c45a66b217676486ee59c9d7d4cdc1/cc/game_utils.cc#L65) contained the process ID.
> 
> Would that work for your use case?

tommadams:
#795 removes the need for this PR

PR #795, opened 2019-03-29
Add process ID to GetOutputName

tommadams:
This is useful because some MLPerf participants want to run multiple selfplay processes on the same host, so adding the process ID avoids the possibility that multiple processes will attempt to write to the same file.

Sadly, the underlying type for a process ID differs between platforms, so this is a little messy. Fair warning: I haven't even attempted to compile the windows code, but it _should_ be close to correct :)


cc/platform/utils.h
```diff
@@ -19,24 +19,39 @@
 
 #if defined(_MSC_VER)
 
+#define WIN32_LEAN_AND_MEAN
```
sethtroisi: lol

cc/platform/utils_osx.cc
```diff
@@ -32,6 +29,10 @@ int GetNumLogicalCpus() {
   return nproc;
 }
 
+bool FdSupportsAnsiColors(int fd) { return isatty(fd); }
+
+ProcessId GetProcessId() { return getpid(); }
+
```
sethtroisi: for two lines below (GetHostname) should you be initializing hostname = {}? if the if condition isn't triggered isn't it possible that hostname doesn't contain a null character and std::string() reads past end of hostname?

cc/game_utils.cc
```diff
@@ -63,7 +63,7 @@ std::string FormatWinStatsTable(
 
 std::string GetOutputName(absl::Time now, size_t game_id) {
   return absl::StrCat(absl::ToUnixSeconds(now), "-", GetHostname(), "-",
-                      game_id);
+                      GetProcessId(), "-", game_id);
```
sethtroisi: I'm fairly sure I have regexes that depend on this. I should reconsider my life choices.

tommadams: sorry :(

cc/platform/utils_osx.cc
tommadams: I've cleaned that code up. I think it was ok because if the `gethostname` call failed, the code would copy the string literal `"hostname"` into the buffer before converting it to a string.

You made me reread the `gethostname` docs for windows and posix because I was worried about corner cases, which of course there are. Fortunately, it seems like using a 256B buffer avoids them :)

cc/game_utils.cc
amj: now you have two regexes!  Happy day.

delock:
@tommadams 
There seems not enough randomness when running in multi-instance mode this way, thread_id, instead of process id are used to set random seed for each thread.  But all instance will have same set of thread_id, and make game out come not random enough.

```
if (player_options.random_seed != 0) {
  player_options.random_seed += 1299283 * thread_id;
}
```

tommadams:
Can you pass a different `random_seed` to each instance? I presume you have already made local changes to that part of the `reference_implementation.py` to launch multiple subprocesses.

delock:
Yes, different random seed should work.  Thanks.
> Can you pass a different `random_seed` to each instance? I presume you have already made local changes to that part of the `reference_implementation.py` to launch multiple subprocesses.

PR #799, opened 2019-04-01
Swish

sethtroisi:
Based on https://arxiv.org/pdf/1710.05941.pdf

This needs to be rebased after #798


tommadams:
From the paper: "For training Swish networks, we found that slightly lowering the learning rate used to train ReLU networks works well."

It would have been nice if they provided actual numbers here :/

amj:
LGTM

sethtroisi:
@tommadams Per personal communication slight is 20-30% on the initial learning rate (which is followed by a 10x cut at the first decay)

PR #801, opened 2019-04-01
hook up inference cache & add thread-safe implementation

tommadams:
This still isn't ready for use yet, so it's disabled by default. After hooking the basics up, I've finally convinced myself that the code should be refactored before progressing any further.

Note that setting `--tree_reuse=false` will break Minigui for now. So don't do that.


cc/mcts_player.cc
```diff
@@ -232,18 +232,40 @@ void MctsPlayer::TreeSearch() {
...
-  for (int i = 0; i < max_iterations; ++i) {
+  int num_cache_misses = 0;
+  int num_cache_hits = 0;
+  while (num_cache_misses < max_cache_misses) {
     auto* leaf = root->SelectLeaf();
     if (leaf->game_over() || leaf->at_move_limit()) {
       float value =
           leaf->position.CalculateScore(game_->options().komi) > 0 ? 1 : -1;
       leaf->IncorporateEndGameResult(value, root);
+      ++num_cache_misses;
```
sethtroisi: this appears to be double counted

sethtroisi: also shouldn't there be a continue here?

cc/mcts_player.cc
```diff
@@ -232,18 +232,40 @@ void MctsPlayer::TreeSearch() {
...
                               std::vector<MctsPlayer::TreePath>* paths) {
-  int max_iterations = num_leaves * 2;
+  DualNet::Output cached_output;
+
+  int max_cache_misses = num_leaves * 2;
   int num_selected = 0;
-  for (int i = 0; i < max_iterations; ++i) {
+  int num_cache_misses = 0;
+  int num_cache_hits = 0;
+  while (num_cache_misses < max_cache_misses) {
```
sethtroisi: I don't see how this condition can ever trigger

cc/mcts_player.cc
tommadams: The rest of the loop is in the else branch so it won't get executed. I've restructured things a bit to make this clearer.

(this bit of code is one of the things I would like to make clearer)

cc/mcts_player.cc
tommadams: It triggers near the end of the game where SelectLeaf would otherwise repeatedly call `IncorporateEndGameResult` because it keeps hitting `game_over` or `at_move_limit` leaves.

cc/gtp.cc
```diff
@@ -66,7 +68,7 @@ DEFINE_double(decay_factor, 0.98,
 DEFINE_string(model, "",
               "Path to a minigo model. The format of the model depends on the "
               "inference engine.");
-DEFINE_int32(cache_size_mb, 512, "Size of the inference cache in MB.");
+DEFINE_int32(cache_size_mb, 0, "Size of the inference cache in MB.");
```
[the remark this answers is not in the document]
tommadams: done

PR #802, opened 2019-04-01
clean up ML_LOG(FATAL) and MG_CHECK

### LINKED ISSUE #717: MG_LOG(FATAL) doesn't print a callstack

tommadams:
Fixes #717


cc/logging.cc
```diff
@@ -24,15 +24,40 @@
...
+  if (level == LogLevel::INFO) {
+    // We don't add a prefix to MG_LOG(INFO) log lines because many things rely
+    // on the exact string being printed (GTP, correct formatting of position &
+    // node descriptions, etc).
+    return;
+  }
+
   char c;
   switch (level) {
     case LogLevel::INFO:
```
sethtroisi: should this case be deleted?

cc/logging.cc
```diff
@@ -24,15 +24,40 @@
...
 namespace internal {
+
 namespace {
+
 absl::Mutex* mutex() {
   static auto* m = new absl::Mutex();
   return m;
 }
+
+void DumpStackTrace(std::ostream* os) {
```
sethtroisi: potentially stupid question: can you pass std::ostream& os and avoid (*os)?

tommadams: I could, but we try and follow the google style guide, which only allows for const reference parameter.

cc/logging.cc
tommadams: good catch, thanks!

PR #805, opened 2019-04-03
Required version of Bazel is now v0.19.2

tommadams:
Tensorflow 1.13 claims that Bazel 0.19.0 is the required version but that predictably doesn't actually work.


sethtroisi:
:+1:

PR #806, opened 2019-04-04
no longer derive GtpPlayer and MiniguiPlayer from MctsPlayer

tommadams:
Renames `GtpPlayer` and `MiniguiPlayer` to `GtpClient` and `MiniguiGtpClient`, both of which now own a `MctsPlayer` instead being inherited from one, which is cleaner and will make other refactoring easier.

Most of these changes are a natural consequence of an "is-a" relationship changing to a "has-a" but less obvious changes include:

- Courtesy pass handling moved from `GtpPlayer::SuggestMove` to `GtpClient::HandleGenmove`.
- The old `MiniguiPlayer` had a `ProcessLeaves` override that would stream the live tree search updates over to Minigui. Since we're no longer deriving from `MctsPlayer` that isn't possible, so I added a tree search callback instead, which `MinigoGtpClient` sets on the `MctsPlayer` in its constructor.
- I removed the `verbosity` and `info` GTP commands because they aren't needed any more.


cc/gtp_player.cc
```diff
@@ -338,21 +331,21 @@ GtpPlayer::Response GtpPlayer::HandleKnownCommand(CmdArgs args) {
...
-GtpPlayer::Response GtpPlayer::HandleKomi(CmdArgs args) {
+GtpClient::Response GtpClient::HandleKomi(CmdArgs args) {
   auto response = CheckArgsExact(1, args);
   if (!response.ok) {
     return response;
   }
 
   double x;
-  if (!absl::SimpleAtod(args[0], &x) || x != game()->options().komi) {
+  if (!absl::SimpleAtod(args[0], &x) || x != player_->game()->options().komi) {
```
sethtroisi: this probably works but is it guaranteed that

<DOUBLELITERAL> == SimpleAtod("<DOUBLELITERAL>")

e.g. 
1.3f == SimpleAtod("1.3")

cc/mcts_player.h
```diff
@@ -119,64 +129,59 @@ class MctsPlayer {
 
   virtual void NewGame();
 
+  // If visitor is non-null, it is called on each batch of leaves visited during
```
sethtroisi: I don't understand this comment.

cc/gtp_player.cc
tommadams: "normal" komi values of X.5 are exactly representable in IEEE floating point format, so this should be ok for all cases we care about.

cc/mcts_player.h
tommadams: That's because it doesn't make any sense :)

It's a comment left over from an earlier attempt at this refactor where I was passing a visitor object around to `SuggestMove`, `ProcessLeaves`, etc instead of having a `SetTreeSearchCallback` method. That was a mess.

PR #809, opened 2019-04-05
Cleanup2

tommadams:
Built on top of PR #806: please wait for that to be pushed before reviewing.

`GtpClient::HandleLoadSgf` and `MiniguiGtpClient::HandleLoadSgf` contained a bunch of duplicated code to load & parse the SGF file.

This PR makes `GtpClient::HandleLoadSgf` take care of loading & parsing the SGF file and defers actually replaying the game to a new `ReplaySgf` method that has different implementations for `GtpClient` and `MiniguiGtpClient` (the former just replays the main line, the latter replays all variations so that the UI knows about them).


cc/gtp_player.h
```diff
@@ -126,14 +126,20 @@ class GtpClient {
...
   // Returns true if we pondered.
   bool MaybePonder();
 
   virtual void Ponder();
 
-  // Begin pondering again if requested.
-  void MaybeStartPondering();
+  // Replay a loaded SGF game.
+  // Called by HandleLoadSgf after the SGF file has been loaded and parsed, and
+  // a new has been started.
```
sethtroisi: "a new has been started." doesn't make a lot of sense

PR #811, opened 2019-04-05
Add Coral EdgeTPU compatibility. 

mtyka:
Adapt minigo python engine to use EdgeTPU by adding a custom dual_net.py

This supports both the EdgeTPU dev board and the USB form factor for
Linux and Raspberry Pi hosts. In order not to use dependencies that
aren't present or available on all plattforms the imports are inside
the if statement. Potentially this could be moved to it's own
dual_net_chooser.py or something.

Also added some convenience scripts in minigui/edgetpu

### DIFF
--- a/dual_net_edge_tpu.py
+++ b/dual_net_edge_tpu.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
+# Copyright 2019 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


dual_net_edge_tpu.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2018 Google LLC
```
tommadams: 2019 :)

dual_net_edge_tpu.py
```diff
@@ -0,0 +1,61 @@
...
+It can only be used for inference and requires a specially quantized and
+compiled model file. For more information see https://coral.withgoogle.com
+"""
+
+import numpy as np
+import features as features_lib
+from edgetpu.basic.basic_engine import BasicEngine
+
+class DualNetworkEdgeTpu(object):
+  def __init__(self, save_file, board_size=19):
```
tommadams: Please use the `go.N` global variable for the board size, which is controlled by the `BOARD_SIZE` environment variable (and defaults to 19):
https://github.com/tensorflow/minigo/blob/6694df26917ca54173878ff1eeaa97c5e40d9265/go.py#L31

tommadams:
Looks like there are a few lint errors too

iopq:
What kind of performance could I expect out of it? I don't own one yet, but if this gets merged I'd be interested in some numbers

tommadams:
I've only trained a small test model so far so I can't give you any accurate numbers. I'm going to train a bigger model next week and we'll post performance numbers when we have them.

tommadams:
Awesome, thanks!

PR #815, opened 2019-04-08
Move verbose logging out of MctsPlayer and into caller

tommadams:
I'm moving Minigui to a model where it doesn't rely on tree reuse. Instead, it will do the standard thing of resetting the board and playing out the move sequence whenever you select a new position in the UI.

This allows for finer-grained control over exactly what gets logged on a per-caller basis. For example, we'd like to print the tree search table in Minigui but not the `Played >>$MOVE` debug because that would now generate a lot of spam as we play out a sequence of moves to reach the requested state.


tommadams:
thanks for the quick review!

PR #818, opened 2019-04-08
Prework for cbt_ratings.py

sethtroisi:


oneoffs/eval_sgf_to_cbt.py
```diff
@@ -89,7 +95,8 @@ def read_existing_paths(bt_table):
     rows = bt_table.read_rows(
         filter_=row_filters.ColumnRangeFilter(
             METADATA, SGF_FILENAME, SGF_FILENAME))
-    names = (row.cell_value(METADATA, SGF_FILENAME).decode() for row in rows)
+    reader = tqdm(rows, desc="eval_game", unit=" rows")
```
sethtroisi: This read takes ~15 seconds so it's nice to have some status

oneoffs/eval_sgf_to_cbt.py
```diff
@@ -103,7 +110,7 @@ def canonical_name(sgf_name):
 
     # Often eval is inside a folder with the run name.
     # include from folder before /eval/ if part of path.
-    with_folder = re.search(r'/([^/]*/eval/.*)', sgf_name)
+    with_folder = re.search(r'(?:^|/)([^/]*/eval/.*)', sgf_name)
```
sethtroisi: if you don't provide a full path this wouldn't always work.

?: is to avoid (^|/) being a capture group (and having to renumber group(1) below)

ratings/ratings.py
```diff
@@ -110,10 +108,6 @@ def import_files(files, bucket=None):
 
             m_num_w = re.match(MODEL_REGEX, pw).group(1)
             m_num_b = re.match(MODEL_REGEX, pb).group(1)
-            # v10 and v9 have the same model and name for 588, so cross eval
-            # games played with this model should be ignored.
-            if m_num_w == '000588' or m_num_b == '000588':
```
sethtroisi: there are at least 3 other broken instances, verifying that #816 is a no-op was stupidly hard.

ratings/schema.sql
```diff
@@ -29,16 +29,18 @@ CREATE TABLE IF NOT EXISTS models (
...
-  unique(bucket, model_name)
+
+  UNIQUE(bucket, model_name)
 );
 
 CREATE TABLE IF NOT EXISTS wins (
     game_id integer primary key,
     model_winner integer not null,
     model_loser integer not null,
-    foreign key(game_id) references games(game_id),
```
sethtroisi: references seems pointless without ON DELETE CASCADE.

also sqlite doesn't enforce these by default which is annoying

ratings/schema.sql
```diff
@@ -52,13 +54,16 @@ CREATE TABLE IF NOT EXISTS games (
...
+  FOREIGN KEY(w_id) REFERENCES models(id) ON DELETE CASCADE
 );
 
 
 CREATE INDEX IF NOT EXISTS model_name_bucket_index ON models (model_name, bucket);
 CREATE INDEX IF NOT EXISTS game_model_b_index ON games (b_id);
 CREATE INDEX IF NOT EXISTS game_model_w_index ON games (w_id);
-CREATE INDEX IF NOT EXISTS game_filename on games (filename);
+CREATE INDEX IF NOT EXISTS game_filename_index on games (filename);
+CREATE INDEX IF NOT EXISTS win_winner_index on wins (model_winner);
```
sethtroisi: This makes #817 fast

PR #825, opened 2019-04-16
Export training brackets from Bigtable to GCS

gitosaurus:
Useful when wanting a repeatable series of training sets independent
of the normal RL loop.

As part of this:

  - Flags to control size and sampling of training window
  - Provide random-access version of the two-queue mix
  - Use multiprocessing to speed up batch export
  - Remove individual progress logs when done
    (they're only interesting when incomplete)


bigtable_input.py
```diff
@@ -599,6 +608,33 @@ def set_fresh_watermark(game_queue, count_from, window_size,
...
+    assert 0 <= deciles <= 10
+    # The prefixes and suffixes below have the following meanings:
+    #   ct_: count
+    #   fr_: fraction
+    #    _r: resign (ordinary)
+    #   _nr: no-resign
+    lesser = 10 - math.floor(deciles)
+    greater = 10 - lesser
+    ct_r, ct_nr = greater, lesser
+    ct_total = ct_r + ct_nr
```
tommadams: I don't think ct_total is really needed, since it's always 10

batch_exporter.py
```diff
@@ -0,0 +1,157 @@
...
+flags.DEFINE_integer('training_moves', 2**21,
+                     'Number of moves to select from training window')
+
+flags.DEFINE_float('training_fresh', 0.05,
+                   'Fraction of fresh games in each new training window')
+
+flags.DEFINE_integer('batch_size', 1024,
+                     'How many TFRecords to pull through tf.Session at a time')
+
+flags.DEFINE_string('output_prefix', 'gs://dtj-minigo-us-central1/tryit_',
```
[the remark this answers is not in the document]
gitosaurus: Definitely! And I'll error out if no value is provided.

bigtable_input.py
gitosaurus: Good point. I'll rearrange it so that `ct_total` is defined up top as 10 and used like a local constant.

PR #846, opened 2019-07-01
Build tensorflow as a dependency of the Minigo WORKSPACE

tommadams:
This PR is based on [chsigg's work](https://github.com/chsigg/minigo-fork/tree/tf1.12) but I removed the option to depend on TensorFlow directly (instead of a prebuilt version), which simplifies things immensely.

Since I was pruning a bunch of deps from the WORKSPACE, I also took the opportunity to delete cc/position_benchmark.cc and remove the benchmark dep.

It works as follows:

- As before, users must first run `//cc/configure_tensorflow.sh` which runs the TensorFlow configuration script and then builds the `//cc/tensorflow:build` rule. You'll note that the script is a _lot_ simpler now.
- The dependencies of the `//cc/tensorflow:build` rule build all the headers, libraries & binaries that we need from TensorFlow.
- The `copy_outputs.sh` script that `//cc/tensorflow:build` subsequently runs copies these generated files to the correct location under `//cc/tensorflow/`.


cc/position_benchmark.cc
```diff
@@ -1,84 +0,0 @@
-// Copyright 2018 Google LLC
```
sethtroisi: Was deleting this intentional?

tommadams: Yeah, I pruned a bunch of deps from our WORKSPACE and this was the only benchmark we have. I haven't run it in a year, so there didn't seem much point in keeping it around.

tommadams:
It looks like I'll need up update the integration tests...

How do I do that?

tommadams:
This PR removes the files `tensorflow.bazelrc` and `tf_configure.bazelrc` because they're auto-generated by `cc/configure_tensorflow.sh` and their contents can differ from machine to machine. For example, my desktop and the continuous integration docker image (that keeps failing) have CUDA installed in different locations.

Unfortunately, deleting these `.bazelrc` files as part of the PR causes the integration tests to fail because although the base Docker image contains the correctly configured `.bazelrc` files, pulling this PR into the image (of course) deletes those files from the repo.

It doesn't appear to be possible to get this PR to pass automated integration testing, so I'm just going to push it. The theory is that when testing later PRs, the `git pull` will no longer delete the required `.bazelrc` files. We shall see.

I've manually verified that the C++ tests pass from the `cc-image` Docker image.

Issue #845, opened 2019-07-01
minigui doesn't support handicap stones


amj:

amj:
closed via #848

PR #848, opened 2019-07-03
add handicap support to minigui

amj:
Quick recap:  as we walk the AST, we transform it into Sgf::Node with our own Move structs, etc.
This CL changes "GetTreeImpl" to look for a HA property, and if so, it adds Nodes for black to play handicap stones and white to pass.

The resulting tree of nodes is handled as before:  the engine calls PlayMove on each.


brilee:
see warnings in docstring:
https://github.com/brilee/MuGo/blob/master/sgf_wrapper.py#L5

amj:
@brilee right now, our sgf parser literally ignores any SGF property on a node that isn't B[] or W[].  the HA property requires that another property be set on the node ('AB') specifying the actual location of the handicap stones.

So yeah your warning in the docstring is understood :)  This PR doesn't add support for positions with existing stones, it just adjusts the semantics of our translation from "SGF Tree" => "Minigo Tree"

amj:
fixes #845

cc/sgf.cc
```diff
@@ -219,6 +220,44 @@ bool GetTreeImpl(const Ast::Tree& tree,
...
+      for (const auto& h_location : stones->values) {
+        Move m;
+        m.color = Color::kBlack;
+        m.c = Coord::FromSgf(h_location, true);
+        if (m.c == Coord::kInvalid) {
+          MG_LOG(ERROR) << "Can't parse node " << node.ToString() << ": \""
+                        << h_location << "\" isn't a valid SGF coord for a handicap stone";
+          return false;
+        }
+        std::string comment;
```
tommadams: remove

cc/sgf.cc
```diff
@@ -219,6 +220,44 @@ bool GetTreeImpl(const Ast::Tree& tree,
...
+        dst->push_back(absl::make_unique<Node>(m, std::move(comment)));
```
tommadams: s/std::move(comment)/""/

cc/sgf.cc
```diff
@@ -219,6 +220,44 @@ bool GetTreeImpl(const Ast::Tree& tree,
...
+        dst = &(dst->back()->children);
+        if (h_location != stones->values.back()) {
+          m.color = Color::kWhite;
+          m.c = Coord::kPass;
+          dst->push_back(absl::make_unique<Node>(m, std::move(comment)));
```
tommadams: same here (this is actually undefined behaviour because you need to reinitialized the moved-from `comment` before using it again. Clang would have probably warned about this)

amj: Done.

cc/sgf.cc
amj: Done.

cc/sgf.cc
amj: Done.

Issue #849, opened 2019-07-03
show the variation move numbers perhaps as a toggle in "clean" mode


amj:
When you're partway down a variation tree, it can be nice to see the numbering from where it diverged from the main line.

Perhaps similar to the 'clean' mode that hides the variations, it'd be a 3 way toggle to show just the variation we're exploring.

("variation" here means the one visible on the game tree widget displayed in minigui)

PR #852, opened 2019-07-18
Show numbers on moves branched from main line

amj:
All this seems to work ok.  It's a little weird -- the annotation layer always copies in its annotations from the underlying Position object, so i ended up making these annotations at the time the new Positions are created and just toggling their visibility.  It's not a great solution but it seems to work ok.  Happy to think about refactors that would make this make more sense.

Switching the last-move indicator to unicode seems to work and also to be aligned correctly :)

The fix on board.ts seems like a new one because of type information added to the context object?  I don't think i changed the original behavior but maybe i did?


amj:
fixes #849

amj:
oh hmm, this actually won't work really well if there's a ko in the variation.  Maybe don't review it just yet XD

amj:
Ok, i added some duplication check to the annotation code in position.ts.  The annotation layer should probably be keeping a map keyed by position instead of a flat array, but maybe that's for next time.  Ready for review here

minigui/board.ts
```diff
@@ -284,7 +284,7 @@ class ClickableBoard extends Board {
   drawImpl() {
     super.drawImpl();
     let p = this.enabled ? this.p : null;
-    this.ctx.canvas.style.cursor = p ? 'pointer' : null;
+    this.ctx.canvas.style.cursor = p ? 'pointer' : 'none';
```
brilee: javascript already has enough null types :angry:

amj: aha, but you see, that is not javascript!  It's CSS!

(sobs quietly)

brilee: oh god no.

So you're saying that before it was replacing `undefined` with `null` and now it's replacing `undefined` with `"none"`.

:sob:

amj: ya.  Typescript (correctly) noted that this css attribute could not be null.  I think what has happened is the type annotations were extended to these objects with the latest es2019 or whatever.  But i dunno.  Having more type checking seems good!

tommadams: *sigh* `'none'` is not actually correct since it will hide the cursor. The intent with assigning `null` was to remove the property value, thereby setting it back to the default.

Will typescript allow you to assign the empty string `''`?

minigui/position.ts
```diff
@@ -125,6 +119,34 @@ class Position {
...
+    if (! p.isMainLine) {
+      let result: Move[] = [];
+      let node: Nullable<Position>
+      for (node = p; node && node.isMainLine != true; node = node.parent) {
+        if (node.lastMove == null) { break; }
+        result.push(node.lastMove);
+      }
+      result.reverse();
+      let playedCount = new Uint16Array(N * N);
+      for (let i=0; i < result.length -1; ++i) {
```
tommadams: nit: missing space `result.length - 1`

minigui/position.ts
```diff
@@ -125,6 +119,34 @@ class Position {
     // Create a new child.
     p.isMainLine = this.isMainLine && this.children.length == 0;
     p.parent = this;
+
+    // If it is not on the main line, prepare the move number annotations.
+    if (! p.isMainLine) {
+      let result: Move[] = [];
+      let node: Nullable<Position>
+      for (node = p; node && node.isMainLine != true; node = node.parent) {
```
tommadams: nit: `s/node.isMainLine != true/!node.isMainLine/`

minigui/position.ts
amj: Done.

minigui/position.ts
amj: Done.

minigui/board.ts
amj: Yes, it does, and that works better.  I made the assumption that the intent was to hide the cursor, so that's my bad.  Fixed.

minigui/layer.ts
```diff
@@ -458,7 +458,20 @@ namespace Variation {
 
 
 class Annotations extends Layer {
-  private annotations = new Map<Annotation.Shape, Annotation[]>();
+  private annotations = new Map<string, Annotation[]>();
```
tommadams: as mentioned below, `annotations` can just be an `Annotation[]`, no need for a map

minigui/layer.ts
```diff
@@ -475,10 +488,12 @@ class Annotations extends Layer {
     let position = this.board.position;
     this.annotations.clear();
     for (let annotation of position.annotations) {
-      let byShape = this.annotations.get(annotation.shape);
+      // this isn't really working with the enums, but its not clear how the
+      // annotations on a position should be grouped, otherwise.
+      let byShape = this.annotations.get(annotation.label);
```
tommadams: We don't actually need this grouping any more if we're just rendering text. Previously, the idea was that we'd group all annotations by type (circle, triangle, etc) first, then render them in batches.

I _think_ you can remove all this and just do `this.annotations = position.annotations`

minigui/layer.ts
```diff
@@ -491,21 +506,24 @@ class Annotations extends Layer {
...
-            ctx.fill();
-          }
-          break;
+    let textHeight = Math.floor(0.8 * sr);
+    ctx.font = `${textHeight}px sans-serif`;
+    ctx.textAlign = 'center';
+    ctx.textBaseline = 'middle';
+
+    // TODO this.annotation should be keyed by the board position, oops.
+    this.annotations.forEach((annotations: Annotation[], label: string) => {
```
tommadams: You can then remove this outer loop

minigui/layer.ts
amj: Done.

minigui/layer.ts
amj: Done.

minigui/layer.ts
amj: Done.

PR #855, opened 2019-07-31
fix 'statement has no effect' compiler warnings in logging macros

tommadams:


cc/logging.h
```diff
@@ -70,6 +72,26 @@ class CheckFailStream {
   LogStream impl_;
 };
 
+// CheckNoOpStream is used in non-debug builds (i.e. when NDEBUG defined) that
+// allows the compiler to completely optimize away MG_DCHECK macros.
```
amj: any compiler settings required for this optimization?

tommadams: Clang 7.0 optimizes `MG_DCHECK` away with `-O1` and `-DNDEBUG`.
gcc 7.1 optimizes it away even with `-O0` and `-DNDEBUG`.

https://godbolt.org/z/tq3mr5

tommadams:
`pull-tf-minigo-cc` passes \o/

PR #856, opened 2019-08-02
run clang-format over C++ code

tommadams:


tommadams:
cool, the `clang-format` presubmit is working better now

PR #859, opened 2019-08-02
Port NVIDIA's TRT inference engine from MLperf 0.6

tommadams:
Original code: https://github.com/mlperf/training_results_v0.6/blob/master/NVIDIA/benchmarks/minigo/implementations/tensorflow/minigo/

The TRT inference engine is faster for 9x9 models & small batch sizes, but slower for the v17 model & larger batch sizes (as measured on my Quadro M2000). My GPU doesn't have native fp16 support, so I haven't been able to compare performance there.


tommadams:
Freezing:

```BOARD_SIZE=19  python  freeze_graph.py  --trunk_layers=19  --conv_width=256  --fc_width=256  --model_path=saved_models/001003-leviathan  --use_tpu=false  --use_SE=true  --use_SE_bias=true  --use_trt=true  --trt_max_batch_size=8```

`trt_max_batch_size` specifies the 
Selfplay:
```bazel-bin/cc/selfplay   --parallel_games=1   --num_games=1   --num_readouts=400  --soft_pick=false --inject_noise=false   --model=trt:32,saved_models/001003-leviathan.trt.pb  --virtual_losses=8  --seed=42  --random_symmetry=false```

tommadams:
Finally working!

I should have mentioned in my earlier comment that you can pass `--trt_precision=fp16` to `freeze_graph.py`

amj:
LGTM

PR #861, opened 2019-08-07
Playout cap oscillation

amj:
Tests pass, cc:all builds, but it's pretty painful.

I did end up making inject_noise default to false.


cc/game.h
```diff
@@ -95,6 +97,9 @@ class Game {
 
   void UndoMove();
 
+  // Marks the most recent move as trainable.
+  void SaveLastMove();
```
tommadams: `SaveLastMove` could mean a bunch of different things. How about something along the lines of `MarkLastMoveAsTrainable`?

amj: Done.

cc/mcts_player.h
```diff
@@ -92,6 +92,23 @@ class MctsPlayer {
     // move is played.
     bool tree_reuse = true;
 
+    // "Playout Cap Oscillation" as per the KataGo paper.
+    // If fastplay_frequency > 0, tree search is modified as follows:
```
brilee: does it make sense to express this in terms of 'slowplay frequency'? I find numbers like 0.02 more semantically meaningful than 0.98, but it's a very asethetic concern and consistency with the katago paper is probably more important.

cc/selfplay.cc
```diff
@@ -94,6 +94,16 @@ DEFINE_string(flags_path, "",
               "Note that flags_path is different from gflags flagfile, which "
               "is only parsed once on startup.");
 
+DEFINE_double(fastplay_frequency, 0.0,
+              "The fraction of moves that should use a lower number of "
+              "playouts, aka 'playout cap oscillation'.\nIf this is set, "
+              "'fastplay_readouts' should also be set.");
+
+DEFINE_int32(fastplay_readouts, 20,
```
brilee: is there a pattern that avoids potentially having conflicting defaults in the flag parser and the options constructor?

cc/selfplay.cc
```diff
@@ -375,16 +391,22 @@ class SelfPlayer {
                        << player->options().virtual_losses << ")";
           MG_LOG(INFO) << root->CalculateTreeStats().ToString();
 
-          MG_LOG(INFO) << root->position.ToPrettyString(use_ansi_colors);
-          MG_LOG(INFO) << "Move: " << position.n()
-                       << " Captures X: " << position.num_captures()[0]
-                       << " O: " << position.num_captures()[1];
-          MG_LOG(INFO) << root->Describe();
+          if (!fastplay) {
```
brilee: the prevalence of `!fastplay` suggests to me that slowplay is the right way to think about it?

oneoffs/joseki.html
```diff
@@ -31,7 +31,7 @@ <h1> Minigo Opening Analysis </h1>
             </v-flex>
           </v-layout>
 
-          <v-layout v-for="opening in top10" row>
+          <v-layout v-for="opening in top10" row :key="opening.sequence">
```
brilee: intentionally in this commit?

cc/mcts_player.h
amj: in that case i should probably reverse these, because katago specifies the slowplay % XD

cc/selfplay.cc
amj: agreed.

oneoffs/joseki.html
amj: Not really, but it fixes up a warning.

cc/selfplay.cc
amj: While that makes sense, "slowplay" is essentially the defaults.  So these flags are specifying the other behavior.

cc/selfplay.cc
amj: Not that i know of.

cc/mcts_player.h
amj: actually thinking about it further, all the existing doc assumes "slowplay" mode only.  Rather than rename everything, the 'fastplay' is the new behavior.

PR #862, opened 2019-08-07
reference implementation of Benson's Algorithm

tommadams:
A super-slow implementation of Benson's Algorithm. I'll replace it with something more efficient that uses fixed size arrays in my a follow-up PR. I've left a bunch of debug logging in there for now.

After working on this, I feel a compelling need to replace "group" with "chain" throughout the C++ code


cc/pass_alive_test.cc
```diff
@@ -0,0 +1,732 @@
...
+          ....OOO..
+          ......OOO
+          ......O..
+          ......OOO
+          ......OXX
+          ......OX.
+          ......OXX
+          ......OX.)",
+       // expected result
+       R"(....OOOOO
```
sethtroisi: Why is the corner pass alive?

If we make it 2x5 it should not be pass alive

cc/pass_alive_test.cc
```diff
@@ -0,0 +1,732 @@
...
+          .OOOO.X.X
+          OO.....XX
+          .........
+          .........
+          X......OO
+          .XX...O.O
+          X.X..OO.O
+          .X.X.O.O.)",
+       // expected result
+       R"(..O.O.XXX
```
sethtroisi: Why is upper left not pass alive? As a human it is

amj: Top-left point is not a liberty of the white group, which means that white region is "large".  (there exist "human pass-alive" groups that do not meet Benson's definition of pass-alive)

amj: Oh hm, i'm not quite correct.  IIUC, the top left region has a non-adjacent point AND the group has only one vital region.  The difference between this and the one you commented on above (line 160) is that the group on line 160 has two vital regions already.   Hope i'm right this time :)

sethtroisi: Is there a polynomial time algorithm that returns the definitive status of groups?

cc/pass_alive_test.cc
amj: not sure, but i think this vs. the example on line 214 is relevant

cc/pass_alive_test.cc
tommadams: Black can capture by surrounding the white group then finally playing at `a`:

```
.XOaOX.
XOOOOX.
OOXXX..
XX.....
.......
```

tommadams: Benson's paper mentions an N! algorithm if that's any help :P

cc/pass_alive_test.cc
tommadams: Yeah, the two tests on lines 194 & 214 give more examples of this behaviour.

cc/pass_alive_test.cc
```diff
@@ -0,0 +1,732 @@
...
+       R"(.........
+          .........
+          ...XX....
+          ..X..X...
+          ..X..X...
+          ...XX....
+          .........
+          .........
+          .........)",
+       // expected result
```
brilee: the entire board isn't just considered dead?

cc/pass_alive_test.cc
```diff
@@ -0,0 +1,732 @@
...
+  RunTests(tests);
+}
+
+TEST_F(PassAliveTest, 19x19) {
+  if (kN != 19) {
+    return;
+  }
+
+  TestCase tests[] = {
+      {// board state
```
brilee: since the parser ignores whitespace, can you add a space in between every character of this test case, to get a square aspect ratio? :D

cc/position.cc
```diff
@@ -424,6 +429,264 @@ float Position::CalculateScore(float komi) {
...
+        regions[c] = r.get();
+        if (stones_[c].empty()) {
+          r->empty_points.insert(c);
+        } else {
+          r->other_color_points.insert(c);
+        }
+
+        for (auto nc : kNeighborCoords[c]) {
+          auto ns = stones_[nc];
+          if (ns.color() != color) {
```
brilee: random question: why is .color() a computed property rather than just dereferencing a struct field? Is it doing a lookup in the parent board object or something like that?

cc/position.cc
```diff
@@ -424,6 +429,264 @@ float Position::CalculateScore(float komi) {
...
+          } else {
+            r->enclosing_groups.insert(groups[nc]);
+          }
+        }
+      }
+
+      // Find the vital groups for this region.
+      for (auto* g : r->enclosing_groups) {
+        bool is_vital =
+            std::includes(g->liberties.begin(), g->liberties.end(),
```
brilee: what a clumsy api for subset calculation... why doesn't this  just take g->liberties and r->empty_points as args? does it assume that the set members are sorted for O(N+M) runtime?

cc/position.cc
```diff
@@ -424,6 +429,264 @@ float Position::CalculateScore(float komi) {
...
+  std::vector<std::unique_ptr<BensonGroup>> group_pool;
+  std::array<BensonRegion*, kN * kN> regions;
+  std::array<BensonGroup*, kN * kN> groups;
+
+  // Build the set of all groups.
+  board_visitor_->Begin();
+  for (int row = 0; row < kN; ++row) {
+    for (int col = 0; col < kN; ++col) {
+      Coord c(row, col);
+      if (stones_[c].color() != color || !board_visitor_->Visit(c)) {
```
brilee: trying to understand this visitor pattern. The outer double-for loop guarantees that every square is visited at least once. But is every square visited exactly once? Otherwise I feel like double-counting would occur when compiling regions/neighboring groups.

cc/position.cc
```diff
@@ -424,6 +429,264 @@ float Position::CalculateScore(float komi) {
...
+  // Run Benson's algorithm.
+  for (;;) {
+    // List of groups removed this iteration.
+    std::vector<std::unique_ptr<BensonGroup>> removed_groups;
+
+    // Iterate over remaining groups.
+    for (size_t i = 0; i < group_pool.size();) {
+      if (group_pool[i]->num_vital_regions < 2) {
+        // This group has fewer than two vital regions, remove it.
+        removed_groups.push_back(std::move(group_pool[i]));
```
brilee: I guess I don't understand what std::move is doing here or why you have to manually fix up the back side of the group_pool vector. Why aren't these three lines just `removed_groups.push_back(group_pool.pop_back())`?

cc/pass_alive_test.cc
tommadams: "Not pass-alive" would be more accurate than "dead".

The original black stones are in the expected result because of how the test is set up: it overlays the result from `CalculatePassAliveRegions` onto the input board position. In this case, `CalculatePassAliveRegions` returns an empty result for both black & white, so the expected result is just the input result.

I think we may be saying the same thing...

cc/position.cc
tommadams: A `Stone` is a bit-packed `uint16` that contains the colour in the bottom 2 bits and an ID of the group that the stone belongs to in the upper 14 bits.

cc/pass_alive_test.cc
tommadams: "Not pass-alive" would be more accurate than "dead".

The original black stones are in the expected result because of how the test is set up: it overlays the result from `CalculatePassAliveRegions` onto the input board position. In this case, `CalculatePassAliveRegions` returns an empty result for both black & white, so the expected result is just the input result.

I think we may be saying the same thing...

cc/position.cc
tommadams: The `BoardVisitor` guarantees that each stone is only visited once. The `Visit` method returns `true` the first time a point is visited and `false` otherwise. `Visit` pushes the `Coord` of the point onto an internal queue, which `Next` pops from.

Here's a simple example of using just the `BoardVisitor` to visit every point on the board:

```C++
   BoardVisitor bv;
   bv.Begin()
   bv.Visit(Coord::FromGtp("A1"));
   while (!bv.Done()) {
     Coord coord = bv.Next();
     std::cout << "Visited " << coord << "\n";
     for (auto neighbor_coord : GetNeighbors(coord)) {
       bv.Visit(neighbor_coord);
     }
   }
```

cc/position.cc
tommadams: We're removing element `i`, not the last element.

Since we don't care about the order of elements in `group_pool`, we can remove element `i` without having to shuffle all the later elements over by swapping `i` and the last element, then popping the last element (which used to be `i`) off.

cc/position.cc
tommadams: All the C++ library algorithms APIs are like this, I have no idea why they never introduced some kind of `range` abstraction :(

`std::includes` does assume both ranges are sorted, which `std::set` guarantees (it's often implemented as a red-black tree).

cc/pass_alive_test.cc
tommadams: *sigh*

PR #866, opened 2019-08-13
make the buffer count part of the model spec for TpuDualNet

tommadams:
So we can tune the buffering without having to recompile.


cc/dual_net/tpu_dual_net.h
```diff
@@ -34,7 +34,7 @@ class TpuDualNet : public DualNet {
   // tpu_name: The gRPC address of the VM's TPU, e.g. "grpc://10.240.2.10:8470".
   // graph_path: A frozen GraphDef proto suitable for running on TPU, e.g.
   //             generated by freeze_graph.py --use_tpu
```
amj: need to add buffer_count_

PR #868, opened 2019-08-14
reduce the size of the nlohmann json library download

tommadams:
The old zip we were downloading was 114MB, which is crazy.
Just download the `json.hpp` that we need.


tommadams:
Oof. I need to update the base image because I changed `WORKSPACE`.

PR #869, opened 2019-08-15
Improved performance instrumentation

tommadams:
- Add [WTF](https://google.github.io/tracing-framework/) tracing.
- Report approximate per-inference time spent in both `BatchingDualNet::RunMany` and TensorFlow's `Session::Run` instead of the old "Milliseconds per 100 inferences", which didn't take into account the number of parallel games.

Since this changes `WORKSPACE`, I'm going to have to rebuild the `cc-base` image again.


cc/dual_net/batching_dual_net.cc
```diff
@@ -17,15 +17,19 @@
...
 namespace minigo {
 
 namespace internal {
 
 ModelBatcher::ModelBatcher(std::unique_ptr<DualNet> model_impl,
-                           size_t buffering)
-    : model_impl_(std::move(model_impl)), buffering_(buffering) {}
+                           size_t buffer_count)
+    : model_impl_(std::move(model_impl)),
+      buffer_count_(buffer_count),
```
sethtroisi: Can you add a note about what buffer_count is used for (and how reasonable values are 0-4?)

tommadams: added a comment to the header

PR #874, opened 2019-08-23
Adds a new Model API

tommadams:
This PR replaces `DualNet` as the API of choice with `Model`. 

`Model::RunMany` differs from `DualNet::RunMany` in that it takes arrays of `Model::Input`s instead of `DualNet::BoardFeatures`, where a `Model::Input` consists of the stones on the board and whose turn it is. The subclass is then responsible for applying symmetries & converting the `Model::Input` into its own preferred set of features. This has a few advantages:

- Simplifies the calling code (it no longer has to worry about symmetries).
- Gives subclasses more implementation freedom. For example:
  - We can now have `LiteDualNet` generate `uint8` input features directly instead of taking `float` features and converting them. I haven't done that in this PR because it's already got way out of hand.
  - Potentially apply symmetries in-place while generating them from the board positions, thus avoiding a copy.
- We don't have DualNet everywhere in the code when we add a third output for the predicted ownership :P

Other things:

`DualNet` is now a subclass of `Model` and takes care of conversion from input board positions to features.

I rewrote how model buffering worked. Previously, `TfDualNet` and `TpuDualNet` both managed their own threads of workers. Now, if buffering is required, the model factory just creates multiple instances and wraps them in a `BufferedModel` which is very simple. This is a lot cleaner and means that `Model` implementations now no longer have to be thread-safe.


brilee:
Wait, you're not going to call it TripleNet?

tommadams:
Oh, I also deleted `TrtDualNet` because it's no longer necessary and I was sick of working on this PR.

It was a port from NVIDIA's MLPerf 0.6 implementation and only differed from `TfDualNet` in that it didn't perform double buffering. This can now be achieved by changing `TfDualNetFactory` to return a single `TfDualNet` instance instead of two per GPU, which is left as an exercise for the reader.

Instead of double buffering their `DualNet` for MLPerf, NVIDIA instead opted to run two selfplay processes per GPU.

PR #880, opened 2019-08-30
Choose inference symmetry deterministically in the MctsPlayer

tommadams:
This is in preparation for an upcoming change to make the inference cache symmetry-aware.

Having the `MctsPlayer` choose the symmetry to apply when running inference means that it will be able check if that symmetry exists in the cache.

This PR was built out of a branch that contains both this change and the symmetry-aware inference cache code and so contains a few other bits and pieces.

Most of these changes are removing the `random_symmetry` and `random_seed` parameters from the `DualNet` related classes.


cc/dual_net/dual_net_test.cc
```diff
@@ -148,17 +148,17 @@ TEST(DualNetTest, TestBackendsEqual) {
...
-  tests.emplace("LiteDualNet",
-                Test(absl::make_unique<LiteDualNetFactory>(false, 0),
-                     "test_model.tflite"));
+  tests.emplace("LiteDualNet", Test(absl::make_unique<LiteDualNetFactory>(),
+                                    "test_model.tflite"));
 #endif
 
   Random rnd(Random::kUniqueSeed, Random::kUniqueStream);
   Model::Input input;
+  input.sym = symmetry::kIdentity;
```
amj: this is a nice way to have it.

cc/mcts_node.cc
```diff
@@ -65,6 +65,38 @@ MctsNode::MctsNode(MctsNode* parent, Coord move)
...
       position(parent->position) {
+  // TODO(tommadams): move this code into the MctsPlayer and only perform it
+  // only if we are using an inference cache.
+  if (parent->HasFlag(Flag::kHasCanonicalSymmetry)) {
+    SetFlag(Flag::kHasCanonicalSymmetry);
+    canonical_symmetry = parent->canonical_symmetry;
+  } else {
+    // TODO(tommadams): skip this check if `move` is kPass or on the diagonal.
+    static_assert(symmetry::kIdentity == 0, "kIdentity must be 0");
+    auto best_symmetry = symmetry::kIdentity;
```
amj: a comment for what this is doing is nice.  "best" & "worst" should have their value system elucidated.

cc/mcts_player.cc
```diff
@@ -123,9 +123,7 @@ Coord MctsPlayer::SuggestMove(int new_readouts, bool inject_noise) {
   // time SuggestMove has been called for a game, or PlayMove was called without
   // a prior call to SuggestMove.
   if (!root_->HasFlag(MctsNode::Flag::kExpanded)) {
-    tree_search_leaves_.clear();
-    SelectLeaves(1, &tree_search_leaves_);
-    ProcessLeaves(tree_search_leaves_);
+    TreeSearch(1);
```
amj: Does this not work with options_.virtual_losses?  It should... I'm only vaguely concerned about fixed-batch size models, although i suppose the model impl can take care of padding?

cc/mcts_player.cc
```diff
@@ -190,29 +188,14 @@ Coord MctsPlayer::PickMove() {
...
-  // history size.
-  int n = kMaxPositionHistory;
-  for (int j = 0; j < n; ++j) {
-    input->position_history.push_back(&node->position.stones());
-    node = node->parent;
-    if (node == nullptr) {
-      break;
-    }
-  }
+// TODO(tommadams): break back this into SelectLeaves and ProcessLeaves.
```
amj: nit: it like looks a word

cc/mcts_player.h
```diff
@@ -194,7 +194,22 @@ class MctsPlayer {
     int last_move = 0;
   };
 
-  void GetModelInput(const MctsNode* node, Model::Input* input) const;
+  symmetry::Symmetry GetCanonicalSymmetry(const MctsNode* node) const {
```
amj: commenting the difference between canonical/inference is useful.

cc/mcts_player_test.cc
```diff
@@ -448,89 +440,6 @@ TEST_F(MctsPlayerTest, ExtractDataResignEnd) {
   EXPECT_EQ("W+R", game_->result_string());
 }
 
-// Fake DualNet implementation used to verify that MctsPlayer symmetries work
```
amj: are there new tests?

tommadams: coming up in the next PR

cc/mcts_node.cc
tommadams: done.

cc/mcts_player.h
tommadams: done.

PR #882, opened 2019-09-03
Symmetry-aware inference cache

tommadams:
This PR makes the inference cache symmetry-aware:
 - inference output is merged into the cache, tagged with the symmetry used for that inference.
 - inference output for different symmetries of the same position are averaged together, giving a more accurate result for a cache hit.
 - when querying the cache for an inference output, the desired symmetry for the position must have already been merged into the cache for it to return a hit.


cc/selfplay.cc
```diff
@@ -135,6 +136,12 @@ DEFINE_int32(parallel_games, 32, "Number of games to play in parallel.");
 DEFINE_int32(num_games, 0,
              "Total number of games to play. Defaults to parallel_games. "
              "Only one of num_games and run_forever must be set.");
+DEFINE_int32(cache_size_mb, 0, "Size of the inference cache in MB.");
```
amj: if the cache is only on selfplay, it will be hard to ensure that this isn't affecting the playing strength.  Should this flag be moved into the mcts_player?

tommadams: I'm trying to keep flags out of the library build targets. This leads to some copy-paste in the selfplay & eval binaries but it's worth it IMO: https://abseil.io/tips/45

Also, we need to share the inference cache between player instances, so it needs to be created at the top level.

tommadams:
> LGTM. I'll run some vs mode tests to see if this is a strength gain. out of curiosity, have you measured the speed improvement vs. no cache vs. regular cache? Even a SWAG would be nice.

Profiling multiple parallel games is problematic because the execution times are so noisy so I timed playing a single 9x9 game with and without cache, which gave reliable timings. When using the same random seed, the game plays the same for the first ~20 moves before diverging because of the effect of the cache merging symmetric inferences.

Without cache:
  game length: 110 moves
  execution time: 33.5239 seconds
  moves/second: 3.281

With cache (30.228% hit rate):
  game length: 83
  execution time: 19.5835
  moves/second: 4.239

The cache has a 30% hit rate and plays moves 1.29x faster

PR #883, opened 2019-09-09
basic tensorboard launcher script

amj:
should make life easier.


tommadams:
It's funny we both wrote the same thing at the same time :)
Mine's in PR #885 and supports GCS. Your code's cleaner though. Which one do you want to go with boss?

amj:
well it was such a bite-sized task, i felt i could manage it... let's use your version though, as the GCS functionality is too good.

PR #889, opened 2019-09-19
Add UndoMove support

tommadams:
For use by a ladder detector.


tommadams:
Nope. It's the callers responsibility to handle that kind of thing, because `PlayMove` takes a `ZobristHistory` object rather than the position owning all that state. In practice (for now at least) this code is only going to be used for ladder features and I don't think we're going to bother performing triple-ko hash checks for those.

amj:
LGTM

PR #890, opened 2019-09-20
Target pruning & avoid playing a move in bensons-regions.

amj:
Not yet ready for review.

### DIFF
--- a/cc/mcts_node.cc
+++ b/cc/mcts_node.cc
@@ -128,6 +128,7 @@ MctsNode::MctsNode(MctsNode* parent, Coord move)
 Coord MctsNode::GetMostVisitedMove(bool restrict_in_bensons) const {
   // Find the set of moves with the largest N.
   inline_vector<Coord, kNumMoves> moves;
+  // CalculatePassAlive does not include the kPass point.
   std::array<Color, kN * kN> out_of_bounds;
 
   if (restrict_in_bensons) {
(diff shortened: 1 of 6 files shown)


cc/mcts_node.cc
```diff
@@ -125,11 +125,24 @@ MctsNode::MctsNode(MctsNode* parent, Coord move)
...
+    for (auto& x : out_of_bounds) {
+      x = Color::kEmpty;
+    }
+  }
+
   int best_N = -1;
   for (int i = 0; i < kNumMoves; ++i) {
+    if (out_of_bounds[i] != Color::kEmpty) {
+      continue;
+    }
```
tommadams: I'm concerned that `GetMostVisitedMove(true)` could plausibly select a move with 0 visits, which might cause us to `CHECK` fail later. I can also imagine no moves being picked at all and the code `MG_CHECK(!moves.empty())` below failing.

amj: It's even worse/better than that, as pruning things can leave you with weird target distributions, causing training to go into NaN-land.

I'll see if i can come up with some breaking examples and get some better tests.  Shelving for now.

tommadams: oof :(

amj:
ok!  @tommadams @sethtroisi  This should be ready to review now.  The changes include:

ReshapeFinalVisits (target pruning) now happens in MctsPlayer::PlayMove, right before the move is added to the game and it's search_pi is finalized.  This means SuggestMove is no longer destructive.

GetMostVisitedMove now optionally returns the best move not in a benson's region.  If there aren't any, it returns pass.  if pass isn't expanded, MctsPlayer does the right thing on the next time through the loop by treating it identical to the case in a new game.

ReshapeFinalVisits reshapes relative to the action score of the most visited move, regardless of bensons'.  However, it also zeros out visits to bensons regions, meaning that it's possible the move played ends with no visits.

There's tests for the node manipulations, no changes to the mcts_player tests.

In a run, the models with target_pruning on came out much stronger than ones with equivalent playouts, 274:26

image
(Red = target pruning.  Orange = without.  Grey = without and twice as many reads)

PR #893, opened 2019-09-22
[EvalServer] AddModel.py to add models to models_for_eval

sethtroisi:
I'd like a review before I add the new cloud bigtable ('model_for_eval').

Then I'll test a couple of times before I send for final review


cluster/eval_server/add_model.py
```diff
@@ -0,0 +1,155 @@
...
+
+flags.mark_flags_as_required([
+    "cbt_project", "cbt_instance", "cbt_table"
+])
+
+FLAGS = flags.FLAGS
+
+
+# TODO(sethtroisi): Unify with oneoffs
+MODEL_ROW = "m_{}_{:0>10}"
```
amj: nit: give example in comment.

cluster/eval_server/add_model.py
```diff
@@ -0,0 +1,155 @@
...
+    assert "-num_readouts=" in flags, "must set playouts"
+    assert "-model" not in flags, "leave model unset"
+
+    # Verify binary_path is to a gtp binary and it exists
+    binary_base = binary_path.replace(GTP_BIN, "")
+    assert binary_path.endswith(GTP_BIN), binary_path
+    assert os.path.isfile(binary_path)
+    # $ ldd bazel-bin/cc/gtp
+    #    libtensorflow_framework.so => (bazel path) => cc/tensorflow/lib/libtensorflow_framework.so
+	#    libtensorflow_cc.so => (bazel path) => cc/tensorflow/lib/libtensorflow_cc.so
```
amj: still needed?

cluster/eval_server/add_model.py
```diff
@@ -0,0 +1,155 @@
...
+    assert gfile.Exists(src)
+    assert not gfile.Exists(dst)
+
+    with gfile.GFile(src, "rb") as src_f, gfile.GFile(dst, "wb") as dst_f:
+        shutil.copyfileobj(src_f, dst_f)
+
+
+def add_model(name, model_path, flags, binary_path):
+    """Add a player (model + flags) to models cbt
+    name: name of the model
```
amj: what is the name here, actually?  e.g. "000990-cormorant", or v17-990, or some other scheme?  This is a good place to take a no-op comment ("name: the name") and give some clarity.

cluster/eval_server/add_model.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Upload model and binary
+    gcs_player_folder = os.path.join(PLAYER_FOLDER, player_hash)
+    new_model_path = os.path.join(gcs_player_folder, os.path.basename(model_path))
+    new_binary_base = os.path.join(gcs_player_folder, "bin")
+    binary_base = binary_path.replace(GTP_BIN, "")
+
+    print (f"Saving to {gcs_player_folder!r}")
+
+    #copy_to_gcs(model_path, new_model_path)
```
amj: The print statement above is a lie?

(i'm assuming this is pending some clever way to dedupe the tf libs so as not to reup the same ones over and over?)

cluster/eval_server/add_model.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # validate that all rows written successfully
+    for status in enumerate(response):
+        if status.code is not 0:
+            print("Failed to write {}".format(i, status))
+
+
+if __name__ == '__main__':
+    remaining_argv = flags.FLAGS(sys.argv, known_only=True)
+    fire.Fire({
```
amj: absl.app?  I'd like to avoid new uses of fire if possible.

cluster/eval_server/add_model.py
sethtroisi: I think this illustratives why I'm copying these two files.

cluster/eval_server/add_model.py
sethtroisi: it's a free form name field. I'll add an assert it conforms to [a-Z0-9-._]

amj: ok.  I'm saying this is the chance to describe how the caller should use it.  I.e. how you think the names should be used or what makes a good name or an example...

cluster/eval_server/add_model.py
sethtroisi: This get's automatically generated as m_eval_{HASH} see line ~104

cluster/eval_server/add_model.py
sethtroisi: Let's discuss this offline but now passing args is harder :/

cluster/eval_server/add_model.py
sethtroisi: fixed.

cluster/eval_server/add_model.py
sethtroisi: Fixed.

PR #897, opened 2019-09-23
add player, result, a little delay to dump_game

amj:


oneoffs/dump_game.py
```diff
@@ -56,6 +70,10 @@ def main(argv):
         to_play = 'B' if x.position.to_play == 1 else 'W'
         print('{}>> {}: {}\n'.format(
             x.position, to_play, coords.to_gtp(x.next_move)))
+        time.sleep(0.1)
```
tommadams: lets make the sleep amount a flag

amj: Done.

PR #901, opened 2019-09-25
Add support for liberty features

tommadams:
This is also a large step towards supporting arbitrary input features. It also has the advantage of writing the features directly into the model's input tensors. I've only added support to TfDualNet for now, TpuDualNet and LiteDualNet will have to wait in the interests of getting ready for the next round of MLperf ASAP (they're not compiled by default anyway).

I ran the MLperf benchmark a couple of times and it finished almost 2x faster with liberty features.


brilee:
You should make the liberty planes 1, 2, and 3+ rather than 1, 2, 3

It's important to have a quick test for "am I really big?"

tommadams:
I should have probably mentioned that I haven't updated the tests yet ;)

tommadams:
> You should make the liberty planes 1, 2, and 3+ rather than 1, 2, 3 It's important to have a quick test for "am I really big?"

The current choice of liberty features is taken from KataGo paper: https://github.com/lightvector/KataGo/blob/e90b29d798c0229eee7ac2c982819b09a490fa69/cpp/neuralnet/nninputs.cpp#L870

I agree we should test how {1, 2, 3+} compares to {1, 2, 3} but I'd prefer to try those tweaks after Andrew's happy with the 9x9 MLPerf hyperparameters.

tommadams:
Tests fixed, PTAL.

`LiteDualNet` is working for AlphaGo Zero features, haven't updated `TpuDualNet` yet.

cc/model/model.h
```diff
@@ -28,33 +28,84 @@ namespace minigo {
...
+
+   private:
+    Tensor<T> tensor_;
+    std::vector<T> buffer_;
+  };
+
+  struct Input {
     // Symmetry to apply to the input features when performing inference.
-    symmetry::Symmetry sym;
+    symmetry::Symmetry sym = symmetry::kNumSymmetries;
```
amj: isn't this out of bounds?  should be kIdentity?

cc/position.cc
```diff
@@ -99,7 +99,8 @@ Position::UndoState Position::PlayMove(Coord c, Color color,
   return undo;
 }
 
-void Position::UndoMove(const UndoState& undo) {
+void Position::UndoMove(const UndoState& undo,
+                        ZobristHistory* zobrist_history) {
```
amj: doesn't UpdateLegalMoves need to use this?  or is this from a diff PR?

cc/dual_net/dual_net_test.cc
```diff
@@ -37,107 +37,173 @@
...
-  for (int i = 0; i < DualNet::kNumStoneFeatures; ++i) {
-    result[i] = features[c * DualNet::kNumStoneFeatures + i];
+struct AgzFeatures {
+  static constexpr Model::FeatureType kFeatureType = Model::FeatureType::kAgz;
+  static constexpr int kNumFeaturePlanes = 17;
+};
+
+struct ExtraFeatures {
+  static constexpr Model::FeatureType kFeatureType = Model::FeatureType::kExtra;
+  static constexpr int kNumFeaturePlanes = 20;
```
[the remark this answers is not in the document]
tommadams: Yes. This was for the sake of expedience during the transition to the variadic template implementation.

cc/dual_net/tf_dual_net.cc
```diff
@@ -108,46 +106,44 @@ TfDualNet::~TfDualNet() {
...
+                      outputs_[1].flat<float>().data());
+  DualNet::GetOutputs(inputs, policy, value, outputs);
 
   if (model_name != nullptr) {
     *model_name = graph_path_;
   }
 }
 
-void TfDualNet::Reserve(size_t capacity) {
+void TfDualNet::Reserve(int capacity) {
```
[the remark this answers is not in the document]
tommadams: To remove a few static_casts

cc/model/model.h
tommadams: Yes. That's intentional so that we'll immediately check fail in `ApplySymmetry` if it's not set (rather than silently using `kIdentity`: https://github.com/tensorflow/minigo/blob/9fc09ee0ad5a061eda9c19ace5b163caf8513d5f/cc/symmetries.h#L195

cc/position.cc
tommadams: I'm not sure what you mean. The code passes `zobrist_history` to `UpdateLegalMoves` at the end of this function

PR #902, opened 2019-09-27
Add test for pass-losing behavior

amj:
These tests are notable because they work with our "pass-hack" but fail without it.

### DIFF
--- a/cc/mcts_player_test.cc
+++ b/cc/mcts_player_test.cc
@@ -54,6 +54,18 @@ static constexpr char kTtFtwBoard[] = R"(
     X.XXXXXXX
     XXXXXXXXX)";
 
+static constexpr char kOneStoneBoard[] = R"(
+    .........
+    .........
+    .........
+    .........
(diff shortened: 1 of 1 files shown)


cc/mcts_player_test.cc
```diff
@@ -290,6 +302,48 @@ TEST_F(MctsPlayerTest, ParallelTreeSearch) {
   EXPECT_EQ(0, CountPendingVirtualLosses(root));
 }
 
+TEST_F(MctsPlayerTest, DontPassOnEmptyLosingBoard) {
+  MctsPlayer::Options options;
+  auto player = CreateBasicPlayer(options);
+  auto* root = player->root();
+  auto board = TestablePosition(kOneStoneBoard, Color::kWhite);
```
sethtroisi: Can you add comment. // Search board with one black stone, white to play.

amj: Done.

PR #903, opened 2019-09-28
Variadic features

tommadams:
This is built on top of #901 so don't bother to look at this too closely until that PR is merged. If you're interested in looking at this now, I suggest limiting yourself to the code in `cc/model/features.h` and `cc/model/features_test.cc`. If you want to peek behind the curtain, most of the template magic is in `cc/model/features_impl.h`.

This PR allows input features to be easily customizable. The key takeaway from this PR is that a model's features are defined as:

```C++
using AgzFeatures = Features<StoneFeatures, ToPlayFeature>;
using ExtraFeatures = Features<StoneFeatures, ToPlayFeature, LibertyFeatures>;
```

This similar to the Python code, only better because the code all gets folded together at compile time :)

The the `Features` class generates a batch of input features from a list of model inputs like so:

```C++
AgzFeatures::Set<float>(input_batch, &feature_tensor);
ExtraFeatures::Set<uint8_t>(input_batch, &feature_tensor);
```

The generated code that sets the features is efficient: probably just a few cycles slower than if you'd written custom feature setting code by hand for each combination.

Where `StoneFeatures` and `ToPlayFeature` are C++ `structs` that have two static members: a `kNumPlanes` integer containing the number of planes in those features, and a templated `Set<T>` method that sets the features on a tensor with element type `T`.

A lot of this stuff happens at compile-time, but we need some kind of run-time information to stitch a loaded model and the compile-time `Features` structs together. This is accomplished by a simple `FeaturesDescriptor`.

I also moved `Tensor`, `Input` and `Output` out of the `Model` class and into a new header `cc/model/types.h` so they could be shared between `cc/model/model.h` and `cc/model/features.h` without a circular dependency.


tommadams:
Oh, and I've deleted the `DualNet` base class too :)

cc/model/buffered_model.cc
```diff
@@ -17,18 +17,19 @@
...
 
 BufferedModel::BufferedModel(std::vector<std::unique_ptr<Model>> impls)
-    : Model(impls[0]->name(), impls[0]->feature_type(),
+    : Model(impls[0]->name(), impls[0]->feature_descriptor(),
             static_cast<int>(impls.size())) {
   for (auto& x : impls) {
     // Make sure all impls use the same name & input features.
     MG_CHECK(x->name() == name());
-    MG_CHECK(x->feature_type() == feature_type());
+    MG_CHECK(x->feature_descriptor().set_bytes ==
```
amj: why is it just set_bytes that gets compared?

cc/model/features.h
```diff
@@ -0,0 +1,183 @@
...
+#include "cc/color.h"
+#include "cc/constants.h"
+#include "cc/logging.h"
+#include "cc/model/features_internal.h"
+#include "cc/model/types.h"
+#include "cc/platform/utils.h"
+#include "cc/position.h"
+
+namespace minigo {
+
```
amj: an overview of these features & the variadic scheme belongs somewhere, either here or in internal.h

cc/dual_net/dual_net_test.cc
```diff
@@ -12,13 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "cc/dual_net/dual_net.h"
```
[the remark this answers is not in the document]
tommadams: yes, I promise I'll do a clean up pass at some point 🤞

cc/model/buffered_model.cc
tommadams: I was being lazy :P

PR #905, opened 2019-10-03
Update validate.py for new holdout paths

amj:
selfplay etc. change their default paths in mlperf due to the per-device paths.

This updates ml_perf reference implementation to correctly find the paths for holdout games, plus adds a flag to control whether or not to run validation on a holdout set.

Also bugfixes where there's only a single device.


ml_perf/reference_implementation.py
```diff
@@ -443,9 +457,23 @@ def parse_win_stats_table(stats_str, num_lines):
 
 async def train_eval(state, tf_records):
     await train(state, tf_records)
+    if FLAGS.validate:
+      await validate(state)
```
tommadams: Maybe add a TODO to run validate and eval in parallel. Although if we're moving to continuous selfplay it might not be worth bothering with.

amj: validate & eval both run on device 0 at the moment so i think eval has to be multi-device for this to work.

validate.py
```diff
@@ -71,12 +71,19 @@ def main(argv):
     if FLAGS.expand_validation_dirs:
         tf_records = []
         with utils.logged_timer("Building lists of holdout files"):
-            for record_dir in validation_paths:
-                tf_records.extend(gfile.Glob(os.path.join(record_dir, '*', '*', '*.zz')))
+            dirs = validation_paths
+            while len(dirs) > 0:
+                d = dirs.pop()
+                for path, newdirs, files in os.walk(d):
+                    tf_records.extend([os.path.join(path, f) for f in files if f.endswith('.zz')])
```
brilee: using [] inside .extend() is an antipattern because you accumulate the entire extension in memory as a list, and then copy it over to the existing list one by one.

amj: whoops, thanks.

PR #907, opened 2019-10-04
Adds "would capture" features

tommadams:
Still need to do MLPerf testing but it's ready for a look.


cc/dual_net/dual_net_test.cc
```diff
@@ -272,5 +271,34 @@ TYPED_TEST(DualNetTest, TestBackendsEqual) {
   }
 }
 
+TEST(WouldCaptureTest, WouldCapture) {
```
amj: can you test this same position with w to play?

tommadams: done

PR #916, opened 2019-10-09
track the variety of positions played during selfplay

tommadams:


brilee:
The hyperloglog algorithm scales parallel perfectly and can count ridiculously large numbers of unique positions in a global manner. Maybe consider HLL? A primer I wrote: http://www.moderndescartes.com/essays/hyperloglog/

tommadams:
> The hyperloglog algorithm scales parallel perfectly and can count ridiculously large numbers of unique positions in a global manner. Maybe consider HLL? A primer I wrote: http://www.moderndescartes.com/essays/hyperloglog/

That's a really nice technique, I've not seen it before. I just wanted something quick and dirty to measure the effect the inference cache had on game diversity since enabling it seems to harm our ability to train.

PR #920, opened 2019-10-16
Adds `//cc:concurrent_selfplay`.

tommadams:
This uses `MctsTree` instead of `MctsPlayer` to play multiple concurrent
games on each thread, which has multiple benefits:

 - we can keep the GPU busy with far fewer threads.
 - having fewer threads means it's easier to generate meaningful profiles
   of the code.
 - explicit batching without needing to use `BatchingModel` will make it
   easier to try out different batching strategies.

Currently each selfplay thread runs both tree search and inference.

Things left to do:

 - support running more tree search threads than inference threads:
   currently there are two performance problems:
    - with vlosses 2, we don't play enough games to keep the GPU busy
      (we'd need to play something like 20,000 games each iteration to get
      ~80% utilization).
    - Even with 3 or more selfplay threads, TensorFlow ends up running
      all the inferences in lock-step, meaning that work on CPU and GPU
      rarely gets overlapped efficiently.
 - write a `//cc:concurrent_eval` binary that works similarly to
   `concurrent selfplay`.
 - delete the `//cc:eval` and `//cc:selfplay`.
 - delete `BatchingModel` and maybe `MctsPlayer`.

Other changes:
 - I deleted `//cc:puzzle` for now because we don't currently use it.
 - Moved some mode code from `MctsPlayer` to `MctsTree`.
 - The inference engine & device to use are now explicit flags rather
   than being part of the model command line argument.
 - Renamed the `--model` and `--model_two` command line arguments to
   `//cc:eval` to `--eval_model` and `--target_model`: conceptually
   `eval_model` is evaluated against `target_model`.


cc/concurrent_selfplay.cc
```diff
@@ -0,0 +1,873 @@
...
+    game_options.resign_enabled = rnd_() >= FLAGS_disable_resign_pct;
+
+    tree_options = tree_options_;
+
+    selfplay_options.num_virtual_losses = FLAGS_virtual_losses;
+    selfplay_options.num_readouts = FLAGS_num_readouts;
+    selfplay_options.fastplay_readouts = FLAGS_fastplay_readouts;
+    selfplay_options.fastplay_frequency = FLAGS_fastplay_frequency;
+    selfplay_options.noise_mix = FLAGS_noise_mix;
+    selfplay_options.dirichlet_alpha = FLAGS_dirichlet_alpha * 361.0 / (kN * kN);
```
amj: if this is getting rescaled here, then dirichlet alpha should be 0.03 in the flags.  My preference would be to not transform the flag value at all.

cc/concurrent_selfplay.cc
```diff
@@ -0,0 +1,873 @@
...
+      auto selfplay_game = selfplayer_->StartNewGame(model_->name(), verbose);
+      if (selfplay_game == nullptr) {
+        selfplay_games_[i] = std::move(selfplay_games_.back());
+        selfplay_games_.pop_back();
+        continue;
+      } else {
+        selfplay_games_[i] = std::move(selfplay_game);
+      }
+    }
+    i += 1;
```
amj: this is funny, maybe worth a comment.

cc/README.md
```diff
@@ -219,27 +219,26 @@ flag.
...
+ `simple_example` and `gtp` binaries, or `--eval_engine` and `--target_engine`
+for the `eval` binary. The model path is specified using the `--model`
+command line argument (or `--eval_model` and `--target_model` for `eval`):
 
  - **tf**: peforms inference using the TensorFlow libraries built by
    `cc/configure_tensorflow.sh`. Compiled & used as the inference by default,
-   disable the engine with `--define=tf=0`. Use by passing `--model=tf,$PATH`,
-   where `$PATH` is the path to a frozen TensorFlow `GraphDef` proto (as
-   generated by `freeze_graph.py`).
+   disable the engine with `--define=tf=0`. The model should be a frozen
```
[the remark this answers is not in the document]
tommadams: nope, it's enabled by default

cc/mcts_tree.cc
```diff
@@ -435,6 +452,8 @@ void MctsTree::BackupValue(MctsNode* leaf, float value) {
 
 void MctsTree::InjectNoise(const std::array<float, kNumMoves>& noise,
                            float mix) {
+  MG_DCHECK(root_->is_expanded);
```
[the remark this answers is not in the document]
tommadams: We can't expand the root here any more because it required performing an inference (which happens without the `InjectNoise` caller being aware of it); concurrent_selfplay requires that all inferences are performed explicitly.

cc/concurrent_selfplay.cc

cc/concurrent_selfplay.cc

PR #927, opened 2019-10-21
Check examples

tommadams:
inspect_examples.py: show stones in PI output


oneoffs/inspect_examples.py
```diff
@@ -34,13 +34,17 @@
 import sys
 sys.path.insert(0, '.')  # nopep8
 
-from absl import app, flags
-import numpy as np
-import tensorflow as tf
+# Hide the GPUs from TF. This makes startup 2x quicker on some machines.
```
amj: lol

PR #928, opened 2019-10-23
Disable pass during bootstrap games, tweak features

tommadams:
Disabling pass during bootstrap games goes a long way to preventing the model from learning that white should pass as soon as it puts a stone down, which speeds up the MLPerf time to beat target by around 1.8x in my tests.

Feature tweaks: reduce "would capture" feature to 1 plane, "num liberties" planes are now {1, 2, 3+} instead of {1, 2, 3}.

Also filter & sort flags before logging the commands in MLPerf RL loop.


cc/concurrent_selfplay.cc
```diff
@@ -105,7 +105,12 @@ DEFINE_double(policy_softmax_temp, 0.98,
               "policy_softmax_temp to encourage diversity in early play.\n");
 DEFINE_bool(restrict_in_bensons, false,
             "Prevent play in benson's regions after 5 passes have been "
-            "played.\n");
+            "played.");
+DEFINE_bool(allow_pass, true,
+            "If false, pass moves will only be read and played if there is no "
+            "other legal alternative. Disabling pass during the initial rounds "
```
amj: you can strike the second sentence here for simplicity.

cc/concurrent_selfplay.cc
```diff
@@ -129,7 +134,7 @@ DEFINE_uint64(seed, 0,
 DEFINE_double(resign_threshold, -0.999, "Resign threshold.");
 DEFINE_double(disable_resign_pct, 0.1,
               "Fraction of games to disable resignation for.");
-DEFINE_int32(num_games, 1, "Total number of games to play.");
+DEFINE_int32(num_games, 0, "Total number of games to play.");
```
amj: maybe expand the docstring here -- does zero have a special meaning?  i.e., 0 = run continuous?  or is this a safety (i.e. "hey you didn't set it so we quit").

tommadams: The previous value of `1` was a bug: `num_games` and `run_forever` are supposed to be mutually exclusive.

cc/concurrent_selfplay.cc

PR #932, opened 2019-10-26
Promising code tweaks & flag values for 9x9

tommadams:
This PR has my most recent flags and code changes that train a model from scratch to beat the MLPerf 0.6 target in ~105 minutes on an 8 v100 box. Moreover, the three runs I did were remarkably consistent: they all finished within 5 minutes of each other and the tensorboard plots all had the same shape.


ml_perf/reference_implementation.py
```diff
@@ -563,34 +605,20 @@ def rl_loop():
...
-                    state.train_model_path, target_model_path, sgf_dir))
-                if (win_rate_vs_target >= FLAGS.bootstrap_target_win_rate and
-                        prev_win_rate_vs_target > 0):
-                    # The tranined model won a sufficient number of games against
-                    # the target. Create the checkpoint that will be used to start
-                    # the real benchmark and exit.
-                    create_checkpoint()
-                    break
-                prev_win_rate_vs_target = win_rate_vs_target
+            _, win_rate_vs_target = wait([train(state, selfplay_processes),
```
amj: this is nicely simpler!

ml_perf/flags/9/train.flags
```diff
@@ -13,6 +13,6 @@
 --lr_rates=0.16
 --lr_rates=0.016
 --lr_rates=0.0016
---lr_boundaries=25000
---lr_boundaries=37500
+--lr_boundaries=10000
```
amj: did they get to the 10k boundary?

tommadams: Just barely. These boundaries don't really make sense but I want to push the flags files without any changes to make sure I don't accidentally break something.

cc/mcts_tree.cc
```diff
@@ -302,13 +302,6 @@ MctsNode* MctsTree::SelectLeaf(bool allow_pass) {
       return node;
     }
 
-    // HACK: if last move was a pass, always investigate double-pass first
```
amj: attn @brilee

PR #933, opened 2019-11-01
19x19 MLPerf

tommadams:
- Switch the `N, P, W, original_P` data in `MctsNode` from array-of-structures to structure-of-arrays, which allows `CalculateChildActionScore` to access memory more efficiently.
 - SSE implementations of `ArgMax` and `CalculateChildActionScore` which were by far the slowest two methods in the C++ code, which both run > 4x faster than the old code.
 - Support writing examples on multiple threads, which is important for the bootstrap games which finish faster than we can compress & write them.
 - Reduce compression level to 2 from the default 6 when writing training examples to greatly speed up writing.
 - Use a TF callable in `TfDualNet` to remove some per-inference overhead, which makes the `Session::Run` calls about 5% faster.
 - Picked some arbitrary flags for training a 19x19 model.
 - Add names to our threads to help with debugging.
 - WTF tracing tweaks like adding argument names and recording how many nodes were visited during `SelectLeaves`.
 - Optimized `inspect_example.py` because it was taking 2 minutes to read the examples from a single 19x19 game.


brilee:
Everything that's old is new again apparently...
http://www.moderndescartes.com/essays/deep_dive_mcts/

tommadams:
> Everything that's old is new again apparently... http://www.moderndescartes.com/essays/deep_dive_mcts/

Oof, that hurts :(

cc/mcts_tree.cc
```diff
@@ -302,13 +351,13 @@ MctsNode* MctsTree::SelectLeaf(bool allow_pass) {
       return node;
     }
 
-    auto child_action_score = node->CalculateChildActionScore();
-    absl::Span<const float> allowed_moves(child_action_score);
+    PaddedArray<float, kNumMoves> child_action_score;
+    node->CalculateChildActionScoreSse(child_action_score);
```
amj: does this need to be extracted to "SelectLeafSse"?  Should we have mcts_tree_sse.cc?  where does the madness stop?

tommadams: As long as we only have a couple of these methods, probably the most pragmatic solution for non-x86 platforms is to add ifdefs inside of the *Sse methods that call the non-SSE implementations.

cc/algorithm_test.cc
```diff
@@ -0,0 +1,49 @@
...
+
+#include <array>
+
+#include "cc/random.h"
+#include "gtest/gtest.h"
+
+namespace minigo {
+namespace {
+
+TEST(AlgorithmTest, ArgMaxSseRandom) {
```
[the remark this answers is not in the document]
tommadams: added some more tests

PR #934, opened 2019-11-05
Add a TensorShape class

tommadams:
This enables a couple of things:
 - policy & value output tensors no longer have to be 4D.
 - I'll be adding a new abstract `WrappedTensor` class in a follow up PR that wraps an inference engine's tensor type (`tensorflow::Tensor`, `TfLiteTensor`, etc) and it'll be cleaner to share the shape code between `Tensor` and `WrappedTensor`.


cc/model/types.h
```diff
@@ -15,42 +15,120 @@
...
+#include <vector>
+
+#include "cc/inline_vector.h"
+#include "cc/logging.h"
 #include "cc/position.h"
 #include "cc/symmetries.h"
 
 namespace minigo {
 
+// Holds the shape of a tensor.
```
amj: nit: maybe a sentence here about why it's needed?

PR #935, opened 2019-11-07
Various tweaks & clean up

tommadams:
- Randomly pick a resign threshold for each game between 0.8 and 1. This sped up 9x9 training by 5%
- Disallow playing in pass-alive territory once the opponent has passed at least 4 times in a row.
- Remove move limit.
- Use new `Game::AddTrainableMove` and `Game::AddNonTrainableMove` methods instead of `Game::MarkLastMoveAsTrainable`. This means we don't calculate `search_pi` for non-trainable moves.
- Add `MctsPlayer::PlayOpponentsMove` for use in eval to update a player's tree after the opponent plays, without updating the shared `Game` object.


cc/concurrent_selfplay.cc
```diff
@@ -514,7 +531,13 @@ bool SelfplayGame::MaybePlayMove() {
   if (ShouldResign()) {
     game_->SetGameOverBecauseOfResign(OtherColor(tree_->to_play()));
   } else {
-    Coord c = tree_->PickMove(&rnd_);
+    // Hold a reference to the number of consecutive passes played because we'll
+    // update in after `PlayMove`.
```
amj: nit: it

cc/concurrent_selfplay.cc
```diff
@@ -533,18 +556,31 @@ bool SelfplayGame::MaybePlayMove() {
...
+                              std::move(model_str), tree_->root()->Q(),
+                              search_pi);
+    } else {
+      game_->AddNonTrainableMove(tree_->to_play(), c, tree_->root()->position,
+                                 std::move(model_str), tree_->root()->Q());
+    }
 
     if (options_.target_pruning && !fastplay_) {
-      tree_->ReshapeFinalVisits();
+      tree_->ReshapeFinalVisits(restrict_pass_alive_moves);
```
amj: Don't you need to reshape it before you add it as a trainable position?

cc/mcts_tree.cc
```diff
@@ -709,6 +709,17 @@ Coord MctsTree::SoftPickMove(Random* rnd) const {
   for (size_t i = 0; i < cdf.size(); ++i) {
     cdf[i] = std::pow(root_->child_N(i), options_.policy_softmax_temp);
   }
+
+  // Clear probabilities for moves in pass-alive territory.
```
amj: This is thorough but i think we can maybe skip this & the call to CalculatePassAliveRegions:  we do softpick in the first 30 moves.  If there's pass alive territory in the first 30 moves, things are probably pretty busted...

tommadams: done

cc/concurrent_selfplay.cc

cc/concurrent_selfplay.cc
tommadams: oof, thanks for catching that

PR #936, opened 2019-11-07
Write N, Q and move played to TF examples

tommadams:


cc/tf_utils.cc
```diff
@@ -69,6 +69,15 @@ tensorflow::Example MakeTfExample(const Tensor<uint8_t>& features,
   // outcome is a single float.
   dst_features["outcome"].mutable_float_list()->add_value(outcome);
 
+  // Q is a single float.
+  dst_features["q"].mutable_float_list()->add_value(Q);
+
+  // Number of reads is a single float.
```
amj: why a float?

cc/tf_utils.cc
```diff
@@ -117,7 +126,9 @@ std::vector<tensorflow::Example> MakeExamples(
     game.GetPositionHistory(i, kMaxPositionHistory, &input.position_history);
 
     feature_desc.set_bytes({&input}, &features);
-    examples.push_back(MakeTfExample(features, move->search_pi.value(), game.result()));
+    examples.push_back(MakeTfExample(
+          features, move->search_pi.value(), move->Q, move->N, move->c,
```
amj: we're probably gonna need to change our BT code too at some point before v18, right

cc/tf_utils.cc
tommadams: bad comment

cc/tf_utils.cc
tommadams: this will work as-is but we'll have to change the BT code if we want to store Q & N in separate columns

PR #940, opened 2019-11-18
multi-machine selfplay

tommadams:
Adds support for running selfplay on multiple machines. I tested this using two v100 VMs, writing the training examples and models to a [Cloud Filestore](https://cloud.google.com/filestore/), which is _waaay_ faster than GCS.

I'm still figuring out the best way to run this whole thing but for now, I've dumped a few simple scripts into `ml_perf/scripts/`. The benchmark can be run as follows:

```bash
# Some location visible to all machines.
BASE_DIR=/mnt/mlperf/results/

# Builds the C++ binaries, initializes the output directory structure, plays
# bootstrap games with a random model.
./ml_perf/scripts/bootstrap.sh --board_size=19 --base_dir/$BASE_DIR

# Starts the selfplay binaries running in the background.
# This can be run on multiple machines, provided they can all see the base_dir.
./ml_perf/scripts/selfplay.sh --board_size=19 --base_dir/$BASE_DIR

# Starts the training loop. When the training loop finishes or is killed, it writes
# an abort file to the output directory, which stops the selfplay jobs.
./ml_perf/scripts/train.sh --board_size=19 --base_dir/$BASE_DIR
```


ml_perf/flags/19/rl_loop.flags
```diff
@@ -15,5 +15,8 @@
 
 --eval_num_games=100
 
---min_games_per_iteration=8192
 --validate=1
+
+# Consider also updating num_games in bootstrap.flags if updating
+# min_games_per_iteration.
+--min_games_per_iteration=8192
```
amj: why is this duplicated in train_loop?

ml_perf/scripts/common.sh
```diff
@@ -0,0 +1,17 @@
+set -eu
```
amj: needs license, usage, at least one comment :)

ml_perf/train_loop.py
```diff
@@ -0,0 +1,245 @@
...
+                   'Fraction of selfplay games to pass to training.')
+
+flags.DEFINE_boolean('validate', False, 'Run validation on holdout games')
+
+flags.DEFINE_integer('min_games_per_iteration', 4096,
+                     'Minimum number of games to play for each training '
+                     'iteration.')
+
+flags.DEFINE_integer('num_read_threads', 8,
+                     'Number of threads to read examples on.')
```
amj: any guidance in setting these?

ml_perf/train_loop.py
```diff
@@ -0,0 +1,245 @@
...
+
+flags.DEFINE_integer('num_write_threads', 8,
+                     'Number of threads to write examples on.')
+
+FLAGS = flags.FLAGS
+
+
+class State:
```
amj: a short comment here would be helpful.

ml_perf/train_loop.py
```diff
@@ -0,0 +1,245 @@
...
+def wait_for_training_examples(state, num_games):
+    """Wait for training examples to be generated by the latest model.
+
+    Args:
+        state: the RL loop State instance.
+        num_games: number of games to wait for.
+    """
+
+    for i in itertools.count():
+        model_dir = os.path.join(fsdb.selfplay_dir(), state.selfplay_model_name)
```
amj: nit: could these be outside the loop?

ml_perf/utils.py
```diff
@@ -26,6 +26,12 @@
 from utils import *
 
 
+# Flags that take multiple values.
+# For these flags, expand_cmd_str appends each value in order.
+# For all other flags, expand_cmd_str takes the last value.
+MULTI_VALUE_FLAGS = set(['--lr_boundaries', '--lr_rates'])
```
amj: nice.  thanks!

ml_perf/scripts/train.sh
```diff
@@ -0,0 +1,20 @@
+source ml_perf/scripts/common.sh
+
+
+# Set up an exit handler that stops the selfplay workers.
+function stop_selfplay {
+  /ml_perf/scripts/stop_selfplay.sh \
+    --board_size="${board_size}" \
+    --base_dir="${base_dir}" \
+}
+trap stop_selfplay EXIT
```
amj: Should we trap sigterm also?

ml_perf/flags/19/rl_loop.flags
tommadams: `rl_loop.flags` is deprecated (along with `ml_perf/reference_implementation.py`, `cc/selfplay.cc`, etc). I'm keeping the files around for now as reference but I'll do a clean up pass once the code for MLPerf 0.7 is settled.

ml_perf/scripts/train.sh
tommadams: IIUC `EXIT` covers `SIGTERM` too because the handler is always run when the script exits, regardless of the reason.

ml_perf/train_loop.py
tommadams: done

ml_perf/train_loop.py
tommadams: done.

ml_perf/flags/19/rl_loop.flags
amj: Sounds good.  As long as multi-vm can fallback it sounds like an elegant solution! :thumbsup:

ml_perf/scripts/train.sh
amj: got it, thanks!

ml_perf/scripts/common.sh
tommadams: done

ml_perf/train_loop.py
tommadams: done

PR #942, opened 2019-11-20
update to TF 1.15.0

tommadams:


tommadams:
I need to update the testing base image before the tests will pass. I'll do that after this PR has been reviewed.

cc/configure_tensorflow.sh
```diff
@@ -19,6 +19,14 @@ if [ -z "${TF_NEED_TENSORRT}" ]; then
   esac
 fi
 
+# //org_tensorflow//:configure script must be run from the TensorFlow repository
+# root. Build the script in order to pull the repository contents from GitHub.
+# The `bazel fetch` and `bazel sync` commands that are usually used to fetch
+# external Bazel dependencies don't work correctly on the TensorFlow repository.
```
amj: :( is there a tensorflow issue or someone we can @ in on this?

It seems like it could be written "the tensorflow repository is misconfigured, causing bazel fetch and bazel sync to break" ;)

tommadams: Getting Bazel fetch & sync working with TF repo is probably infeasible (or at least not worth the effort). The problem is that the configure script disables a bunch of dependencies (mobile, windows, etc) that would otherwise not work. Running Bazel fetch or sync before the configure script tries to pull down a bunch of unsupported deps.

A far more tractable problem to fix is to make the configure script work as an external dependency. If I understand the problem correctly, it should be a fairly small, localized change. I'll do that once I've finished upgrading to TF 1.15.0.

PR #943, opened 2019-11-21
use explicit imports for tf.contrib modules

tommadams:
TF 2.0 doesn't include a contrib module, so this should simplify an eventual upgrade.


dual_net.py
```diff
@@ -30,10 +30,13 @@
...
 import tensorflow as tf
-from tensorflow.contrib import summary
-from tensorflow.contrib.tpu.python.tpu import tpu_config
-from tensorflow.contrib.tpu.python.tpu import tpu_estimator
-from tensorflow.contrib.tpu.python.tpu import tpu_optimizer
+from tensorflow.contrib import cluster_resolver as contrib_cluster_resolver
+from tensorflow.contrib import quantize as contrib_quantize
+from tensorflow.contrib import summary as contrib_summary
+from tensorflow.contrib import tpu as contrib_tpu
+from tensorflow.contrib.tpu.python.tpu import tpu_config as contrib_tpu_python_tpu_tpu_config
```
brilee: I think some of this stuff is under tf1.15's tf.compat.v2.distribute.tpu*

tommadams: Oh, those changes be part of some future PR that cleans up all the deprecated imports.

PR #950, opened 2019-12-13
Add support for NCHW input features

tommadams:
Change summary:
- The feature types (`StoneFeatures`, `WouldCaptureFeature`, etc) previously just at a single static `Set` method; now they have both `SetNhwc` and `SetNchw`.
- Added NCHW symmetry support to both C++ and Python training code.
- The input layout (NHWC/NCHW) must be specified when creating a `FeatureDescriptor`.
- The `FeatureDescriptor` now has a helper `GetTensorShape` that calculates the correct tensor shape for a given batch size.
- Added a benchmark to prove that NCHW features are actually measurably faster.
- Fixed a bug in the logging code where logging an empty string didn't work correctly (e.g. `MG_LOG(INFO) << "";`).
- Added support to eval code for different number of readouts for eval and target model.
- Added support for running eval on multiple GPUs.

Results from `//cc/model:features_benchmark`:

19x19:agz:nhwc:float 321.232811ms
19x19:agz:nchw:float 266.719199ms
19x19:agz:nhwc:uint8 216.62693ms
19x19:agz:nchw:uint8 114.057009ms

19x19:mlperf07:nhwc:float 438.667451ms
19x19:mlperf07:nchw:float 481.728811ms
19x19:mlperf07:nhwc:uint8 323.920743ms
19x19:mlperf07:nchw:uint8 196.413271ms


cc/model/features.h
```diff
@@ -44,52 +47,159 @@ namespace minigo {
...
+    static_assert(sizeof(Stone) == 2, "");
+    __m128i color_mask = _mm_set1_epi16(3);
+
+    // Initialize useful single byte values.
+    __m128i one = _mm_set1_epi8(1);
+    __m128i my_color_mm = _mm_set1_epi8(static_cast<int>(my_color));
+    __m128i their_color_mm = _mm_set1_epi8(static_cast<int>(their_color));
+
+    // Write the features for the position history that we have.
+    int safe_size = (kN * kN / 16) * 16;
```
sethtroisi: consider adding
int layer_size = kN * kN;
to aid comprehension

symmetries.py
```diff
@@ -128,3 +129,49 @@ def rotate(tensor):
...
+
+    pi_move = tf.slice(pi, [0], [squares], name="slice_moves")
+    pi_pass = tf.slice(pi, [squares], [1], name="slice_pass")
+    # Add a final dim so that x and pi have same shape: [N,N,num_features].
+    pi_n_by_n = tf.reshape(pi_move, [1, go.N, go.N])
+
+    with tf.control_dependencies([assert_shape_x, assert_shape_pi]):
+        pi_rot = tf.concat(
+            [tf.reshape(rotate(pi_n_by_n), [squares]), pi_pass],
+            axis=0)
```
sethtroisi: consider passing in layout to rotate then not having to repeat this block.
I'm not a big fan but maybe.

cc/model/features.h
tommadams: Good call. Added a `kNumPoints = kN * kN` constant to `cc/constants.h`. I should have done that a long time ago.

symmetries.py
tommadams: Yeah, I spent a while trying to figure out the cleanest way to do this and then just punted and went with the golang approach: copy-paste.

PR #963, opened 2020-01-22
Revert TPU hack

### LINKED ISSUE #870: Remove tf.control_dependencies from dual_net.py
In order to get multiple models running on the same TPU, we had to add the following hack to work around a bug in TensorFlow: https://github.com/tensorflow/minigo/blob/d5e9697318b364ccc2b977a4805a96f1e72e7abc/dual_net.py#L523

This bug got fixed  in TF 1.12.1, so we can remove our hack now:
https://github.com/tensorflow/tensorflow/commit/735d26a821d9f69f58ad12ea841da9bc8c5a65ee#diff-c344ded643ece6ebf6370acab3a9a524L306

sethtroisi:
Original Hack added in https://github.com/tensorflow/minigo/pull/612
Closes #870

### DIFF
--- a/dual_net.py
+++ b/dual_net.py
@@ -567,13 +567,9 @@ def model_inference_fn(features, training, params):
 def tpu_model_inference_fn(features):
     """Builds the model graph suitable for running on TPU.
 
-    It does two things:
      1) Mark all weights as constant, which improves TPU inference performance
         because it prevents the weights being transferred to the TPU every call
         to Session.run().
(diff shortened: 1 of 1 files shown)


tommadams:
Thanks for the clean up but the TPU folks tell me that neither of the hacks in `tpu_model_inference_fn` are needed anymore. 

I'm going to close this PR because TPU best practices have changed a lot since I wrote `TpuDualNet` and we're working on a new version.

PR #967, opened 2020-01-31
Adds //cc:concurrent_eval

tommadams:
This will eventually replace `//cc:eval`.

It started life as a copy-paste of `//cc:concurrent_selfplay` with some simplifications (it doesn't write training examples, reload models, soft pick, etc) and added support for playing two models on two separate trees.

I've done a cursory verification that the code is correct by playing a handful of models against each other and verifying that the winrates produced by `//cc:concurrent_eval` are close to those from `//cc:eval`. We should do some more testing before replacing the old eval code completely.

Once this code is pushed, we should do another pass over it to factor out logic shared between this and `//cc:concurrent_selfplay`.

I also dropped in a oneoff script that scrapes a directory of eval logs and generates a plot of their win rate.

### DIFF
--- a/cc/concurrent_eval.cc
+++ b/cc/concurrent_eval.cc
@@ -1,0 +1,758 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cmath>
+#include <functional>
+#include <memory>
+#include <vector>
+
+#include "absl/memory/memory.h"
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_format.h"
+#include "absl/strings/str_join.h"
+#include "absl/strings/str_replace.h"
+#include "absl/strings/str_split.h"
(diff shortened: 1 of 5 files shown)


cc/concurrent_eval.cc
```diff
@@ -0,0 +1,760 @@
...
+  const Options options_;
+  std::unique_ptr<Game> game_;
+  std::unique_ptr<MctsTree> black_tree_;
+  std::unique_ptr<MctsTree> white_tree_;
+  Random rnd_;
+
+  // Number of consecutive passes played by black and white respectively.
+  // Used to determine when to disallow playing in pass-alive territory.
+  // `num_consecutive_passes_` latches once it reaches
+  // `restrict_pass_alive_play_threshold` is is not reset to 0 when a non-pass
```
amj: nit: 'is is'

cc/concurrent_eval.cc
```diff
@@ -0,0 +1,760 @@
...
+  // Plays a move.
+  void PlayMove(Color color);
+
+ private:
+  MctsTree* GetTree(Color color) {
+    return color == Color::kBlack ? black_tree_.get() : white_tree_.get();
+  }
+
+  // Returns true if the predicted win rate is below `resign_threshold`.
+  bool ShouldResign(Color color) const;
```
amj: nit:  Color toPlay, presumably?

cc/concurrent_eval.cc
```diff
@@ -0,0 +1,760 @@
...
+  // Run the eval threads.
+  for (auto& t : eval_threads) {
+    t->Start();
+  }
+  for (auto& t : eval_threads) {
+    t->Join();
+  }
+
+  // Stop the output threads by pushing one null game onto the output queue
+  // for each thread, causing the treads to exit when the pop them off.
```
amj: nit: they

amj (approved):
all nits, LGTM

PR #979, opened 2020-03-17
Minigui improvements

tommadams:
- Rewrite the SGF parsing code to leave the SGF structure intact, rather than converting it to a tree of moves. This gives us somewhere to put the root node's comment.
 - Fix the win-rate estimation by returning the Q of the most promising child node, rather than the current node's Q.


cc/sgf.h
```diff
@@ -33,48 +34,62 @@ namespace sgf {
...
 constexpr char kProgramIdentifier[] = "Minigo";
 
-// Abstract syntax tree for an SGF file.
-// The Ast class just holds the structure and contents of the tree and doesn't
-// infer any meaning from the property IDs or values.
-class Ast {
- public:
-  struct Property {
-    std::string ToString() const;
+//  Collection = GameTree { GameTree }
```
amj: maybe put a link to https://www.red-bean.com/sgf/ at the top of this comment.

minigui/static/position.js
```diff
@@ -111,6 +111,9 @@ define(["require", "exports", "./base", "./util"], function (require, exports, b
             return null;
         }
         update(update) {
+            if (update.comment !== undefined) {
```
amj: are there .ts changes missing?

PR #988, opened 2020-04-15
Minor formatting updates to RESULTS.md

w-hat:


RESULTS.md
```diff
@@ -566,7 +576,10 @@ v10, while reaching professional strength, would still struggle with ladders.
 In a ladder sequence, only one node in the MCTS tree would read the ladder,
 while the rest of the batch would need to pick non-ladder moves.  It's
 conceivable that the averaging done by MCTS would essentially encourage the
-models to avoid sharp, narrow paths with high expected value in favor of flatter paths with a less dramatic outcome.  We figured we'd test this by setting batch size to 2, doubling the number of games played at once (to achieve roughly the same throughput), and run again.
+models to avoid sharp, narrow paths with high expected value in favor of flatter
+paths with a less dramatic outcome.  We figured we'd test this by setting batch
+size to 2, doubling the number of games played at once (to achieve roughly the
```
w-hat, line 581: Should this say "halving the number of games played at once"?

tommadams: No. With a virtual loss of 4, each game selects mini-batches of four leaves at a time to run inference one. We run multiple games in parallel on a single accelerator during selfplay, so the number of parallel inferences is `num_parallel_games * num_virtual_losses`. So when we reduced the number of virtual losses to 2, we doubled the number of parallel games to achieve roughly the same throughput.

w-hat: Ah, I had read this as going from not using virtual losses at all to selecting two positions per inference per game.  Thanks!

Issue #982, opened 2020-03-19
Wrong argument passed in minigui/fetch-and-run.sh


smolendawid:
In _minigui/fetch-and-run.sh_ `--conv_width` argument is passed to the `gtp.py`, but the script doesn't take it

sethtroisi:
Thanks for reporting, and thanks walmsley for fixing.

PR #990, opened 2020-04-23
Fixed MiniGUI-related crashes

walmsley:
On a fresh setup, I was trying to get MiniGUI running, and came across two minor crashes that seem to be caused by legacy code that lagged behind changes made elsewhere in the repository. This PR fixes both of those crashes.

One of the issues is this one: https://github.com/tensorflow/minigo/issues/982


sethtroisi:
Thanks for testing and fixing @walmsley

PR #994, opened 2020-05-27
optimization in git clone using --depth

Rajpratik71:
optimize the git clone using --depth flag in term of size of clone
and also in term's of time taken to fetch the files and commit history
of whole repository .

More detail can be found at blog
https://www.atlassian.com/git/tutorials/big-repositories

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,7 +456,7 @@ GCE_VM_NAME=minigo-etpu-test
 GCE_ZONE=us-central1-f
 
 # Clone the Minigo Github repository:
-git clone https://github.com/tensorflow/minigo
+git clone --depth 1 https://github.com/tensorflow/minigo
 cd minigo
 
 # Install virtualenv.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -456,7 +456,7 @@ GCE_VM_NAME=minigo-etpu-test
 GCE_ZONE=us-central1-f
 
 # Clone the Minigo Github repository:
-git clone https://github.com/tensorflow/minigo
+git clone --depth 1 https://github.com/tensorflow/minigo
```
sethtroisi, line 459: can we keep this one as is? It's nice to have full history when developing.

Rajpratik71: @sethtroisi  , this section tutorial is for retraining a model on GCP , Only being stored in Cloud bucket to be used for model training .

That why i think no one is going to look at history .

sethtroisi: ahh. I had just assumed :) Thanks for the explination.

sethtroisi:
LGTM, thanks @Rajpratik71

